PR #22, opened 2020-03-26
Near-finalized ingest of core-node socket data

zone117x:
* Closes https://github.com/blockstack/blockstack-core-sidecar/issues/12: Consume core-node socket messages

* Closes https://github.com/blockstack/blockstack-core-sidecar/issues/15: Consume Clarity values from core-node stream

* Progress on https://github.com/blockstack/blockstack-core-sidecar/issues/10: Define db schema

* Progress on https://github.com/blockstack/blockstack-core-sidecar/issues/8: Fork-tolerance, determine and pressure test approaches

The core-node event stream payloads are about finalized in PR https://github.com/blockstack/stacks-blockchain/pull/1319. This PR consumes all available core-node msg data and transforms into database-ready objects. 

Blocks, transactions, token events, and smart-contract events are transformed into a WIP db structure, along with some naive upserts implemented in a postgres datastore and an in-memory datastore. 

Careful consideration is going into the postgres schema in order to scale to large future datasets. An approach similar to bitcore's mongodb is being taken, i.e. rows have an indexed boolean value indicating their "canonical" (is valid for current chain-tip) state. This, along with an indexed block-height column should allow efficient adjustments for last N blocks during re-org events. Benchmarking will take place once mature table schemas and aggregate sql queries are ready.

### DIFF
--- a/src/datastore/memory-store.ts
+++ b/src/datastore/memory-store.ts
@@ -3,9 +3,9 @@ import { DataStore, DbBlock, DbTx, DbStxEvent, DbFtEvent, DbNftEvent, DbSmartCon
 export class MemoryDataStore implements DataStore {
   readonly blocks: Map<string, DbBlock> = new Map();
   readonly txs: Map<string, DbTx> = new Map();
-  readonly stxEvents: Map<string, DbStxEvent> = new Map();
-  readonly ftEvents: Map<string, DbFtEvent> = new Map();
-  readonly nftEvents: Map<string, DbNftEvent> = new Map();
+  readonly stxTokenEvents: Map<string, DbStxEvent> = new Map();
+  readonly fungibleTokenEvents: Map<string, DbFtEvent> = new Map();
+  readonly nonFungibleTokenEvents: Map<string, DbNftEvent> = new Map();
   readonly smartContractEvents: Map<string, DbSmartContractEventTypeId> = new Map();
 
   updateBlock(block: DbBlock): Promise<void> {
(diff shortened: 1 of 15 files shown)


src/datastore/memory-store.ts
```diff
@@ -1,8 +1,12 @@
-import { DataStore, DbBlock, DbTx } from './common';
+import { DataStore, DbBlock, DbTx, DbStxEvent, DbFtEvent, DbNftEvent, DbSmartContractEventTypeId } from './common';
 
 export class MemoryDataStore implements DataStore {
   readonly blocks: Map<string, DbBlock> = new Map();
   readonly txs: Map<string, DbTx> = new Map();
+  readonly stxEvents: Map<string, DbStxEvent> = new Map();
+  readonly ftEvents: Map<string, DbFtEvent> = new Map();
```
kyranjamie: I'd err on the side of being explicit here and not using acronym for readabilities sake

tests/datastore-tests.ts
```diff
@@ -1,19 +1,22 @@
...
+import { DbBlock, DbTx, DbTxTypeId } from '../src/datastore/common';
+import { PgDataStore, cycleMigrations, runMigrations } from '../src/datastore/postgres-store';
 
 describe('in-memory datastore', () => {
   let db: MemoryDataStore;
+
   beforeAll(() => {
     db = new MemoryDataStore();
   });
+
```
kyranjamie: Don't we want to initialise on every test? `beforeEach()`

zone117x: Good call, fixed

PR #24, opened 2020-03-30
Feat/api expansion

zone117x:
### Changes
* Options to configure sidecar and API host and port via env vars.
* Update Express routes for async handlers.
* Dev / debug endpoints w/ minimal front-ends for testing token transfers and contract deployments.
* Near completed API response for transactions -- created from database entries.
* Endpoint for streaming transaction updates.

### Testing
* Run the core-node testnet using the `feature/genesis` branch -- configure the `Stacks.toml` to use the same addresses as defined in https://github.com/blockstack/stacks-blockchain-sidecar/blob/57c9336afd0be5ac2398e05c251f5263cf5e833b/src/api/routes/debug.ts#L13-L38
* Update the mempool paths in both the core-node `Stacks.toml` file and the sidecar `.env` file.
* Run the sidecar, run the core-node testnet, then open the urls:
  * http://127.0.0.1:3999/debug/watch-tx
  * http://127.0.0.1:3999/debug/broadcast/token-transfer
  * http://127.0.0.1:3999/debug/broadcast/contract-deploy
 These allow transaction broadcasting to the testnet (similar to a faucet), and a display of the API data for the tx results.

image

image


src/api/routes/tx.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export function createTxRouter(): express.Router {
+  const router = addAsync(express.Router());
+
+  router.use(cors());
+
+  router.getAsync('/', async (req, res) => {
+    const db: DataStore = req.app.get('db');
+    const txs = await db.getTxList();
+    const fullTxs = await Bluebird.mapSeries(txs.results, async tx => {
```
hstove: Are there memory (or other) limitations that necessitate `mapSeries`? The `Bluebird.map` function supports a `concurrency` option.

Additionally, are there limitations that prevent us from getting all TX info (including events) via a single query with JOIN? Right now it looks like the number of queries is `n^2`

hstove: After further reading, I see that we have a memory store and a PG store, which probably answers my questions.

zone117x: Yep -- the async map is a temp hack until the proper sql backend is implemented. This will use an optimized join queries when ready

PR #27, opened 2020-04-02
feat: render types from json schema

kyranjamie:
Adds JSON Schema files for transactions coming from the sidecar. We generate TypeScript typings, from these files, that utilise discriminating union types.

[Checkout the output types](https://gist.github.com/kyranjamie/11a6acaed43c0e9c3334bf2c002ebb96#file-stacks-entities-tsx-L143)


scripts/generate-types.ts
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env ts-node
```
zone117x, line 1: The main way to run this file should probably be a script in package.json e.g. `"something": "ts-node ./scripts/generate-types.ts"`, so as not to require *nix or a global ts-node install of unknown version

PR #37, opened 2020-04-17
Feat/core node proxy

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/29

All core-node RPC endpoints (`/v2/*`) are directly proxied through to the sidecar API without modification. 

This is a breaking changes for the sidecar endpoints. These are now prefixed with `/sidecar/v1`. E.g.: the previous endpoint `http://127.0.0.1:3999/status` is now `http://127.0.0.1:3999/sidecar/v1/status`.

Test core-node RPC proxy by visiting an endpoint like `http://127.0.0.1:3999/v2/info`.


hstove:
Haven't tested yet, but does this include CORS headers? That'd be great. And this is great, thank you.

zone117x:
@hstove CORS now added to RPC proxy

PR #39, opened 2020-04-17
feat: adds STX faucet

hstove:
Adds a faucet API as `/debug/faucet?address={address}`

Requires an ENV variable `FAUCET_PRIVATE_KEY`

### DIFF
--- a/src/api/routes/debug.ts
+++ b/src/api/routes/debug.ts
@@ -374,5 +374,38 @@ export function createDebugRouter(db: DataStore): RouterWithAsync {
     res.set('Content-Type', 'text/html').send(txWatchHtml);
   });
 
+  router.postAsync('/faucet', async (req, res) => {
+    try {
+      const { address } = req.query;
+      const { FAUCET_PRIVATE_KEY } = process.env;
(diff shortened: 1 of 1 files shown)


src/api/routes/debug.ts
```diff
@@ -374,5 +374,40 @@ export function createDebugRouter(db: DataStore): RouterWithAsync {
...
+        return;
+      }
+
+      const tx = makeSTXTokenTransfer(
+        address,
+        new BN(10e3),
+        new BN(10),
+        process.env.FAUCET_PRIVATE_KEY,
+        {
+          nonce: new BN(0),
```
zone117x, line 394: What is the intended usage here? Shouldn't this use the core-node RPC to retrieve the nonce for `FAUCET_PRIVATE_KEY` and then to submit the transaction?

hstove: Ah yeah my bad - forgot to add that. This faucet was written before the RPC API was available. Will add.

hstove: Good catch - added it back

PR #43, opened 2020-04-21
refactor: touch up post-condition types

kyranjamie:
This PR deserialises post-conditions, principal structures for the API, and adds these new fields to the JSON Schema.

Looking for feedback on how this can be tested, and how `post_condition_mode` can be added to DbTx.


docs/entities/post-conditions/post-condition-2-non-fungible.schema.json
```diff
@@ -4,26 +4,29 @@
...
+        }
+      }
     },
     {
       "type": "object",
       "required": ["conditionCode", "amount", "assetInfoId", "assetValue", "asset"],
       "properties": {
         "conditionCode": {
-          "enum": [10, 11],
+          "enum": [16, 17],
```
zone117x, line 20: I think since we're expanding the opaque encoding to a readable JSON encoding, we should use descriptive string constants for these. 
`["sent", "not_sent"]` as opposed to numbers. Thoughts?

docs/entities/post-conditions/post-condition-1-fungible.schema.json
```diff
@@ -0,0 +1,49 @@
...
+          "$ref": "./post-condition-principal.schema.json"
+        }
+      }
+    },
+    {
+      "type": "object",
+      "required": ["conditionCode", "amount", "assetInfoId", "asset"],
+      "properties": {
+        "conditionCode": {
+          "enum": [1, 2, 3, 4, 5],
```
zone117x, line 20: Same here -- descriptive string constants rather than magic numbers?

docs/entities/post-conditions/post-condition-principal.schema.json
```diff
@@ -0,0 +1,68 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "PostConditionPrincipal",
+  "type": "object",
+  "definitions": {
+    "StacksAddress": {
```
zone117x, line 6: This should be a string type (a STX address string)

src/datastore/common.ts
```diff
@@ -69,6 +69,9 @@ export interface DbTx {
   smart_contract_contract_id?: string;
   smart_contract_source_code?: string;
 
+  /** Only valid for `smart_contract` & `contract_call` tx types. */
```
zone117x, line 72: I don't think we really want this is a separate field in the db (it will never be used in a query)

docs/entities/transactions/transaction-2-contract-call.schema.json
```diff
@@ -38,8 +38,15 @@
             }
           }
         },
+        "post_condition_mode": {
+          "type": "integer",
+          "enum": [1, 2]
```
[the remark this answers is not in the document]
kyranjamie: thx, missed this one

docs/entities/post-conditions/post-condition-0-stx.schema.json
```diff
@@ -0,0 +1,32 @@
...
+      "required": ["principal"],
+      "properties": {
+        "principal": {
+          "$ref": "./post-condition-principal.schema.json"
+        }
+      }
+    },
+    {
+      "type": "object",
+      "required": ["condition_code", "amount", "asset_info_id"],
```
zone117x, line 17: Mismatch with `condition_code` here. A tx has a single condition code, then an array of post conditions (as opposed to each post condition having a condition code)

zone117x: nvm! I mixed up condition code and condition mode!

PR #44, opened 2020-04-21
feat: broadcast from faucet

hstove:
Forgot to include this functionality.

Also, can we setup the ENV variable with a private key for a genesis block address? So we can fully wire this up?


zone117x:
Just use `testnetKeys[0].secretKey` (or any others in the array), these are already configured for STX balance

hstove:
Thanks Matt, I've added that as a default

PR #67, opened 2020-04-28
feat: handle reorg events with in-memory datastore

zone117x:
Detect and process reorg events (supported with the in-memory datastore for now). 

This makes the sidecar more functional during neon & follower node reboots. 

A `GET /sidecar/v1/block` endpoint was also added (currently useful for testing, schemas TODO).


src/api/controllers/db-controller.ts
```diff
@@ -86,13 +87,22 @@ function getAssetEventTypeString(
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface Block extends DbBlock {}
```
kyranjamie, line 91:
```suggestion
export type Block = DbBlock;
```

should remove need for eslint override?

PR #70, opened 2020-04-28
feat: add BTC faucet

hstove:
Requires a few ENV variables for RPC client and the faucet private key


src/types/bitcoin-core.d.ts
```diff
@@ -0,0 +1,1099 @@
+/* eslint-disable tsdoc/syntax */
+declare module 'bitcoin-core' {
```
zone117x, line 2: Pulling in this typedef is kinda crazy. It looks like this lib is hostile towards typescript https://github.com/ruimarinho/bitcoin-core/issues?q=typescript

Would you be open to using a different one like https://github.com/vansergen/rpc-bitcoin ?

hstove: Great suggestion - I looked around for a sec but didn't see that package. Much better. I've implemented it.

src/tests/faucet-tests.ts
```diff
@@ -0,0 +1,10 @@
+import * as RpcClient from 'bitcoin-core';
```
zone117x, line 1: Bad import here

hstove:
@zone117x thanks for the catch, fixed and rebased

PR #78, opened 2020-05-04
feat: add hex prefix if not given, dev-friendly

kyranjamie:
Minor update but hopefully a helpful one. Added this as noticed that `broadcastTransaction` from stacks-transaction returns non-prefixed txid.


src/api/routes/tx.ts
```diff
@@ -78,6 +78,11 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
 
   router.getAsync('/:tx_id', async (req, res) => {
     const { tx_id } = req.params;
+
+    if (!has0xPrefix(tx_id)) {
```
zone117x, line 82: Missing `has0xPrefix` import?

PR #79, opened 2020-05-04
Feat/pg retain orphaned events

zone117x:
* Full tx event data is now returned for orphaned transaction queries. 
* The db schema change also allows event data to be recovered when a chain tip flips from non-canonical back to canonical status -- however, the reorg logic for this scenario is not yet implemented.

### DIFF
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -107,7 +107,7 @@ export async function getTxFromDataStore(
   if (!txQuery.found) {
     return { found: false };
   }
-  const { results: dbTxEvents } = await db.getTxEvents(txId);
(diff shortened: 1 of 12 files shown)


aulneau:
@zone117x can you elaborate on what "orphaned transaction queries" are specifically, and how someone would get one?

zone117x:
@aulneau Sure, when a fork happens, all transactions (and associated event data) contained within the orphaned blocks are marked as `non-canonical`, which means they are no longer a part of the new, valid blockchain. 

In the future, on mainnet, these reorg scenarios should typically only result in the last few blocks being orphaned. The current testnet has a bug that results in _all_ blocks being wiped out (effectively orphaned), several times a day. This is why any transactions deployed right now don't stick around long. 

This PR expands the reorg handling logic at the sidecar database layer, so that this orphaned data stays around.

PR #81, opened 2020-05-05
Feat/core node http stream

zone117x:
Update the event stream listener from a TCP server to an HTTP server. See https://github.com/blockstack/stacks-blockchain/pull/1535


src/event-stream/event-server.ts
```diff
@@ -202,55 +189,68 @@ async function handleClientMessage(clientSocket: Readable, db: DataStore): Promi
...
-      throw new Error(
-        `STACKS_SIDECAR_SOCKET_HOST must be specified, e.g. "STACKS_SIDECAR_SOCKET_HOST=127.0.0.1"`
-      );
-    }
-    if (!socketPort) {
-      throw new Error(
-        `STACKS_SIDECAR_SOCKET_PORT must be specified, e.g. "STACKS_SIDECAR_SOCKET_PORT=3700"`
-      );
+  let eventHost = process.env['STACKS_SIDECAR_SOCKET_HOST'];
+  const eventPort = Number.parseInt(process.env['STACKS_SIDECAR_SOCKET_PORT'] ?? '', 10);
```
kyranjamie, line 208: `parseInt` is available in global scope. Is `Number.` needed?

PR #87, opened 2020-05-06
Btc faucet fixes & bitcoind integration tests

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/47

Updates the btc faucet code to longer be dependent on RPC wallet config, or any other privileged RPC access. The faucet can operate with only:
* A btc key that has been granted tokens (a process that only needs done once when the regtest node chain is created for the first time)
* A handful of RPC calls that pose no security risks except for possible DDoS: `scantxoutset`, `getrawtransaction`, `sendrawtransaction` -- the core-node could whitelist these only to the sidecar service if needed.


src/btc-faucet.ts
```diff
@@ -1,87 +1,159 @@
...
+  const faucetWallet = getFaucetWallet(network);
 
-  const network = btc.networks.regtest;
+  const faucetAmountSats = faucetAmount * 1e8;
 
-  const faucetPK = btc.ECPair.fromWIF(BTC_FAUCET_PK, btc.networks.regtest);
-  const faucetPayment = btc.payments.p2wpkh({
-    pubkey: faucetPK.publicKey,
-    network,
+  const txOutSet: TxOutSet = await client.scantxoutset({
```
lgalabru, line 85: qq - why are we using `scantxoutset` (vs `listunspent` - not flagged with an experimental warning in the Core API)?

zone117x: `listunspent` is a wallet function. If we are going to use wallet functions then the whole faucet is maybe 3 lines of code that call `sendtoaddress`. My understanding is that we want the faucet to be able to operate given a keypair with some regtest btc. 

Also `scantxoutset` has been around for 2-3 years.

zone117x: @ludo let me know if the wallet RPC functions are preferable. I'm going to merge this for now so devops can get a deployment working

lgalabru: stacks-blockchain is already using some wallet functions namely "listunspent" and "importaddress". I was also under the impression that the function `sendtoaddress` would be use, but don't have a strong opinion, and can add this additional call.

PR #88, opened 2020-05-07
fix: default import issues with ts-node

kyranjamie:


zone117x:
Currently fixed in master

PR #91, opened 2020-05-08
feat: paginate tx results

kyranjamie:


src/api/routes/tx.ts
```diff
@@ -3,28 +3,41 @@ import { addAsync, RouterWithAsync } from '@awaitjs/express';
...
 
 export function createTxRouter(db: DataStore): RouterWithAsync {
   const router = addAsync(express.Router());
 
   router.getAsync('/', async (req, res) => {
-    const txs = await db.getTxList();
+    const limit = parsePagingInput(req.query.limit.toString());
+    const offset = parsePagingInput(req.query.offset.toString());
+
+    const { results: txResults, total } = await db.getTxList({ offset, limit });
```
zone117x, line 28: I think we'll need an upper bound here to prevent massive requests from slowing down the db layer. Maybe like 200 is good for now? It will likely have to be decreased in production.

Also, if we make this endpoint only responsible for returning an array of tx_id strings, then we can have a much higher upper bound without putting much strain on the system. 

Thoughts?

zone117x: @aulneau if you describe how these list endpoints will be used, specific optimized queries can probably be crafted to serve that purpose

aulneau: This will be useful to get the latest transactions and display them on the explorer. 

It'd be great to be able to fetch `n` of tx_type of all transactions, eg `fetchContractCreationTxs(count)`.

> Also, if we make this endpoint only responsible for returning an array of tx_id strings, then we can have a much higher upper bound without putting much strain on the system.

Both would be ok, ideally we'd richer data, eg type, amount, addresses etc. I don't think there is a way to effectively fetch many tx's from the api, so it would be slow to do on the frontend.

zone117x: @aulneau could you create an issue for an API endpoint like that? Definitely sounds doable.

@kyranjamie Can you add a check here for limit < 200, which returns an error when exceeded?

src/datastore/postgres-store.ts
```diff
@@ -528,19 +528,25 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
     return { found: true, result: tx };
   }
 
-  async getTxList(count = 50) {
+  async getTxList({ limit, offset }: { limit: number; offset: number }) {
+    const total = await this.pool.query<TxQueryResult>(`
+      SELECT *
```
zone117x, line 533: This query will run into scalability problems very quickly. Should be something like `SELECT count(*) ...`

In general `SELECT *` should be avoided

kyranjamie: Done,

PR #106, opened 2020-05-13
Feat/logging

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/85 - Implement structured logging
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/86 - Implement 5xx response error log tagging

Also fixes a debug assertion failure that caused btc balance lookups to error while comparing floating point numbers.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -1,6 +1,7 @@
 import { execSync } from 'child_process';
 import * as dotenv from 'dotenv';
 import * as path from 'path';
+import * as winston from 'winston';
 
 export const isDevEnv = process.env.NODE_ENV === 'development';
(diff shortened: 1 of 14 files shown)


src/helpers.ts
```diff
@@ -96,13 +97,42 @@ export function loadDotEnv(): void {
...
+    winston.format.errors({ stack: true })
+  ),
+  transports: [
+    new winston.transports.Console({
+      handleExceptions: true,
+    }),
+  ],
+});
+
+export function logError(message: string, error?: Error) {
```
kyranjamie, line 121: is hoisted, but more readable if it were declared above use.

PR #108, opened 2020-05-13
Clarity values returned as repr strings

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/99 - Clarity values should be returned as repr strings

All API responses which previously returned a hex string for Clarity values now returns an object:
```ts
{
  hex: string, // the raw value as a hex string
  repr: string // the new repr string
}
```

Updated fields:
* asset value in post conditions for non-fungible tokens
* asset value in non-fungible token transfer events
* contract log value in smart contract log events
* function arg values in contract call transactions


kyranjamie:
So far I've been copying the `index.d.ts` to the docs folder and running `npm publish` from there.

PR #126, opened 2020-05-28
Include tx failed status code

zone117x:
Reference: https://github.com/blockstack/stacks-blockchain/pull/1612
Closes: https://github.com/blockstack/stacks-blockchain-sidecar/issues/97

The `tx_status` field is now one of the following values: `success`, `abort_by_response`, or `abort_by_post_condition`.

Note that `failed` is no longer a possible value. 

See https://github.com/blockstack/stacks-blockchain/pull/1612 for description of the abort value meanings.

### DIFF
--- a/docs/entities/transactions/transaction-status.schema.json
+++ b/docs/entities/transactions/transaction-status.schema.json
@@ -3,5 +3,5 @@
   "title": "TransactionStatus",
   "description": "All states a transaction can have",
   "type": "string",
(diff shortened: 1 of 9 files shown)


aulneau:
This looks good to me @zone117x, can you let me know when this will be deployed?

zone117x:
@aulneau this should be live on testnet now

PR #127, opened 2020-05-28
feat: ability to filter tx list by types

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/98

The recent transaction list API can use query param arrays to filter by type. The Express query param array formats are supported. Example usage:

* `/sidecar/v1/tx?type=token_transfer`
* `/sidecar/v1/tx?type=contract_call&type=smart_contract`
* `/sidecar/v1/tx?type[]=contract_call&type[]=smart_contract`

### DIFF
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -13,6 +13,7 @@ import {
   ContractCallTransaction,
   TransactionEvent,
   Block,
+  TransactionType,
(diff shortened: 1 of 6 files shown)


aulneau:
Would this show 200 of the type passed?

zone117x:
200 count? Yes the limit-offset params are applied after the filter

PR #128, opened 2020-05-29
fix: add missing offset-limit-total values in block-list response object

zone117x:


src/api/routes/block.ts
```diff
@@ -21,18 +22,18 @@ export function createBlockRouter(db: DataStore): RouterWithAsync {
...
-    const result = await Bluebird.mapSeries(blocks.results, async block => {
+    const results = await Bluebird.mapSeries(blocks, async block => {
       const blockQuery = await getBlockFromDataStore(block.block_hash, db);
       if (!blockQuery.found) {
         throw new Error('unexpected block not found -- fix block enumeration query');
       }
       return blockQuery.result;
     });
-
+    const response: BlockResults = { limit, offset, total, results };
```
kyranjamie, line 34: This could also implement the `validate` fn

zone117x: I'll leave this as a TODO for now, hopefully tackle later with a more integrated `openapi <-> express` solution

PR #132, opened 2020-06-03
Feat/stx addr api

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/113

This PR provides several stx address API endpoints which can be used to create a useful front-end page that displays address details:
* Balance (with total sent and received) for STX
* Balances for all fungible-tokens that the address has ever sent or received
* Ownership counts for all non-fungible-tokens that the address has ever sent or received
* List of latest transactions associated with the address, w/ pagination support
* List of latest asset events (transfers, mints, burns) associated with the address, w/ pagination support

The new endpoints are:
* `/sidecar/v1/address/<stx address>/balances` - STX, FT, and NFT balances/counts
* `/sidecar/v1/address/<stx address>/transactions` - recent transactions (limit-offset params supported)
* `/sidecar/v1/address/<stx address>/assets` - recent asset events (limit-offset params supported)

Note: the API schemas/types for the new responses are still a WIP.

This PR also fixes the critical bugs:
* Several fields displayed an incorrect stx address (the pubkeyhash was being double-hashed).
* Missing `recipient` address field in NFT and FT asset events.
* Regression in BTC and STX faucet where address could not be specified as form data.

### DIFF
--- a/src/api/routes/debug.ts
+++ b/src/api/routes/debug.ts
@@ -353,7 +353,7 @@ export function createDebugRouter(db: DataStore): RouterWithAsync {
   });
 
   router.postAsync('/faucet', async (req, res) => {
-    const address: string = `${req.query.address}` || `${req.body.address}`;
+    const address: string = req.query.address || req.body.addres;
     const ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress;
     const lastRequest = await db.getSTXFaucetRequest(address);
 
--- a/src/api/routes/debug.ts
+++ b/src/api/routes/debug.ts
@@ -337,7 +337,7 @@ export function createDebugRouter(db: DataStore): RouterWithAsync {
   });
 
   router.postAsync('/faucet', async (req, res) => {
-    const address: string = req.query.address || req.body.addres;
+    const address: string = req.query.address || req.body.address;
     const ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress;
     const lastRequest = await db.getSTXFaucetRequest(address);
 
--- a/src/api/routes/debug.ts
+++ b/src/api/routes/debug.ts
@@ -5,17 +5,12 @@ import { addAsync, RouterWithAsync } from '@awaitjs/express';
 import { htmlEscape } from 'escape-goat';
 import {
   makeSTXTokenTransfer,
-  TransactionVersion,
   makeSmartContractDeploy,
   PostConditionMode,
   makeContractCall,
   ClarityValue,
-  AddressHashMode,
-  addressHashModeToVersion,
-  addressFromPublicKeys,
-  addressToString,
-  pubKeyfromPrivKey,
   StacksTestnet,
+  getAddressFromPrivateKey,
 } from '@blockstack/stacks-transactions';
 import { SampleContracts } from '../../sample-data/broadcast-contract-default';
 import { DataStore, DbFaucetRequestCurrency } from '../../datastore/common';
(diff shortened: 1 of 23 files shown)


src/api/routes/debug.ts
```diff
@@ -353,7 +337,7 @@ export function createDebugRouter(db: DataStore): RouterWithAsync {
   });
 
   router.postAsync('/faucet', async (req, res) => {
-    const address: string = `${req.query.address}` || `${req.body.address}`;
+    const address: string = req.query.address || req.body.addres;
```
kyranjamie, line 340: spelling

src/event-stream/reader.ts
```diff
@@ -8,9 +7,28 @@ import { readTransaction, TransactionPayloadTypeID, RecipientPrincipalTypeId } f
...
 import { c32address } from 'c32check';
 
+export function getAddressFromPublicKeyHash(
+  publicKeyHash: Buffer,
+  transactionVersion: TransactionVersion
+): string {
+  const addrVer = addressHashModeToVersion(AddressHashMode.SerializeP2PKH, transactionVersion);
+  if (publicKeyHash.length !== 20) {
+    throw new Error('expected 20-byte pubkeyhash');
+  }
```
kyranjamie, line 26: arent bytes !== string length?

src/helpers.ts
```diff
@@ -133,6 +133,17 @@ export function logError(message: string, error?: Error) {
...
+export function formatMapToObject<TKey extends string, TValue, TFormatted>(
+  map: Map<TKey, TValue>,
+  formatter: (value: TValue) => TFormatted
+): Record<TKey, TFormatted> {
+  const obj = {} as Record<TKey, TFormatted>;
+  for (const [key, value] of map) {
+    obj[key] = formatter(value);
+  }
+  return obj;
+}
```
kyranjamie, line 145: 🏆

aulneau:
@zone117x looks good, but question: is it possible to have a `/sidecar/v1/address/<stx address>` that returns all data for an address? I'd prefer to not do three requests for each address on the explorer. Otherwise looks good!

zone117x:
@aulneau  Note these endpoints should only be fetched for one address at a time (e.g. a page for a specific address rather than list of addresses). If we want a page showing a list of addresses with data, we need to refine what to show because the current queries are relatively expensive. 

WRT a single endpoint -- we could do that, but how would the pagination work?

aulneau:
> @aulneau Note these endpoints should only be fetched for one address at a time (e.g. a page for a specific address rather than list of addresses). If we want a page showing a list of addresses with data, we need to refine what to show because the current queries are relatively expensive.

I think for now we don't need a page for a listing of addresses, but at some point we might, eg the rich list.


> WRT a single endpoint -- we could do that, but how would the pagination work?

Well I think it could go two ways:
- We have a single endpoint that combines all three, and we pass the same `limit` and `offset` as we would, but they apply to each three, or
- For each address page, there will be 3x fetches per load/refresh

I think generally it would make sense to use the initial fetch (of all three routes in one) and not paginate it, and then we can for each type use the individual route to fetch more. We won't want to fetch each if the user is only loading more from say transactions.

zone117x:
> I think generally it would make sense to use the initial fetch (of all three routes in one) and not paginate it, and then we can for each type use the individual route to fetch more. We won't want to fetch each if the user is only loading more from say transactions.

This seems reasonable to me. Okay if we open an issue to implement after this PR?

aulneau:
yeah def!

Issue #115, opened 2020-05-19
Websocket / socket.io for pushing tx status updates


zone117x:
Provide an endpoint that pushes notifications (i.e. websocket) to clients when a transaction status changes. 

For example, client opens a websocket connection to `/sidecar/v1/ws/tx/<txid>` and the server pushes json payloads when the tx status changes.

The lifecycle of a transaction is something like:
1. The node does not know about the transaction ID. This indicates that the transaction has not yet propagated to the node (it does not exist in the mempool). It could also indicate the transaction was rejected or never submitted, in which case there will never be a status change. The server cannot distinguish between either of these, so the former scenario should be assumed.
2. The node received the transaction and it enters the mempool. This is a `zero confirmation transaction`. 
3. The transaction is mined into a block and marked as either `aborted` or `successful`. 
4. The transaction's "confirmations" increase forever as every new block is mined. 
5. If a re-org occurs at or below the transaction's block height, then the tx is marked as `non-canonical`.

Note: at the time of writing, mempool changes are not yet observable so zero-conf tx updates are not yet possible.

A simplistic implementation using [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) exists at https://github.com/blockstack/stacks-blockchain-sidecar/blob/0dab05d187ea1c2492c41d65ed45aeacc4c3141c/src/api/routes/tx.ts#L55

The sidecar datastore implements an `EventEmitter` which can be consumed with code like:
```js
db.addListener('txUpdate', (tx) => { ... });
```

PR #145, opened 2020-06-16
feat: Transaction Updates via Websockets

reedrosenbluth:
This PR adds a websocket server that can be accessed via the endpoint `ws://<sidecar-address>/sidecar/v1`. Clients can use this endpoint to open a websocket connection to the sidecar. They can subscribe to transactions by sending a message containing just a `txid` to the server via the websocket connection. Once subscribed to a transaction, the server will stream updates to the client when the state of the transaction changes.

Example client usage with [ws](https://github.com/websockets/ws):

```javascript
const WebSocket = require('ws');

const ws = new WebSocket(`ws://localhost:3999/sidecar/v1`);

// subscribe to tx once connection is open
ws.on('open', function open() {
  ws.send('0x54da30065d7265347c43f5a6e9ec5f3a7409d802b0c5eb8e72735fec364fe075');
});

// log updates to the subscribed tx
ws.on('message', function incoming(data) {
  console.log(data);
});

```

Streaming Tx update messages take the following form:

```javascript
{
    txId: '0x54da30065d7265347c43f5a6e9ec5f3a7409d802b0c5eb8e72735fec364fe075',
    status: 'success'
}
```

Issue: #115 

Review by commit.

Considerations + Questions:
- I currently don't have a test for mempool tx updates. Not entirely sure how to implement this at the moment...
- I haven't yet implemented a way for clients to unsubscribe from transactions. Is this necessary?
- This websockets API is very simple right now (messages just contain txids) as there's only one possible action. Do we anticipate there being other use cases for websockets?

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -25,7 +25,10 @@ export interface ApiServer {
   address: string;
 }
 
-export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
+export async function startApiServer(
+  datastore: DataStore,
+  tx_subscribers: Map<string, Set<WebSocket>>
+): Promise<ApiServer> {
   const app = addAsync(express());
 
   const apiHost = process.env['STACKS_SIDECAR_API_HOST'];
@@ -94,15 +97,15 @@ export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
     })
   );
 
-  const dbTxUpdate = async (tx: DbTx): Promise<void> => {
-    if (tx_subscribers.has(tx.tx_id)) {
+  const dbTxUpdate = async (txId: string): Promise<void> => {
+    if (tx_subscribers.has(txId)) {
       try {
-        const txQuery = await getTxFromDataStore(tx.tx_id, datastore);
+        const txQuery = await getTxFromDataStore(txId, datastore);
         if (!txQuery.found) {
           throw new Error('error in tx stream, tx not found');
         }
         tx_subscribers
-          .get(tx.tx_id)
+          .get(txId)
           ?.forEach(subscriber =>
             sendWsTxUpdate(subscriber, txQuery.result.tx_id, txQuery.result.tx_status)
           );
@@ -113,9 +116,9 @@ export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
   };
 
   // EventEmitters don't like being passed Promise functions so wrap the async handler
-  const onTxUpdate = (tx: DbTx): void => {
+  const onTxUpdate = (txId: string): void => {
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    dbTxUpdate(tx);
+    dbTxUpdate(txId);
   };
 
   datastore.addListener('txUpdate', onTxUpdate);
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -1,4 +1,5 @@
 import { Server, createServer } from 'http';
+import { Socket } from 'net';
 import * as express from 'express';
 import * as expressWinston from 'express-winston';
 import { v4 as uuid } from 'uuid';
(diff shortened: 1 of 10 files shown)


src/tests/websocket-tests.ts
```diff
@@ -0,0 +1,129 @@
...
+        },
+      ],
+    };
+
+    // ws
+    const addr = apiServer.address;
+    const wsAddress = `ws://${addr}/sidecar/v1`;
+    // const ws = new WebSocket(wsAddress);
+    const wsp = new WebSocketAsPromised(wsAddress, {
+      // @ts-ignore
```
reedrosenbluth, line 99: wasn't able to get WebSocketAsPromised working without this `ts-ignore` line... Not ideal but this library makes this kind of testing a bit easier...

kyranjamie:
Looks good Reed. We used [WAMP](https://wamp-proto.org/) at Blockport which worked pretty well. Has a client, [Autobahn](https://github.com/crossbario/autobahn-js), that manages subscribing to multiple data streams from the same websocket.

src/event-stream/event-server.ts
```diff
@@ -41,6 +46,10 @@ async function handleMempoolTxsMessage(rawTxs: string[], db: DataStore): Promise
     };
   });
   for (const tx of decodedTxs) {
+    // Send update to websocket subscribers
+    tx_subscribers
```
zone117x, line 50: Client websocket related code shouldn't be necessary in this file. Tx status updates should be sent to clients after the db layer has completed processing. The `datastore.addListener('txUpdate', ...)` hook emits mempool updates, so this can just be removed.

src/api/init.ts
```diff
@@ -89,9 +95,58 @@ export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
...
+  };
+
+  datastore.addListener('txUpdate', onTxUpdate);
+
+  let server = createServer(app);
+  const wss = new WebSocket.Server({ server, path: '/sidecar/v1' });
+  wss.on('connection', function (ws) {
+    ws.on('message', txid => {
+      const id = txid.toString();
+      const connections = tx_subscribers.get(id);
```
zone117x, line 129: Nit: snake_case is only used to match JSON schemas and sql names, the rest should all be camelCase

zone117x:
> I currently don't have a test for mempool tx updates. Not entirely sure how to implement this at the moment...

The function `db.updateMempoolTx(...)` will emit a mempool tx updated -- it can be used the same way the `db.update(...)` function is currently used in tests

> I haven't yet implemented a way for clients to unsubscribe from transactions. Is this necessary?

Sorta related -- I think certain page transitions happen client-side in a way that could leave the websocket connection dangling. I assume that needs explicitly handled client-side by either closing the websocket or implementing an unsubscribe feature? @aulneau thoughts?

> This websockets API is very simple right now (messages just contain txids) as there's only one possible action. Do we anticipate there being other use cases for websockets?

I think in the near future we'd want a websocket channel for pushing new blocks and new transactions. I think if the payloads used objects similar to a json-rpc request (rather than just the txid string for example), then it would be easier to extend in the future.

kyranjamie:
Out of scope for this PR, but curious if there are plans to, or how much work it'd be, to subscribe to transactions of a given STX address. A WS companion of `/sidecar/v1/address/<address>/transactions`.

Asking [in context of the Stacks Wallet](https://www.figma.com/file/VcXbxT1EwXvE1nUlOiBEES/Wallet?node-id=442%3A1839)

src/event-stream/event-server.ts
```diff
@@ -41,6 +46,10 @@ async function handleMempoolTxsMessage(rawTxs: string[], db: DataStore): Promise
     };
   });
   for (const tx of decodedTxs) {
+    // Send update to websocket subscribers
+    txSubscribers
```
zone117x, line 50: (looks like this comment was removed after a rebase) the websocket code here should be removed, it's more appropriately handled in the db event emitter hooks

reedrosenbluth: my bad, thought i'd already done this! removing...

PR #147, opened 2020-06-19
feat: search endpoint for blocks, txs, and principals

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/114

This PR implements a new API endpoint: `GET /sidecar/v1/search/<term>`

The `term` can currently be a block hash, txid, standard account address, or contract address. 

This endpoint is designed for efficiency, for example, used with a search box that performs aggressive auto-complete queries on input change. This should be used as an alternative to the client performing something like `await Promise.any([ fetchBlock(term),  fetchTx(term), fetchAddr(term) ])`.

The following normalization is automatically applied:
* The `term` string whitespace is trimmed.
* Hex strings can be with or without the `0x` prefix, and are case-insensitive.

The response for a found entity is a 200 with:
```ts
{
  found: true;
  result: {
    entity_type: 
      | 'standard_address' 
      | 'contract_address' 
      | 'block_hash' 
      | 'tx_id' 
      | 'mempool_tx_id';
    // The normalized string for use with other endpoints
    entity_id: string;
  }
}
```

The result for a mined (non-mempool) transaction (`tx_id`), or a `contract_address` associated with a tx contains the additional fields:
```ts
{
  result: {
    entity_type: 'tx_id';
    tx_data: {
      canonical: boolean;
      block_hash: string;
      burn_block_time: number;
      block_height: number;
      tx_type: string;
    }
  }
}
```

The result for a mempool transaction (`mempool_tx_id`), or a `contract_address` associated with a mempool tx contains the additional fields:
```ts
{
  result: {
    entity_type: 'mempool_tx_id';
    tx_data: {
      tx_type: string;
    }
  }
}
```

The result for a block contains the additional fields:
```ts
  result: {
    entity_type: 'block_hash';
    block_data: {
      canonical: boolean;
      hash: string;
      parent_block_hash: string;
      burn_block_time: number;
      height: number;
    }
}
```

Response for not found is a 404 with:
```ts
{
  found: false;
  result: {
    entity_type:
      // Syntactically valid STX address, but not found in database.
      | 'standard_address'
      // Syntactically valid contract address, but not found in database.
      | 'contract_address'
      // Syntactically valid block hash or tx_id, but not found in database.
      | 'unknown_hash' 
      // The given term is not syntactically valid for any supported search types.
      | 'invalid_term' 
  };
  error: string;
}
```

Note: a given principal (account or contract address) must exist in the database as the `sender` or `recipient` in a transaction, asset transfer event, or contract-deploy ID -- otherwise a 404 will be returned even if the principal is syntactically valid.

## Examples:

* Search by txid:

`/sidecar/v1/search/0x4b8d094151a118ee7b62b6e215be69a5b9fb7314b02ec624855dd519ca1972b4`
```js
{
  found: true,
  result: {
    entity_type: 'tx_id',
    entity_id: '0x4b8d094151a118ee7b62b6e215be69a5b9fb7314b02ec624855dd519ca1972b4',
  }
}
```

* Search by txid (without 0x prefix):

`/sidecar/v1/search/4b8d094151a118ee7b62b6e215be69a5b9fb7314b02ec624855dd519ca1972b4`
```js
{
  found: true,
  result: {
    entity_type: 'tx_id',
    entity_id: '0x4b8d094151a118ee7b62b6e215be69a5b9fb7314b02ec624855dd519ca1972b4',
  }
}
```

* Search by txid (when tx is in the mempool):

`/sidecar/v1/search/4b8d094151a118ee7b62b6e215be69a5b9fb7314b02ec624855dd519ca1972b4`
```js
{
  found: true,
  result: {
    entity_type: 'mempool_tx_id',
    entity_id: '0x4b8d094151a118ee7b62b6e215be69a5b9fb7314b02ec624855dd519ca1972b4',
  }
}
```

* Search by block hash:

`/sidecar/v1/search/0x7d1e52e99f43d1454b1838e54a1d4287edb116e5d1e3410a9490af26a6479a19`
```js
{
  found: true,
  result: {
    entity_type: 'block_hash',
    entity_id: '0x7d1e52e99f43d1454b1838e54a1d4287edb116e5d1e3410a9490af26a6479a19',
  }
}
```

* Search by standard account address:

`/sidecar/v1/search/ST3J8EVYHVKH6XXPD61EE8XEHW4Y2K83861225AB1`
```js
{
  found: true,
  result: {
    entity_type: 'standard_address',
    entity_id: 'ST3J8EVYHVKH6XXPD61EE8XEHW4Y2K83861225AB1',
  }
}
```

* Search by contract address:

`/sidecar/v1/search/STSPS4JYDEYCPPCSHE3MM2NCEGR07KPBETNEZCBQ.test-contract`
```js
{
  found: true,
  result: {
    entity_type: 'contract_address',
    entity_id: 'STSPS4JYDEYCPPCSHE3MM2NCEGR07KPBETNEZCBQ.test-contract',
  }
}
```

* Search by standard account address with extra whitespace:

`/sidecar/v1/search/  ST3J8EVYHVKH6XXPD61EE8XEHW4Y2K83861225AB1  `
```js
{
  found: true,
  result: {
    entity_type: 'standard_address',
    entity_id: 'ST3J8EVYHVKH6XXPD61EE8XEHW4Y2K83861225AB1',
  }
}
```

### DIFF
--- a/src/api/routes/search.ts
+++ b/src/api/routes/search.ts
@@ -3,12 +3,31 @@ import { addAsync, RouterWithAsync } from '@awaitjs/express';
 import { DataStore } from '../../datastore/common';
 import { isValidPrincipal } from '../../helpers';
 
+type SearchResult =
+  | {
+      found: false;
+      result: {
+        entity_type: 'standard_address' | 'contract_address' | 'unknown_hash' | 'invalid_term';
+      };
+      error: string;
+    }
+  | {
+      found: true;
+      result: {
+        entity_type:
+          | 'standard_address'
+          | 'contract_address'
+          | 'block_hash'
+          | 'tx_id'
+          | 'mempool_tx_id';
+        entity_id: string;
+      };
+    };
+
 export function createSearchRouter(db: DataStore): RouterWithAsync {
   const router = addAsync(express.Router());
-  router.getAsync('/:term', async (req, res) => {
-    const { term: rawTerm } = req.params;
-    const term = rawTerm.trim();
 
+  const performSearch = async (term: string): Promise<SearchResult> => {
     // Check if term is a 32-byte hash, e.g.:
     //   `0x4ac9b89ec7f2a0ca3b4399888904f171d7bdf3460b1c63ea86c28a83c2feaad8`
     //   `4ac9b89ec7f2a0ca3b4399888904f171d7bdf3460b1c63ea86c28a83c2feaad8`
@@ -22,23 +41,54 @@ export function createSearchRouter(db: DataStore): RouterWithAsync {
       const hash = '0x' + hashBuffer.toString('hex');
       const hashQueryResult = await db.searchHash({ hash });
       if (hashQueryResult.found) {
-        res.json(hashQueryResult);
-        return;
+        return hashQueryResult;
+      } else {
+        return {
+          found: false,
+          result: { entity_type: 'unknown_hash' },
+          error: `No block or transaction found with hash "${hash}"`,
+        };
       }
     }
 
     // Check if term is an account or contract principal address, e.g.:
     //   `ST3DQ94YDRH07GRKTCNN5FTW962ACVADVJZD7GSK3`
     //   `ST2TJRHDHMYBQ417HFB0BDX430TQA5PXRX6495G1V.contract-name`
-    if (isValidPrincipal(term)) {
+    const principalCheck = isValidPrincipal(term);
+    if (principalCheck) {
       const principalQueryResult = await db.searchPrincipal({ principal: term });
       if (principalQueryResult.found) {
-        res.json(principalQueryResult);
-        return;
+        return principalQueryResult;
+      } else {
+        return {
+          found: false,
+          result: {
+            entity_type:
+              principalCheck.type === 'contractAddress' ? 'contract_address' : 'standard_address',
+          },
+          error: `No principal found with address "${term}"`,
+        };
       }
     }
 
-    res.status(404).json({ found: false, error: `cannot find entity by "${term}"` });
+    return {
+      found: false,
+      result: {
+        entity_type: 'invalid_term',
+      },
+      error: `The term "${term}" is not a valid block hash, transaction ID, contract principal, or account address principal`,
+    };
+  };
+
+  router.getAsync('/:term', async (req, res) => {
+    const { term: rawTerm } = req.params;
+    const term = rawTerm.trim();
+
+    const searchResult = await performSearch(term);
+    if (!searchResult.found) {
+      res.status(404);
+    }
+    res.json(searchResult);
   });
 
   return router;
--- a/src/api/routes/search.ts
+++ b/src/api/routes/search.ts
@@ -1,0 +1,45 @@
+import * as express from 'express';
+import { addAsync, RouterWithAsync } from '@awaitjs/express';
+import { DataStore } from '../../datastore/common';
+import { isValidPrincipal } from '../../helpers';
+
+export function createSearchRouter(db: DataStore): RouterWithAsync {
+  const router = addAsync(express.Router());
+  router.getAsync('/:term', async (req, res) => {
+    const { term: rawTerm } = req.params;
+    const term = rawTerm.trim();
+
+    // Check if term is a 32-byte hash, e.g.:
+    //   `0x4ac9b89ec7f2a0ca3b4399888904f171d7bdf3460b1c63ea86c28a83c2feaad8`
+    //   `4ac9b89ec7f2a0ca3b4399888904f171d7bdf3460b1c63ea86c28a83c2feaad8`
+    let hashBuffer: Buffer | undefined;
+    if (term.length === 66 && term.toLowerCase().startsWith('0x')) {
+      hashBuffer = Buffer.from(term.slice(2), 'hex');
+    } else if (term.length === 64) {
+      hashBuffer = Buffer.from(term, 'hex');
+    }
+    if (hashBuffer !== undefined && hashBuffer.length === 32) {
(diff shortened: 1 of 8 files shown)


zone117x:
@aulneau There are some additional fields that can be returned with essentially zero additional server cost -- for example a `entity_type: 'mempool_tx_id' | 'tx_id'` response could include the tx-type. Just let me know what additional fields might be useful.

src/api/routes/search.ts
```diff
@@ -0,0 +1,45 @@
...
+  const router = addAsync(express.Router());
+  router.getAsync('/:term', async (req, res) => {
+    const { term: rawTerm } = req.params;
+    const term = rawTerm.trim();
+
+    // Check if term is a 32-byte hash, e.g.:
+    //   `0x4ac9b89ec7f2a0ca3b4399888904f171d7bdf3460b1c63ea86c28a83c2feaad8`
+    //   `4ac9b89ec7f2a0ca3b4399888904f171d7bdf3460b1c63ea86c28a83c2feaad8`
+    let hashBuffer: Buffer | undefined;
+    if (term.length === 66 && term.toLowerCase().startsWith('0x')) {
```
kyranjamie, line 16: What about my handy `has0xPrefix` function lol

src/datastore/memory-store.ts
```diff
@@ -303,6 +304,18 @@ export class MemoryDataStore extends (EventEmitter as { new (): DataStoreEventEm
     throw new Error('not yet implemented');
   }
 
+  searchHash(args: {
+    hash: string;
+  }): Promise<{ found: false } | { found: true; result: DbSearchResult }> {
```
kyranjamie, line 309: Loads of these `{found : false } | {found: true, result: X}` types. Could make a generic:

```
type FoundOrNot<T> = { found: false } | { found: true; result: T }
```

src/api/routes/search.ts
```diff
@@ -0,0 +1,45 @@
...
+      hashBuffer = Buffer.from(term.slice(2), 'hex');
+    } else if (term.length === 64) {
+      hashBuffer = Buffer.from(term, 'hex');
+    }
+    if (hashBuffer !== undefined && hashBuffer.length === 32) {
```
hstove, line 21: Is it possible that a valid contract principle could trigger this, when it shouldn't?

src/datastore/postgres-store.ts
```diff
@@ -1607,6 +1608,129 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
+  }
+
+  async searchPrincipal({
+    principal,
+  }: {
+    principal: string;
+  }): Promise<{ found: false } | { found: true; result: DbSearchResult }> {
+    const isContract = principal.includes('.');
+    const entityType = isContract ? 'contract_address' : 'standard_address';
+    const successResult: DbSearchResult = {
```
hstove, line 1664: Maybe just have this variable be `{ found: true, result: { entityType, entityId }`. Then you could just `return successResult` instead of the expanded format for each query.

src/datastore/postgres-store.ts
```diff
@@ -1607,6 +1608,129 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
     return { results: parsed, total: count };
   }
 
+  async searchHash({
```
hstove, line 1611: Especially since we're going for speed, would we want to perform each of these queries in parallel, instead of serially?

Same for the principal search.

src/api/routes/search.ts
zone117x: I don't think so. At the very minimum, a contract principal has a `.` character (not valid hexadecimal).

src/datastore/postgres-store.ts
zone117x: Hmm. The downside is that only one query needs to succeed in finding a match. Which means that unless the the last query happens to be the only match, all the parallel queries would likely slow down the single query that ends up finding the match. This of course is dependent on how the managed postgres service is configured (for example, if clustering of some kind is setup that might negate this penalty).

hstove: I think the infra would support many parallel queries by default, so I don't think it would have much of an effect. Also, the client code already uses a connection pool.

> The downside is that only one query needs to succeed in finding a match

Yeah, but you could still use the same prioritization logic you have here to return one result.

I think the way you have it right now is fine, because it's already prioritized in a way that makes sense. Way more people will look for a TXID than a block hash.

src/datastore/memory-store.ts
zone117x: Great idea 👍 
That change will touch a bunch of files, will do in a follow up PR

src/api/routes/search.ts
zone117x: Fixed

src/datastore/postgres-store.ts
zone117x: I think this is really dependent on how the managed pg instance is configured. For example, the pool is currently using the default `10` for max connections, let's say the managed pg instance is clustered to handle 10 parallel queries in a way that doesn't slow any of them down.

In this situation, we'd see performance gains until usage gets to a point where there are often more than 10 parallel queries at a given time. For example, a handful of users doing a search, a handful loading other pages, etc. Then you'd have the 11th and greater queries pending -- slowed down due to the several queries that were not strictly necessary. 

The number of "free" parallel queries might be _much_ higher than 10. My reservations are based on assumptions about the managed pg instance and future usage rates, I could definitely be off on those -- I need to do some research into the available parallel capacity. I'll make them parallel if you feel strongly about it?

hstove: I don't feel strongly about it! I'm definitely good with keeping it as-is.

PR #159, opened 2020-07-07
feat: contract event enumeration endpoint

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/150
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/141
Closes https://github.com/blockstack/stacks-blockchain-sidecar/issues/148

This PR creates a new endpoint that returns the contract-log-events emitted on a given contract:
`GET /sidecar/v1/contract/<contract_id>/events` 

* `limit` and `offset` query params supported.
* Ordered by most recent -- DESC block height, tx index, event index.

Example: `/sidecar/v1/contract/hello-world/events?limit=20&offset=0`

```ts
{
  limit: 20,
  offset: 0,
  results: [
    {
      event_index: 4,
      event_type: 'smart_contract_log',
      contract_log: {
        contract_id: 'hello-world',
        topic: 'print', // note that the Clarity `print` statement is the only possibly event topic for now
        value: { hex: '0x0200000008736f6d652076616c', repr: '"some val"' },
      },
    },
  ],
}
```

----

* This is currently returning event objects matching the existing json schema for `TransactionEventSmartContractLog` (sub-type of `TransactionEvent`). Should the entries be extended to include more info like the txid?

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -287,6 +287,8 @@ export function assertNotNullish<T>(val: T, onNullish?: () => string): Exclude<T
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type ElementType<T extends any[]> = T extends (infer U)[] ? U : never;
 
+export type FoundOrNot<T> = { found: true; result: T } | { found: false };
+
 export function timeout(ms: number): Promise<void> {
   return new Promise(resolve => {
     setTimeout(() => {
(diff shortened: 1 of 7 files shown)


src/helpers.ts
```diff
@@ -287,6 +287,8 @@ export function assertNotNullish<T>(val: T, onNullish?: () => string): Exclude<T
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type ElementType<T extends any[]> = T extends (infer U)[] ? U : never;
 
+export type FoundOrNot<T> = { found: true; result: T } | { found: false };
```
kyranjamie, line 290: 🆒

kyranjamie: You can also proudly close https://github.com/blockstack/stacks-blockchain-sidecar/issues/148

PR #163, opened 2020-07-09
Fix address endpoint missing contract-calls and contract-deploys txs

zone117x:
Adds missing `contract-call` and `contract-deploy` txs to the `/address/:contract/transactions` endpoint response. 
Thx @hstove for finding this bug.

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -1653,7 +1653,12 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
       WITH transactions AS (
         SELECT *, (COUNT(*) OVER())::integer as count
         FROM txs
-        WHERE canonical = true AND (sender_address = $1 OR token_transfer_recipient_address = $1)
+        WHERE canonical = true AND (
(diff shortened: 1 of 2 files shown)


hstove:
I'm using this locally, and the first contract-call TX shows up correctly in this API, but any subsequent ones do not. Still investigating why that might be.

zone117x:
@hstove 
> Edit: The newer TXs come into the database, but the `canonical` column is false. What does that column mean?

`canonical=false` means the transaction was included in a block that was reorged (forked) and is no longer part of the longest/canonical chaintip. How are you running your stacks node?

hstove:
Here's my local PG database - I have more recent TX's on the same chain, but the more recent ones have `canonical=false`.

image

> How are you running your stacks node?

Sidecar running with `npm run dev:watch`, and I'm compiling & running the Stacks node manually, with a slight change to the TOML file to dispatch events. Mocknet.

> Can you reproduce?

Yes. In fact, I just spun up a local node again, and redeployed & used my contract. The new TX is in the database, but `canonical = false`. Here is a look at my PG database:

PR #176, opened 2020-07-21
Support sponsored transactions

zone117x:
Depends on https://github.com/blockstack/stacks-transactions-js/pull/99

This PR updates the sidecar to support processing and returning sponsored tx data. 

### Testing

The `/sidecar/v1/debug/broadcast` endpoints now have a checkbox to create and broadcast a sponsored transaction. 

For example: 
* Visit `http://localhost:3999/sidecar/v1/debug/broadcast/contract-deploy`
* Check `[x] Create sponsored transaction` and click Submit
* The tx response will then contain sponsored tx info, for example: 
```js
// localhost:3999/sidecar/v1/tx/0x2537c7090545c01db3f86300efb75b72e07695c6d4bf5358d218242d2da859a1
{
  ...
  "sponsored": true,
  "sponsor_address": "STRYYQQ9M8KAF4NS7WNZQYY59X93XEKR31JP64CP",
  ...
}
```

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -100,7 +100,7 @@ export async function cycleMigrations(): Promise<void> {
 const TX_COLUMNS = `
   -- required columns
   tx_id, raw_tx, tx_index, index_block_hash, block_hash, block_height, burn_block_time, type_id, status, 
-  canonical, post_conditions, fee_rate, sponsored, sender_address, origin_hash_mode,
+  canonical, post_conditions, fee_rate, sponsored, sponsor_address, sender_address, origin_hash_mode,
 
   -- token-transfer tx columns
   token_transfer_recipient_address, token_transfer_amount, token_transfer_memo,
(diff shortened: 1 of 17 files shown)


src/datastore/postgres-store.ts
```diff
@@ -816,7 +818,7 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
       `
       INSERT INTO txs(
         ${TX_COLUMNS}
-      ) values($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25, $26, $27)
+      ) values($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25, $26, $27, $28)
```
kyranjamie, line 821: Is there no way to do this without having to manually type out a series of numbers? Seems like a hassle 😅

zone117x: You'd think the sql driver would support a spread operator or something (maybe it does and I haven't found it idk). 

At this point, probably makes sense to make a helper function that generates this string based off the array of input values.

PR #177, opened 2020-07-23
Fix/account balance require props

kyranjamie:
- `07e18fc` fix(docs): required props, dictionary for fts, nfts
- `f079167` fix(docs): conform to 'Response' naming convention


docs/index.d.ts
```diff
@@ -47,7 +67,7 @@ export interface AccountTransactions {
 /**
  * GET request that returns blocks
  */
-export interface BlockResults {
+export interface BlockResultsResponse {
```
agraebe, line 70: Maybe this could be `BlockListResponse`?

PR #178, opened 2020-07-24
Add gitpod support

zone117x:
Reference issue: https://github.com/blockstack/stacks-blockchain-sidecar/issues/166

Test this PR branch on gitpod using this link: https://gitpod.io/#https://github.com/blockstack/stacks-blockchain-sidecar/tree/feat/gitpod

* Gitpod environment:
  * Postgres launched in the background
  * The core-node is ran in a split terminal
  * `npm run dev` as main command
* Updated the root path `/` from a 404 to a redirect to `/sidecar/v1/status`


.gitpod.yml
```diff
@@ -0,0 +1,15 @@
...
+
+ports:
+  # Sidecar API port
+  - port: 3999
+    onOpen: open-preview
+  - port: 1000-65535
+    onOpen: ignore
+
+tasks:
+  - init: npm install
```
agraebe, line 12: it took me 3+ minutes to get install done. is there a way to get this done on the docker level? So it would happen only once and everyone opening the pod will be all set?

.gitpod.yml
```diff
@@ -0,0 +1,15 @@
+image:
+  file: .gitpod.Dockerfile
+
+ports:
+  # Sidecar API port
+  - port: 3999
```
agraebe, line 6: I'm getting an error when I access code node endpoints (e.g. `/v2/info`). it says "Port 3999 didn't respond"... Any idea why?

readme.md
```diff
@@ -2,6 +2,8 @@
 
 [![Build Status](https://img.shields.io/endpoint.svg?url=https%3A%2F%2Factions-badge.atrox.dev%2Fblockstack%2Fstacks-blockchain-sidecar%2Fbadge%3Fref%3Dmaster&style=flat)](https://actions-badge.atrox.dev/blockstack/stacks-blockchain-sidecar/goto?ref=master)
 
+[![Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/blockstack/stacks-blockchain-sidecar)
```
agraebe, line 5: I'm also getting an error running the faucet. Is there some hardcoded base URL in our codebase?

Screen Shot 2020-07-24 at 8 45 42 AM

.gitpod.yml
zone117x: We can enable gitpod integration into the repo for "prebuilt workspace" https://www.gitpod.io/docs/prebuilds/

.gitpod.yml
zone117x: Should be fixed now. Too many environmental variables 😅

readme.md
zone117x: Fixed

.gitpod.yml
zone117x: I just enabled the prebuild feature. Kinda funky but seems faster

zone117x: (5 hours later) clicked the gitpod link and it wasn't using a prebuilt workspace like it did initially, maybe it doesn't stay cached unless on master? No idea, not much documentation on their end.

agraebe: Let's just move ahead for now. Can look into that when we see it used more and someone complaining about it :)

Issue #166, opened 2020-07-14
Add GitPod one-click action to readme


zone117x:

agraebe:
As mentioned in the heroku PR, we should drop that requirement. I believe the ROI for that is too low right now. Let's focus on the gitpod only.

PR #179, opened 2020-07-24
[wip] feat: heroku deploy

agraebe:
irt #166

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: node ./lib/index.js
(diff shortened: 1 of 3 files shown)


agraebe:
After some brief research, I think we shouldn't focus on this for the following reasons:
- we need to use a container to deploy on heroku
- the one-click deploy feature doesn't support container images [_anymore_](https://devcenter.heroku.com/articles/app-json-schema#image)

The alternative would be to tailor our codebase to make this possible, which is likely not trivial. The gitpod image should be good enough for developers to quickly spin up an instance and interact with it.

PR #183, opened 2020-07-28
fix: make address stx balance take fees into account

zone117x:
Fixed bug where address STX balance value was not taking fees into account.

Also added a new API for STX-only wallets which doesn't perform queries on FTs/NFTs: `/extended/v1/address/{principal}/stx`

### DIFF
--- a/docs/api/address/get-address-stx-balance.schema.json
+++ b/docs/api/address/get-address-stx-balance.schema.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "GET request that returns address balances",
+  "title": "AddressStxBalanceResponse",
+  "type": "object",
+  "required": ["stx", "fungible_tokens", "non_fungible_tokens"],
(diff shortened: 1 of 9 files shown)


docs/api/address/get-address-stx-balance.schema.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "GET request that returns address balances",
+  "title": "AddressStxBalanceResponse",
+  "type": "object",
+  "required": ["stx", "fungible_tokens", "non_fungible_tokens"],
```
kyranjamie, line 6: This doesn't quite work. The latter 2 aren't required for the stx balance, right? and also the emitted type comes out as non-required

kyranjamie: Better to require it the other way around. Make this it's own entity `stx-balance.schema.json`. and use an `allOf` in balance to merge a `$ref` to this, and the additional ft, nft schemas.

PR #195, opened 2020-08-21
rosetta api: base

fariedt:
This PR contains the implementation for two Rosetta API endpoints -- `/network/list` and `/network/options` -- and stubs for the others.

`/network/status` is not included in this PR because it requires several changes to the datastore interface, and to the postgres and memory store implementations.

Type information is not present because that requires a large number of entity/schema files.  That's in the next pull request.

The rosetta config files are included, and `rosetta-cli --configuration-file rosetta-cli-config/rosetta-config.json view:networks` can test the implemented endpoints.

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -17,10 +17,10 @@ import { createBlockRouter } from './routes/block';
 import { createFaucetRouter } from './routes/faucets';
 import { createAddressRouter } from './routes/address';
 import { createSearchRouter } from './routes/search';
-import { createRNetworkRouter } from './routes/rosetta/network';
-import { createRMempoolRouter } from './routes/rosetta/mempool';
-import { createRBlockRouter } from './routes/rosetta/block';
-import { createRAccountRouter } from './routes/rosetta/account';
+import { createRosettaNetworkRouter } from './routes/rosetta/network';
+import { createRosettaMempoolRouter } from './routes/rosetta/mempool';
+import { createRosettaBlockRouter } from './routes/rosetta/block';
+import { createRosettaAccountRouter } from './routes/rosetta/account';
 import { logger } from '../helpers';
 import { createWsRpcRouter } from './routes/ws-rpc';
 
@@ -91,10 +91,10 @@ export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
     (() => {
       const router = addAsync(express.Router());
       router.use(cors());
-      router.use('/network', createRNetworkRouter(datastore));
-      router.use('/mempool', createRMempoolRouter(datastore));
-      router.use('/block', createRBlockRouter(datastore));
-      router.use('/account', createRAccountRouter(datastore));
+      router.use('/network', createRosettaNetworkRouter(datastore));
+      router.use('/mempool', createRosettaMempoolRouter(datastore));
+      router.use('/block', createRosettaBlockRouter(datastore));
+      router.use('/account', createRosettaAccountRouter(datastore));
       return router;
     })()
   );
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -17,6 +17,10 @@ import { createBlockRouter } from './routes/block';
 import { createFaucetRouter } from './routes/faucets';
 import { createAddressRouter } from './routes/address';
 import { createSearchRouter } from './routes/search';
+import { createRNetworkRouter } from './routes/rosetta/network';
+import { createRMempoolRouter } from './routes/rosetta/mempool';
+import { createRBlockRouter } from './routes/rosetta/block';
+import { createRAccountRouter } from './routes/rosetta/account';
 import { logger } from '../helpers';
 import { createWsRpcRouter } from './routes/ws-rpc';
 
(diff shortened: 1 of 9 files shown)


src/api/routes/rosetta/network.ts
```diff
@@ -0,0 +1,50 @@
...
+  router.post('/status', (req, res) => {
+    res.json({ status: 'ready' });
+  });
+
+  router.post('/options', (_req, res) => {
+    const response = {
+      version: {
+        rosetta_version: RosettaConstants.rosettaVersion,
+        node_version: process.version,
+        middleware_version: process.env.npm_package_version,
```
zone117x, line 36: I don't think this env var is available if the compiled js is ran via `node lib/index.js`. 
Maybe better to use the `require('./package.json').version` technique?

src/api/init.ts
```diff
@@ -79,6 +83,26 @@ export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
   // Setup direct proxy to core-node RPC endpoints (/v2)
   app.use('/v2', createCoreNodeRpcProxyRouter());
 
+  // Rosetta API -- https://www.rosetta-api.org
+  app.use(
```
zone117x, line 87: Great job keeping this code consistent 💯

zone117x:
Perhaps we should add the `rosetta-cli --configuration-file rosetta-cli-config/rosetta-config.json ...` command as a github action? The `Setup integration environment` action might be best place to add because it already does the work to boostrap all the required services.

zone117x:
@fariedt 
> /network/status is not included in this PR because it requires several changes to the datastore interface, and to the postgres and memory store implementations.

FYI, the MemoryStore is essentially deprecated -- haven't gotten around to remove it yet. It was originally implemented as a stopgap while the postgres datastore was being implemented. 

Don't worry about getting the memory store working (and feel free to submit a PR that removes it if you feel inclined, no worries).

diwakergupta:
@fariedt can you include testing instructions in the PR -- basically steps to test this PR locally? Thanks!

src/api/init.ts
```diff
@@ -79,6 +83,26 @@ export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
...
+    '/rosetta/v1',
+    (() => {
+      const router = addAsync(express.Router());
+      router.use(cors());
+      router.use('/network', createRNetworkRouter(datastore));
```
zone117x, line 92: nit: I prefer more explicit naming e.g. `createRNetworkRouter` vs `createRosettaNetworkRouter`. Thoughts?

fariedt: The [original commit](https://github.com/fdmirza/stacks-blockchain-rosetta-api/commit/47db1a61cf6bd014a42c39952af3e9e3f3651f34#diff-f0977764d44e25489d36496dcb219a87) for this had the longer name, but was not consistent.  I shortened it for this PR; I can change them all to say `Rosetta` instead of `R` everywhere.

fariedt:
Setting up to test these calls:

- start the docker containers: `npm run devenv:deploy`
- start the api server: `npm run dev` or `npm run build && node lib/index.js`

The network list call:
```
curl -s -d '' http://localhost:3999/rosetta/v1/network/list | jq .
```
It will return something like
```
{
  "network_identifiers": [
    {
      "blockchain": "stacks",
      "network": "testnet"
    }
  ]
}
```

The network options call:
```
curl -s -d '' http://localhost:3999/rosetta/v1/network/options | jq .
```
This call (and all the other Rosetta calls) should take a network identifier in the request body and validate that; that's waiting on the next PR (#196).

fariedt:
I can have someone look into calling `rosetta-cli` in the github setup action.  It's implemented in Go, and currently requires Go 1.13.x; installing it with `go get ...` doesn't work with more recent versions: [issue](https://github.com/coinbase/rosetta-cli/issues/116).

zone117x:
@CharlieC3 can we remove `semantic-release-bot` from the CLAassistant checks?

@fariedt looks like there's a problem with the rebase on this PR

CharlieC3:
> @CharlieC3 can we remove semantic-release-bot from the CLAassistant checks?

@zone117x I don't think so, but this is also an issue we shouldn't be facing via a rebase. I would have expected to see only new commits from this feature branch being merged into the master. Perhaps changes from master will _pulled_ into the feature branch instead of rebased?

diwakergupta:
> > @CharlieC3 can we remove semantic-release-bot from the CLAassistant checks?
> 
> @zone117x I don't think so, but this is also an issue we shouldn't be facing via a rebase. I would have expected to see only new commits from this feature branch being merged into the master. Perhaps changes from master will _pulled_ into the feature branch instead of rebased?

@CharlieC3 actually you should able to configure CLA assistant to ignore `semantic-release-bot`, especially if we expect this bot to ever open PRs. Just head on over to https://cla-assistant.io/ and login with Github.

CharlieC3:
@diwakergupta Good to know, however when I log in it doesn't show the CLA assistant is registered with any repositories, so there's nothing for me to configure. It doesn't seem attached to our devops account either, so I'm not sure whose account with which it was originally configured.

Also, I would suggest we fix the rebase issues instead of merging this, as it can cause merge conflicts with other branches down the line.

PR #205, opened 2020-08-26
feat: add mempool openapi docs

kyranjamie:
Adds OpenAPI section for mempool transactions.

### DIFF
--- a/docs/api/transaction/get-mempool-transactions.example.json
+++ b/docs/api/transaction/get-mempool-transactions.example.json
@@ -1,0 +1,87 @@
+{
+  "limit": 96,
+  "offset": 0,
+  "total": 5,
+  "results": [
+    {
+      "tx_id": "0xb31df5a363dad31723324cb5e0eefa04d491519fd30827a521cbc830114aa50c",
+      "tx_status": "pending",
+      "tx_type": "token_transfer",
+      "receipt_time": 1598288370,
+      "receipt_time_iso": "2020-08-24T16:59:30.000Z",
+      "fee_rate": "180",
+      "sender_address": "STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6",
+      "sponsored": false,
+      "post_condition_mode": "deny",
+      "token_transfer": {
+        "recipient_address": "ST1GY25DM8RZV4X15X07THRZ2C5NMWPGQWKFGV87F",
+        "amount": "500000",
+        "memo": "0x46617563657400000000000000000000000000000000000000000000000000000000"
+      }
+    },
+    {
+      "tx_id": "0x5bed8e3f801cb4e2c74d2815a092f7c1c6a35f2fce4a80c80ca70848d34cb395",
+      "tx_status": "pending",
+      "tx_type": "token_transfer",
+      "receipt_time": 1598282326,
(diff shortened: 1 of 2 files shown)


docs/api/transaction/get-mempool-transactions.example.json
```diff
@@ -0,0 +1,391 @@
...
+      "fee_rate": "2000",
+      "sender_address": "ST2R1XSFXYHCSFE426HP45TTD8ZWV9XHX2SRP3XA8",
+      "sponsored": false,
+      "post_condition_mode": "deny",
+      "post_conditions": [],
+      "smart_contract": {
+        "contract_id": "ST2R1XSFXYHCSFE426HP45TTD8ZWV9XHX2SRP3XA8.test-loans-1",
+        "source_code": "(define-data-var stx-loaned int 0)\n(define-data-var lockup-period int 0)\n(define-data-var stx-return int 0)\n(define-public (get-stx-return (stx uint) (months uint))\n    (ok\n        (begin\n            (var-set stx-loaned (to-int(stx))\n            (var-set lockup-period (to-int(months))\n            (calculate-stx-return)\n            (print (var-get stx-return))\n            (transfer-to-server)\n        )\n    )\n)\n(define-private (calculate-stx-return)\n    (ok\n        (begin\n            (var-set stx-return (- (+ (* (/ (* (var-get stx-loaned) 5) 100) (var-get lockup-period)) (var-get stx-loaned)) 5))\n        )\n    )\n)\n(define-private (transfer-to-server)\n  (begin\n    (unwrap-panic (stx-transfer? (to-uint (var-get stx-loaned)) tx-sender 'ST2R1XSFXYHCSFE426HP45TTD8ZWV9XHX2SRP3XA8))\n    (ok (to-uint (var-get stx-loaned)))\n  )\n)"
```
agraebe, line 67: how do we feel about the length of these fields? This would be shown in the docs and will be hard to read anyways

docs/api/transaction/get-mempool-transactions.example.json
```diff
@@ -0,0 +1,391 @@
+{
+  "limit": 96,
+  "offset": 0,
+  "total": 24,
+  "results": [
+    {
```
agraebe, line 6: It might be enough to have 1-2 JSON objects in here as a sample

kyranjamie: cut it to 5, which I think covers enough variants

docs/api/transaction/get-mempool-transactions.example.json
kyranjamie: It doesn't show up in outputs docs, so think it's worth having a complete contract

agraebe: Wouldn't it show up on the right side of these docs as a sample? https://blockstack.github.io/stacks-blockchain-api/

kyranjamie: Only if you expand it 
image

agraebe: yeah - and if you do, it's not formatted well enough to be readable. the benefit of this lengthy example would be outside of the docs then I'd assume. fine with me :)

PR #231, opened 2020-09-11
Rosetta: /construction/derive endpoint

asimm241:
- [ ] Rosetta construction/derive endpoint

### DIFF
--- a/src/rosetta-helpers.ts
+++ b/src/rosetta-helpers.ts
@@ -6,6 +6,7 @@ import { RosettaOperation } from '@blockstack/stacks-blockchain-api-types';
 
 import * as btc from 'bitcoinjs-lib';
 import * as c32check from 'c32check';
+import { RosettaNetworks } from './api/rosetta-constants';
 
 enum CoinAction {
   CoinSpent = 'coin_spent',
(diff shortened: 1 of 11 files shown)


src/rosetta-helpers.ts
```diff
@@ -169,3 +172,17 @@ function makePoisonMicroblockOperation(tx: DbMempoolTx | DbTx, index: number): R
 
   return sender;
 }
+
+export function publicKeyToBitcoinAddress(publicKey: string): string {
+  const publicKeyBuffer = Buffer.from(publicKey, 'hex');
+
+  const address = btc.payments.p2pkh({
+    pubkey: publicKeyBuffer,
+    network: btc.networks.regtest,
```
zone117x, line 181: This can't be hardcoded to regtest.

asimm241: I have updated the PR with this fix.

src/rosetta-helpers.ts
```diff
@@ -169,3 +173,24 @@ function makePoisonMicroblockOperation(tx: DbMempoolTx | DbTx, index: number): R
...
+    btcNetwork = btc.networks.bitcoin;
+  } else {
+    btcNetwork = btc.networks.regtest;
+  }
+
+  const address = btc.payments.p2pkh({
+    pubkey: publicKeyBuffer,
+    network: btcNetwork,
+  });
+  return address.address ? address.address : '';
```
zone117x, line 191: This is a failure condition correct? If so, it should throw rather than fail silently.

asimm241: This function is being called in a try-catch block that will handle if `p2pkh` fails. That being said this function should be autonomous. I will fix and update the PR.

zone117x (approved):
LGTM!

Issue #222, opened 2020-09-01
OpenAPI & client missing pagination & proof request parameters


agraebe:
It is currently not possible to paginate through API responses with the client because the request params are not documented in the OpenAPI spec. 

The same is true for the `proofs` parameter.

We should update the spec, generate the docs, and update the client lib

PR #236, opened 2020-09-14
fix: pagination and proof params in openapi spec

agraebe:
fixing #222

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -101,6 +101,26 @@ paths:
         If using TypeScript, import typings for this response from our types package:
 
         `import type { TransactionResults } from '@blockstack/stacks-blockchain-api-types';`
+      parameters:
+        - name: limit
+          in: query
+          description: max number of transactions to fetch
+          required: false
+          schema:
+            type: integer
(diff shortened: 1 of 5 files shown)


docs/openapi.yaml
```diff
@@ -101,6 +101,26 @@ paths:
...
+          description: index of first transaction to fetch
+          required: false
+          schema:
+            type: integer
+        - name: type
+          in: query
+          description: Filter by transaction type
+          required: false
+          schema:
+            type: string
```
zone117x, line 122: This is type of `Array<string>`. Looks like the Express query array pattern is supported by openapi, 
e.g. `?type=coinbase&type=token_transfer`

https://swagger.io/docs/specification/serialization:
image

zone117x: So something like
```
type: array
items:
  type: string
  enum: [coinbase, token_transfer, smart_contract, contract_call, poison_microblock]
```

agraebe: Cool, updated!

zone117x (approved):
LGTM. Thanks @agraebe!

PR #238, opened 2020-09-15
feat: add prometheus metrics endpoint

CharlieC3:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/237

* Implements a Prometheus metrics endpoint with the [same libraries being used in the new Explorer](https://github.com/blockstack/explorer/blob/master/server/index.js#L7-L8)
* This endpoint can be accessed at the root of port `9153`.
* Fixed a build error 
* Tested using `docker-compose.yml`


src/api/init.ts
```diff
@@ -38,6 +38,11 @@ export async function startApiServer(datastore: DataStore): Promise<ApiServer> {
 
   const apiHost = process.env['STACKS_BLOCKCHAIN_API_HOST'];
   const apiPort = parseInt(process.env['STACKS_BLOCKCHAIN_API_PORT'] ?? '');
+  const dev = process.env.NODE_ENV !== 'production';
+
+  const { createMiddleware: createPrometheusMiddleware } = require('@promster/express');
```
zone117x, line 43: Modules should be imported using esm syntax. This needs to be something like: 
`import { createMiddleware as createPrometheusMiddleware } from '@promster/express';`
and located at the top of the file alongside the other imports.

CharlieC3: @zone117x [This is what I tried previously here](https://github.com/blockstack/stacks-blockchain-api/pull/238/commits/ff8f620aaff5040dd166c31f2b902124eef3a10f#diff-f0977764d44e25489d36496dcb219a87R26-R27), but it resulted in various errors when trying to use the module. Not sure if there is a functional difference between the syntax I used and your example. The promster documentation for using `import` is lacking and they seem to be very specific about using `require`, so that's why I changed it back. Will look into it more today and try your syntax.

CharlieC3:
I reverted back to using `import` instead of `require` as requested, however [line 59](https://github.com/blockstack/stacks-blockchain-api/blob/5591801329d2604c831a362f4a1b0347396c5b8a/src/api/init.ts#L59) now gives the following error:
```
Type 'ExpressWithAsync' is missing the following properties from type 'TApp': close, address, getConnections, ref, and 9 more.ts(2740)
```

zone117x (approved):
Thanks @CharlieC3!

zone117x:
@timstackblock moving this to QA. We just need a basic regression test to make sure the service starts as expected, since this feature is enabled by default.

CharlieC3:
> Thanks @CharlieC3!

Thank you for your help in this!

PR #246, opened 2020-09-21
Revisit faucet generosity

lgalabru:
Address https://github.com/blockstack/stacks-blockchain/issues/1883

### DIFF
--- a/src/api/routes/faucets.ts
+++ b/src/api/routes/faucets.ts
@@ -75,7 +75,7 @@ export function createFaucetRouter(db: DataStore): RouterWithAsync {
       // Only based on address for now, but we're keeping the IP in case
       // we want to escalate and implement a per IP policy
       const now = Date.now();
-      const window = 5 * 60 * 1000; // 5 minutes
+      const window = 2 * 24 * 60 * 60 * 1000; // 2 days
       const requestsInWindow = lastRequests.results
         .map(r => now - r.occurred_at)
         .filter(r => r <= window);
@@ -88,7 +88,7 @@ export function createFaucetRouter(db: DataStore): RouterWithAsync {
         return;
       }
 
-      const stxAmount = 500_000; // 0.5 STX
+      const stxAmount = 3_333_333_333_333_333; // Minimum required for stacking
       const tx = await makeSTXTokenTransfer({
         recipient: address,
         amount: new BN(stxAmount),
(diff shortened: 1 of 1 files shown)


src/api/routes/faucets.ts
```diff
@@ -75,7 +75,7 @@ export function createFaucetRouter(db: DataStore): RouterWithAsync {
       // Only based on address for now, but we're keeping the IP in case
       // we want to escalate and implement a per IP policy
       const now = Date.now();
-      const window = 5 * 60 * 1000; // 5 minutes
+      const window = 2 * 24 * 60 * 60 * 1000; // 2 days
```
zone117x, line 78: I think people will run into issues with a 2 day throttle, but not sure of a good alternative.

psq: not too hard to "circumvent" by using multiple addresses...

zone117x:
@aulneau @hstove @markmhx can we proceed with a 2 day faucet cooldown?

aulneau:
I'd ask the community. It seems like two days would be too much imo.

diwakergupta:
@lgalabru what was your rationale for changing it? If there's nothing fundamental driving the 2 day cooldown, we should just keep it low, maybe 10-15min.

zone117x:
I think it's because the faucet can only be used around <120 times before running out, because of the large stx amount?

zone117x:
Maybe we can have a different faucet endpoint with this aggressive throttle for those who want to try out Stacking, and preserve the behavior of the current one?

diwakergupta:
> Maybe we can have a different faucet endpoint with this aggressive throttle for those who want to try out Stacking, and preserve the behavior of the current one?

yep, might be a good middle-ground.

> I think it's because the faucet can only be used around <120 times before running out, because of the large stx amount?

sure, but in theory we could just give the faucet a LOT more STX so this becomes a non-issue. If 10-15min is too short, maybe we can make it an hour. 2 days will definitely frustrate a LOT of people.

zone117x:
> but in theory we could just give the faucet a LOT more STX so this becomes a non-issue

I might have misunderstood during the last meeting discussing Stacking, but I think we can't give the faucet more because it needs to give out 1/120th of the liquid supply, so if we increase the faucet supply then we have to proportionally increase the faucet disperse amount?

src/api/routes/faucets.ts
lgalabru: Updated to 1 hour

lgalabru: As @psq mentioned, if for some reason (?) this backoff is an issue, users can just use a different address.

lgalabru: Opening https://github.com/blockstack/stacks-blockchain-api/issues/247, for the sustainable approach

PR #249, opened 2020-09-22
Implementation of construction/metadata and construction/hash endpoints

asimm241:
1. Implementation of rosetta construction/metadata endpoint 
 Details [here](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionmetadata)
- [ ] New feature implementation 
- [ ] Unit test coverage

2. Implementation of rosetta construction/hash endpoint
Details [here](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionhash)
- [ ]  New feature implementation
- [ ] Unit test coverate


src/api/routes/rosetta/construction.ts
```diff
@@ -120,5 +131,100 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+    }
+
+    const transaction = deserializeTransaction(BufferReader.fromBuffer(buffer));
+    const hash = transaction.txid();
+    if (
+      transaction.auth.spendingCondition?.hashMode == AddressHashMode.SerializeP2PKH ||
+      transaction.auth.spendingCondition?.hashMode == AddressHashMode.SerializeP2WPKH
+    ) {
+      const signature = transaction.auth.spendingCondition.signature;
+      if (parseInt(signature.data, 16) == 0) {
```
zone117x, line 213: This is testing for loose equality, and it's not very clear what this trying to check for

asimm241: This is trying to check whether the transaction is signed or not. If the signature data is empty i.e `0x00` or  `0`, the transaction is not signed if it is non zero it is signed.

asimm241: But yeah I agree it is loose equality but I couldn't found any other way to differentiate between signed and unsigned transactions.

zone117x: Can it be made into an explicit check without loose equality? If the stacks-transactions-js lib changes this field, I'd prefer something to break in an explicit, clear way.

asimm241:
@zone117x,  I have updated the PR. I have done some refactoring but essentially it's the same check.  I have also confirmed that currently, this is the only way to check whether the transaction is not signed or not.

zone117x (approved):
LGTM!

PR #250, opened 2020-09-24
noop handler for STXLockEvent

zone117x:
Noop handler for the stx-lock-event introduced in https://github.com/blockstack/stacks-blockchain/pull/1899

This _should_ prevent rejects when the event is received.


lgalabru (approved):
Thank you for addressing this!!

PR #251, opened 2020-09-24
Updating DNS to use CNAME records

wileyj:
Updating all hardcoded records to use CNAME'd records. 

This update will point to the latest stage of the testnet, vs. a specific stage like `krypton`. 
During a stage cutover, the CNAME's will be updated at the DNS level to point to the current stage. 

- for example (bitcoin):
`bitcoind.blockstack.org IN CNAME bitcoind.krypton.blockstack.org`


agraebe:
Cool, I also update the Postman Collection: https://www.getpostman.com/collections/614feab5c108d292bffa

agraebe:
I generated the client based on the changes and updated the API references

zone117x (approved):
LGTM. Some of the docker commands need tested, but I'd rather do that once images are pushed (rather than building locally)

Issue #223, opened 2020-09-01
Support for new Clarity string types


zone117x:
Bump stacks-transactions-js lib version https://github.com/blockstack/stacks-transactions-js/pull/118

Update any usages as needed.

zone117x:
This repo still uses its own tx deserialization code in some areas (this was implemented before the stacks-transactions-js lib was ready). Those areas need to either be refactored to use `stacks-transaction-js`, or (less ideally) the deserialization code needs updated to handle the new string types.

PR #252, opened 2020-09-24
fix: support new Clarity string types #223

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/223


lgalabru (approved):
LGTM thank you @zone117x!

PR #253, opened 2020-09-24
feat: hacks to get started with pox

hstove:
This is just some hacky code that I used to get PoX working with the sidecar. Main issues:

- Faucet needed to send more STX
- Server would abort when it saw an unrecognized event. In this case we have a new event `stx_lock_event`
- The server doesn't know about smart contracts that were created during boot. So, if you tried to fetch a TX for a contract call to `pox.clar`, it'd throw, since that contract is not in the DB

This code should be considered all hacky workarounds, do not merge, only use if you need to use some PoX features locally today.

### DIFF
--- a/.env
+++ b/.env
@@ -1,8 +1,8 @@
 PG_HOST=127.0.0.1
-PG_PORT=5490
+PG_PORT=5432
 PG_USER=postgres
 PG_PASSWORD=postgres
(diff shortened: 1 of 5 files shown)


zone117x:
@hstove can you give current master a try? These issues should now be resolved.

PR #254, opened 2020-09-24
feat: add v2/pox endpoint proxy and documentation

agraebe:
@zone117x thought I'd help out and add the proxy for the new `v2/pox` RPC endpoint. This will also include updates to the spec + client lib


zone117x (approved):
Thanks @agraebe!

PR #255, opened 2020-09-25
Reduce amount of STX per /faucet

lgalabru:
Give away amount in line with:
```json
{"contract_id":"ST000000000000000000002AMW42H.pox","first_burnchain_block_height":0,"min_amount_ustx":2000001725000,"prepare_cycle_length":30,"rejection_fraction":25,"reward_cycle_id":4,"reward_cycle_length":120}
```

### DIFF
--- a/src/api/routes/faucets.ts
+++ b/src/api/routes/faucets.ts
@@ -88,7 +88,7 @@ export function createFaucetRouter(db: DataStore): RouterWithAsync {
         return;
       }
 
-      const stxAmount = 3_333_333_333_333_333; // Minimum required for stacking
+      const stxAmount = 3_000_000_000_000; // Minimum required for stacking
       const tx = await makeSTXTokenTransfer({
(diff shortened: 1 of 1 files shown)


src/api/routes/faucets.ts
```diff
@@ -88,7 +88,7 @@ export function createFaucetRouter(db: DataStore): RouterWithAsync {
         return;
       }
 
-      const stxAmount = 3_333_333_333_333_333; // Minimum required for stacking
+      const stxAmount = 3_000_000_000_000; // Minimum required for stacking
```
psq, line 91: @lgalabru once you include fees to make the stacking request, this won't be enough... So everyone needs to make at least 2 requests

psq: never mind, it looks like the amount is 2_000_000_000_000, so plenty for fees...

PR #259, opened 2020-09-28
Rosetta construction parse api

fdmirza:


src/tests-rosetta/api-construction.ts
```diff
@@ -551,6 +553,191 @@ describe('Rosetta API', () => {
...
+
+  test('parse api signed', async () => {
+    const request: RosettaConstructionParseRequest = {
+      network_identifier: {
+        blockchain: 'stacks',
+        network: 'testnet',
+      },
+      signed: true,
+      transaction:
+        '0x80800000000400164247d6f2b425ac5771423ae6c80c754f7172b0000000000000000000000000000000b400011ae06c14c967f999184ea8a7913125f09ab64004446fca89940f092509124b9e773aef483e925476c78ec58166dcecab3875b8fab8e9aa4213179d164463962803020000000000051a1ae3f911d8f1d46d7416bfbe4b593fd41eac19cb00000000000003e800000000000000000000000000000000000000000000000000000000000000000000',
```
zone117x, line 565: Can you change these raw tx payloads to a tx created by the stacks-transactions-js lib? Otherwise it's hard to compare with the below output

fdmirza: Yes, that make sense.

zone117x (approved):
LGTM!

zone117x (changes requested):
Actually, looks like there's a problem with the Rosetta tests.
There are two files that attempt to bootstrap a new db but are reusing the same core-node instance. One in `src/tests-rosetta/api.ts` and another in `src/tests-rosetta/api-construction.ts`.

The duplicate `beforeAll` hooks are causing the issue. Easiest fix is to consolidate the Rosetta tests into one `describe` test group.

PR #265, opened 2020-10-01
fix: add get contract data map entry schema #264

### LINKED ISSUE #264: [client] SmartContractApis.getContractDataMapEntry returns Promise<void>
**Describe the bug**
When querying a data map entry only `SmartContractApis.getContractDataMapEntryRaw` works because `SmartContractApis.getContractDataMapEntry` does not return anything.

**To Reproduce**
See https://github.com/blockstack/stacks-blockchain-api/blob/master/client/src/generated/apis/SmartContractsApi.ts#L198

**Expected behavior**
The methods should return an object `{data, proof}`

friedger:
This PR
* adds schema and example for 200 repsonse of `/v2/map_entry/{stacks_address}/{contract_name}/{map_name}`
* adds a test call with updated url to krypton testnet
* fixes #264


friedger:
closing this in favor of #267

PR #272, opened 2020-10-02
ConflictingNonceInMempool fix: increment faucet mempool tx nonces

zone117x:
Allows multiple faucet transactions per block by incrementing the nonce on `ConflictingNonceInMempool` error. Up to 5 faucet transactions are allowed to be stacked in the mempool (this is hardcoded in core). 

This **_does not_** fix the related problem with txs getting permanently stuck or lost, which can also cause the `ConflictingNonceInMempool` error. For reference, see https://github.com/blockstack/stacks-blockchain/issues/1941 and https://github.com/blockstack/stacks-blockchain/issues/1940


kantai (approved):
LGTM!

PR #275, opened 2020-10-02
Implementation of construction/payloads and construction/combine endpoints

asimm241:
- [x] [construction/payloads](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionpayloads)
 -implment construction/payloads endpoint 
 -unit test coveratge

- [x] [construction/combine](https://www.rosetta-api.org/docs/ConstructionApi.html#constructioncombine)
 -implment construction/combine endpoint 
 -unit test coveratge

### DIFF
--- a/src/api/routes/rosetta/construction.ts
+++ b/src/api/routes/rosetta/construction.ts
@@ -417,7 +417,7 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync 
         amount: new BN(amount),
         fee: new BN(fees),
         publicKeys: publicKeysStrings,
-        numSignatures: 2,
+        numSignatures: publicKeys.length,
         network: GetStacksTestnetNetwork(),
         nonce: accountInfo.nonce ? new BN(accountInfo.nonce) : new BN(0),
       };
--- a/src/api/routes/rosetta/construction.ts
+++ b/src/api/routes/rosetta/construction.ts
@@ -1,4 +1,5 @@
 import { addAsync, RouterWithAsync } from '@awaitjs/express';
+import * as BN from 'bn.js';
 import {
   NetworkIdentifier,
   RosettaConstructionDeriveResponse,
@@ -13,6 +14,7 @@ import {
   RosettaConstructionSubmitResponse,
   RosettaConstructionPreprocessRequest,
   RosettaConstructionMetadataRequest,
+  RosettaConstructionPayloadResponse,
 } from '@blockstack/stacks-blockchain-api-types';
 import {
   emptyMessageSignature,
@@ -20,10 +22,14 @@ import {
 } from '@blockstack/stacks-transactions/lib/authorization';
 import { BufferReader } from '@blockstack/stacks-transactions/lib/bufferReader';
 import { deserializeTransaction } from '@blockstack/stacks-transactions/lib/transaction';
+import {
+  UnsignedTokenTransferOptions,
+  makeUnsignedSTXTokenTransfer,
+} from '@blockstack/stacks-transactions';
 import * as express from 'express';
 import { StacksCoreRpcClient } from '../../../core-rpc/client';
 import { DataStore, DbBlock } from '../../../datastore/common';
-import { FoundOrNot, hexToBuffer, isValidC32Address } from '../../../helpers';
+import { FoundOrNot, hexToBuffer, isValidC32Address, digestSha512_256 } from '../../../helpers';
 import { RosettaConstants, RosettaErrors } from '../../rosetta-constants';
 import {
   bitcoinAddressToSTXAddress,
@@ -36,6 +42,7 @@ import {
   publicKeyToBitcoinAddress,
   rawTxToBaseTx,
   rawTxToStacksTransaction,
+  GetStacksTestnetNetwork,
 } from './../../../rosetta-helpers';
 import { makeRosettaError, rosettaValidateRequest, ValidSchema } from './../../rosetta-validate';
 
@@ -329,7 +336,73 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync 
   });
 
   //construction/payloads endpoint
-  router.postAsync('/payloads', async (req, res) => {});
+  router.postAsync('/payloads', async (req, res) => {
+    const valid: ValidSchema = await rosettaValidateRequest(req.originalUrl, req.body);
+    if (!valid.valid) {
+      res.status(400).json(makeRosettaError(valid));
+      return;
+    }
+
+    const options = getOptionsFromOperations(req.body.operations);
+    if (options == null) {
+      res.status(400).json(RosettaErrors.invalidOperation);
+      return;
+    }
+
+    const amount = options.amount;
+    if (!amount) {
+      res.status(400).json(RosettaErrors.invalidAmount);
+      return;
+    }
+
+    const fees = options.fee;
+    if (!fees) {
+      res.status(400).json(RosettaErrors.invalidFees);
+      return;
+    }
+
+    const publicKeys: RosettaPublicKey[] = req.body.public_keys;
+    if (!publicKeys) {
+      res.status(400).json(RosettaErrors.emptyPublicKey);
+      return;
+    }
+
+    if (publicKeys[0].curve_type !== 'secp256k1') {
+      res.status(400).json(RosettaErrors.invalidCurveType);
+      return;
+    }
+
+    const recipientAddress = options.token_transfer_recipient_address
+      ? options.token_transfer_recipient_address
+      : '';
+    const senderAddress = options.sender_address ? options.sender_address : '';
+
+    const accountInfo = await new StacksCoreRpcClient().getAccount(senderAddress);
+
+    const tokenTransferOptions: UnsignedTokenTransferOptions = {
+      recipient: recipientAddress,
+      amount: new BN(amount),
+      fee: new BN(fees),
+      publicKey: publicKeys[0].hex_bytes,
+      network: GetStacksTestnetNetwork(),
+      nonce: accountInfo.nonce ? new BN(accountInfo.nonce) : new BN(0),
+    };
+
+    const transaction = await makeUnsignedSTXTokenTransfer(tokenTransferOptions);
+    const unsignedTransaction = transaction.serialize();
+    const hexBytes = digestSha512_256(unsignedTransaction).toString('hex');
+    const response: RosettaConstructionPayloadResponse = {
+      unsigned_transaction: unsignedTransaction.toString('hex'),
+      payloads: [
+        {
+          address: senderAddress,
+          hex_bytes: '0x' + hexBytes,
+          signature_type: 'ecdsa',
+        },
+      ],
+    };
+    res.json(response);
+  });
 
   return router;
 }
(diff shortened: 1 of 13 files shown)


src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +322,147 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+      return;
+    }
+
+    const publicKeys: RosettaPublicKey[] = req.body.public_keys;
+    if (!publicKeys) {
+      res.status(400).json(RosettaErrors.emptyPublicKey);
+      return;
+    }
+
+    if (publicKeys[0].curve_type !== 'secp256k1') {
```
zone117x, line 357: What if there's more than 1 `req.body.public_keys` ? Seems like it would need to return a multisig transaction?

src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +322,147 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+
+      if (!verifySignature(preSignHash, signature.public_key.hex_bytes, newSignature)) {
+        res.status(400).json(RosettaErrors.signatureNotVerified);
+      }
+
+      if (transaction.auth.spendingCondition && isSingleSig(transaction.auth.spendingCondition)) {
+        transaction.auth.spendingCondition.signature = newSignature;
+      } else {
+        const authField = createTransactionAuthField(newSignature);
+        transaction.auth.spendingCondition?.fields.push(authField);
```
zone117x, line 454: It doesn't look like this path (multisig) is tested?

src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +322,147 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+      res.status(400).json(RosettaErrors.invalidCurveType);
+      return;
+    }
+
+    const recipientAddress = options.token_transfer_recipient_address
+      ? options.token_transfer_recipient_address
+      : '';
+    const senderAddress = options.sender_address ? options.sender_address : '';
+
+    const accountInfo = await new StacksCoreRpcClient().getAccount(senderAddress);
```
zone117x, line 367: What happens when the above `sender_address` or `recipient_address` are empty/undefined? Seems like this would be an error condition?

asimm241:
@zone117x Fixed all the issues and updated PR.

src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +324,174 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+      //multi signature
+      const publicKeysStrings = publicKeys.map(key => {
+        return key.hex_bytes;
+      });
+      tokenTransferOptions = {
+        recipient: recipientAddress,
+        amount: new BN(amount),
+        fee: new BN(fees),
+        publicKeys: publicKeysStrings,
+        numSignatures: 2,
```
zone117x, line 391: This appears to only allow for 2-of-2 multisig transactions, it should support N-of-M

asimm241: It should be equal to the number of public keys the user provides right?

src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +324,174 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+      //multi signature
+      const publicKeysStrings = publicKeys.map(key => {
+        return key.hex_bytes;
+      });
+      tokenTransferOptions = {
+        recipient: recipientAddress,
+        amount: new BN(amount),
+        fee: new BN(fees),
+        publicKeys: publicKeysStrings,
+        numSignatures: publicKeys.length,
```
reedrosenbluth, line 391: did you intend for `publicKeys.length` to be the value here? This will mean that all of the keys associated with `publicKeys` will have to eventually sign this tx...

typically the creator of a multi-sig transaction is able to specify what this value should be.

asimm241: I'm assuming the caller is only specifying keys whose signatures are required. It means the caller is the creator of mulit-sig/single-sig transaction here. I can not find any other param in rosetta which will indicate signatures and public keys can be different.

src/tests-rosetta/api.ts
```diff
@@ -1074,6 +1090,717 @@ describe('Rosetta API', () => {
...
+          hex_bytes: '0x' + hexBytes,
+          signature_type: 'ecdsa',
+        },
+      ],
+    };
+
+    expect(JSON.parse(result.text)).toEqual(expectedResponse);
+  });
+
+  test('payloads multi sign success', async () => {
```
zone117x, line 1213: typo? Test looks like `payloads multi sig success`, the test is only unsigned txs

src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +324,174 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+      }
+
+      if (!verifySignature(preSignHash, signature.public_key.hex_bytes, newSignature)) {
+        res.status(400).json(RosettaErrors.signatureNotVerified);
+      }
+
+      if (transaction.auth.spendingCondition && isSingleSig(transaction.auth.spendingCondition)) {
+        transaction.auth.spendingCondition.signature = newSignature;
+      } else {
+        const authField = createTransactionAuthField(newSignature);
```
zone117x, line 482: This branch is still not covered by tests

src/api/routes/rosetta/construction.ts
zone117x: @asimm241 can you verify that other Rosetta implementations do not support m-of-n multisig transactions?

src/tests-rosetta/api.ts
asimm241: No,  It creates unsigned txs with multiple signatures.

src/api/routes/rosetta/construction.ts
asimm241: api.ts-> Line number 1575 covers this branch.

src/api/routes/rosetta/construction.ts
asimm241: @zone117x I haven't found any implementation to support multi-sig for rosetta in other blockchains. I have found this question https://community.rosetta-api.org/t/construction-api-and-multi-sig-accounts/208. This states that they are targeting m-of-n multi-sig for rosetta `v1.5.0`

Keeping this in view should we postpone the support of multi-sig. 

Moreover, BTW rosetta docs do not require public keys in payloads request but `makeUnsignedTokenTransferTransaction` requires a public key(s), is there any other way to create a transaction without specifying the public key(s)?

src/api/routes/rosetta/construction.ts
zone117x: Check the code coverage output, the line is never it. It appears the test exits early

src/tests-rosetta/api.ts
```diff
@@ -1074,6 +1090,717 @@ describe('Rosetta API', () => {
...
+
+    const unsignedTransaction = await makeUnsignedSTXTokenTransfer(txOptions);
+    const unsignedSerializedTx = unsignedTransaction.serialize().toString('hex');
+
+    const signer = new TransactionSigner(unsignedTransaction);
+    signer.signOrigin(createStacksPrivateKey(testnetKeys[0].secretKey));
+    const signedSerializedTx = unsignedTransaction.serialize().toString('hex');
+
+    const signature = getSignature(unsignedTransaction);
+    if (!signature) return;
```
zone117x, line 1598: The test exits here early without error

src/api/routes/rosetta/construction.ts
zone117x: > This states that they are targeting m-of-n multi-sig for rosetta v1.5.0. Keeping this in view should we postpone the support of multi-sig.

We can postpone m-of-n until then 👍 

> rosetta docs do not require public keys in payloads request but makeUnsignedTokenTransferTransaction requires a public key(s)

It appears to be not-required to allow support for blockchains that do not need public keys. Is there anywhere that states the endpoint must work when public keys are not provided?

src/tests-rosetta/api.ts
asimm241: I should fix this or remove this test and multi-sig support altogether depends on how we proceed on https://github.com/blockstack/stacks-blockchain-api/pull/275#discussion_r501574721.

src/api/routes/rosetta/construction.ts
asimm241: No, I haven't encountered anything like that so far, I'm searching for it, I'm also thinking of asking a question on their community forum. Let's see what happens.

src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +324,161 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+      return;
+    }
+
+    const accountInfo = await new StacksCoreRpcClient().getAccount(senderAddress);
+
+    let tokenTransferOptions: UnsignedTokenTransferOptions | UnsignedMultiSigTokenTransferOptions;
+
+    if (publicKeys.length !== 1) {
+      //TODO support multi-sig in the future.
+      res.status(400).json(RosettaErrors.needOnePublicKey);
```
zone117x, line 377: Do other rosetta implementations do the same thing when there are more than 1 public keys?

src/api/routes/rosetta/construction.ts
```diff
@@ -297,5 +324,161 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
...
+
+    try {
+      unsigned_transaction_buffer = hexToBuffer('0x' + combineRequest.unsigned_transaction);
+      transaction = deserializeTransaction(BufferReader.fromBuffer(unsigned_transaction_buffer));
+    } catch (e) {
+      res.status(400).json(RosettaErrors.invalidTransactionString);
+      return;
+    }
+
+    if (signatures.length !== 1) res.status(400).json(RosettaErrors.needOnlyOneSignature);
```
zone117x, line 443: Same question here -- is this what other Rosetta implementations do?

asimm241: Yes!
 https://github.com/iotexproject/iotex-core-rosetta-gateway/blob/master/services/construction.go#L65

zone117x: Alright, we should create a github issue for tracking multi-sig support once the Rosetta spec adds it

src/api/routes/rosetta/construction.ts
asimm241: Other implementations that I have looked at don't require a public key to build an unsigned transaction. 
Is there a way we can create a transaction without specifying the public key?

asimm241: Here is the PR containing those changes. https://github.com/blockstack/stacks-blockchain-api/pull/288

src/api/routes/rosetta/construction.ts
zone117x: Created issue https://github.com/blockstack/stacks-blockchain-api/issues/289

src/api/routes/rosetta/construction.ts
zone117x: PR #288 merged. This PR needs a rebase to fix the conflicting file and then looks good to merge 👍

zone117x (approved):

PR #285, opened 2020-10-09
feat: adding new pox & stacking fields

agraebe:
in reference to https://github.com/blockstack/stacks-blockchain/pull/1920/

the generated client seems to also fix an issue with missing Rosetta endpoints (likely from https://github.com/blockstack/stacks-blockchain-api/pull/269)


docs/index.d.ts
```diff
@@ -200,6 +202,8 @@ export interface CoreNodePoxResponse {
   rejection_fraction: number;
   reward_cycle_id: number;
   reward_cycle_length: number;
+  rejection_votes_left_required: number;
+  total_liquid_supply_ustx: number;
```
zone117x, line 206: The field seems to be inconsistent with the other fields using a string to represent large integer values (e.g. `balance`, `locked`). Not going to hold up this PR over it though.

zone117x: cc @agraebe

PR #293, opened 2020-10-14
fix: adding schema for fee estimates

agraebe:
without this, the response is simply undefined


zone117x:
Looks like the client code needs rebuilt and committed?

agraebe:
@zone117x 🤦 done now

zone117x (approved):

PR #311, opened 2020-10-27
rosetta docker file and instructions on how to use it

fariedt:
This pull request adds a docker file to meet Rosetta's [requirements](https://www.rosetta-api.org/docs/node_deployment.html#dockerfile-expectations), and instructions on how to run `rosetta-cli` against it to test the implementation.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

No.

No.

## Testing information

2. If it’s a bug fix, list steps to reproduce the bug
3. Briefly mention affected code paths
4. List other affected projects if possible
5. Things to watch out for when testing

## Checklist
- [X] Code is commented where needed
- [X] `npm run test` passes
- [ ] Changelog is updated
- [ ] Tag 1 of @person1 or @person2


stx-rosetta.Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+        && nvm alias default $NODE_VERSION"
+ENV PATH=$PATH:/data/stacky/.nvm/versions/node/v${NODE_VERSION}/bin
+RUN node -e 'console.log("Node.js runs")'
+
+### Setup stacks-node
+COPY --from=stacks-node-build /stacks/stacks-node stacks-node/
+ENV PATH="$PATH:$HOME/stacks-node"
+
+#### Copy stacks-node mocknet config
+COPY ./stacks-blockchain/Stacks-mocknet.toml ./
```
zone117x, line 73: Should this use an `ARG` to toggle between mocknet, testnet, and (eventually) mainnet?

fariedt: Can I assume that there will be configuration files named `Stacks-{mocknet,testnet,mainnet}` in the `stacks-blockchain` directory?

This argument will also need to be passed to the `run.sh` script and the API server, probably in an environment variable or in a generated `.env` file.  At present, the string `testnet` is hard-coded in a couple of places in the rosetta code (`src/api/rosetta-constants.ts` and in the tests).

fariedt: I've modified the dockerfile so that it doesn't need a copy of the repository cloned to build, and it now copies all the `toml` files.  However, I haven't added the `ARG` yet.

fariedt: I've pushed a couple of commits to make the rosetta docker file and the rosetta code depend on the `STACKS_NETWORK` ARG (copied to the environment in the running container).  This lets users switch between mocknet, testnet, dev, and later, mainnet.

stx-rosetta.Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+ENV HOME=/data/stacky
+WORKDIR $HOME
+USER stacky
+RUN sudo chown -R stacky:stacky $HOME
+RUN mkdir /data/stacky/.bashrc.d
+
+### Node.js
+# ENV NODE_VERSION=14.13.1
+ENV NODE_VERSION=13.14.0
+RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash \
```
zone117x, line 62: Curious to why `nvm` is used to install nodejs? As opposed to the typical apt-get command e.g.
```
curl -sL https://deb.nodesource.com/setup_14.x | bash -
apt-get install -y nodejs
```

zone117x: Ah, I'm assuming it's because of this requirement from rosetta docs?
> any required files must be downloaded from the internet with explicit versioning. If it is not possible to fetch a specific version, the hash of downloaded files must be compared with a constant defined in the Dockerfile.

fariedt: I used [follower.Dockerfile](https://github.com/blockstack/stacks-blockchain-api/blob/master/follower.Dockerfile) as the template for this dockerfile, and that uses nvm.  I can switch it to install the debs directly.

zone117x: Using the debs directly seems like a better choice (although rosetta dockerfile requirements seem ambiguous about what "external" files must be version pinned").

We can also switch to nodejs v14.x now that it's LTS.

fariedt:
Right now, the docker file uses the v0.29.2 tagged release of stacks-blockchain-api to build the image.  The changes to select the network at runtime are in the above PR, but not yet in the release.  This PR will need to be merged, and then the version in `stx-rosetta.Dockerfile` bumped to v0.29.3.

Also, the announcement messages refer to the same dockerfile in this PR.

stx-rosetta.Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get -y update && apt-get -y install openjdk-11-jre-headless
 WORKDIR /app
 
 RUN git clone -b $API_TAG --depth 1 https://github.com/blockstack/stacks-blockchain-api.git .
-COPY ./rosetta-constants.ts /app/src/api/
+# COPY ./rosetta-constants.ts /app/src/api/
 RUN grep process.env.STACKS_NETWORK /app/src/api/rosetta-constants.ts
```
zone117x, line 12: Getting an error from this line when building the dockerfile standalone 
```
Step 6/63 : RUN grep process.env.STACKS_NETWORK /app/src/api/rosetta-constants.ts
 ---> Running in 045d16eb19c2
The command '/bin/sh -c grep process.env.STACKS_NETWORK /app/src/api/rosetta-constants.ts' returned a non-zero code: 1
```
If I comment out this line, everything seems to work as expected. Should these lines, 11-12, be removed?

zone117x (approved):
Works great, thanks @fariedt!

PR #313, opened 2020-10-28
Changes after check construction

asimm241:
Changes needed in the rosetta API's for check: construction to run

[x] fix payload API to return pre signature hash
[x] rotate the signature string to get the correct signature,  https://github.com/coinbase/rosetta-sdk-go/issues/201
[x] manage funded account 
[x] miscellaneous fixes 

@fariedt

### DIFF
--- a/docs/entities/transactions/transaction-status.schema.json
+++ b/docs/entities/transactions/transaction-status.schema.json
@@ -3,5 +3,5 @@
   "title": "TransactionStatus",
   "description": "Status of the transaction",
   "type": "string",
-  "enum": ["success", "pending", "abort_by_response", "abort_by_post_condition"]
+  "enum": ["success", "pending", "abort_by_response", "abort_by_post_condition", ""]
 }
--- a/docs/entities/transactions/transaction-status.schema.json
+++ b/docs/entities/transactions/transaction-status.schema.json
@@ -3,5 +3,5 @@
   "title": "TransactionStatus",
   "description": "Status of the transaction",
   "type": "string",
-  "enum": ["success", "pending", "abort_by_response", "abort_by_post_condition", ""]
+  "enum": ["success", "pending", "abort_by_response", "abort_by_post_condition"]
 }
--- a/src/api/routes/rosetta/account.ts
+++ b/src/api/routes/rosetta/account.ts
@@ -10,7 +10,6 @@ import {
 } from '@blockstack/stacks-blockchain-api-types';
 import { RosettaErrors, RosettaConstants } from '../../rosetta-constants';
 import { rosettaValidateRequest, ValidSchema, makeRosettaError } from '../../rosetta-validate';
-import { StacksCoreRpcClient } from '../../../core-rpc/client';
 
 export function createRosettaAccountRouter(db: DataStore): RouterWithAsync {
   const router = addAsync(express.Router());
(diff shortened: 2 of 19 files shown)


docs/entities/transactions/transaction-status.schema.json
```diff
@@ -3,5 +3,5 @@
   "title": "TransactionStatus",
   "description": "Status of the transaction",
   "type": "string",
-  "enum": ["success", "pending", "abort_by_response", "abort_by_post_condition"]
+  "enum": ["success", "pending", "abort_by_response", "abort_by_post_condition", ""]
```
zone117x, line 6: An empty string here is strange. Looking through the code, I'm not entirely sure why this is necessary or what it indicates. Could you explain?

src/datastore/common.ts
```diff
@@ -38,6 +38,7 @@ export enum DbTxStatus {
   Success = 1,
   AbortByResponse = -1,
   AbortByPostCondition = -2,
+  Empty = -3,
```
zone117x, line 41: Same question here -- what is the purpose of this flag?

docs/entities/transactions/transaction-status.schema.json
asimm241: @fdmirza

src/datastore/common.ts
asimm241: @fdmirza

src/api/routes/rosetta/construction.ts
```diff
@@ -484,13 +520,32 @@ export function createRosettaConstructionRouter(db: DataStore): RouterWithAsync
     let newSignature: MessageSignature;
 
     try {
-      newSignature = createMessageSignature(signatures[0].signing_payload.hex_bytes);
+      let hash = signatures[0].hex_bytes;
+      if (!hash.startsWith('01') && hash.slice(128) == '01') {
```
zone117x, line 524: This could use a comment explaining what's going on here

src/api/routes/rosetta/account.ts
```diff
@@ -48,6 +49,13 @@ export function createRosettaAccountRouter(db: DataStore): RouterWithAsync {
 
     const block = blockQuery.result;
     const result = await db.getStxBalanceAtBlock(accountIdentifier.address, block.block_height);
+    let value = result.balance.toString();
+
+    if (value == '0') {
```
zone117x, line 54: This condition is really confusing, can you explain what's going on here?

docs/entities/transactions/transaction-status.schema.json
fdmirza: This was required for doing validation with rosetta-cli. The operations that are created through rosetta-cli for check:construction has empty status. So, the validation fails if we do not handle empty status. That's why we had to introduce this.

asimm241: https://github.com/coinbase/rosetta-cli/issues/180

zone117x: The problem is that this empty status field is introduced in two areas that don't make sense and can cause confusion. 
1. Adding an empty status here results in it surfacing to the non-rosetta API docs and clients.
2. In code below, an empty status is added as a database value even though such a value should never end up in the database. 

It seems like there would be a more appropriate way to handle this situation?

fdmirza: @zone117x I understand the problem. It is an intentional behaviour in rosetta-cli for construction. As long as the transaction is not on-chain, its status need to be empty. So with the current implementation of rosetta-cli validation its required for us to have empty status for running check:construction. Its kind of a blocker for us. If you can suggest a better way to deal with this situation that would be great.

zone117x (approved):

PR #331, opened 2020-11-13
BTC rewards

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/320 - Expose BTC reward events

Three new endpoints added under `/v1/extended/burnchain` which show BTC pox reward information:

## All rewards

`GET /extended/v1/burnchain/rewards`

Returns latest an array of the latest BTC rewards. Pagination (offset/limit) is supported. Response result entries looks like:
  ```ts
 interface BurnchainReward {
    /** Set to `true` if block corresponds to the canonical burchchain tip */
    canonical: boolean;
    /** The hash representing the burnchain block */
    burn_block_hash: string;
    /** Height of the burnchain block */
    burn_block_height: number;
    /** The total amount of burnchain tokens burned for this burnchain block, in the smallest unit (e.g. satoshis for Bitcoin) */
    burn_amount: string;
    /** The recipient address that received the burnchain rewards, in the format native to the burnchain (e.g. B58 encoded for Bitcoin) */
    reward_recipient: string;
    /** The amount of burnchain tokens rewarded to the recipient, in the smallest unit (e.g. satoshis for Bitcoin) */
    reward_amount: string;
    /** The index position of the reward entry, useful for ordering when there's more than one recipient per burnchain block */
    reward_index: number;
  }
  ````

## Rewards for an address

`GET /extended/v1/burnchain/rewards/{address}`

A per-address endpoint is also implemented. This returns the same data, but only for the given address:

This endpoint accepts both Bitcoin addresses as well as STX addresses. 

## Total rewards for an address

`GET /extended/v1/burnchain/rewards/{address}/total`

Similar to the above endpoint, but returns the sum of all rewards. Result interface:
```ts
interface BurnchainRewardsTotal {
  /** The recipient address that received the burnchain rewards, in the format native to the burnchain (e.g. B58 encoded for Bitcoin)  */
  reward_recipient: string;
  /** The total amount of burnchain tokens rewarded to the recipient, in the smallest unit (e.g. satoshis for Bitcoin) */
  reward_amount: string;
}
```

## **WARNING:**
Keep in mind the underlying btc transaction info is received and stored using the native Bitcoin address. When passed a STX address, a simple trancoding (i.e. `c32check.c32ToB58(addr)`) is performed to convert the STX address into a BTC address. 

A client can perform a PoX lock up from `<STX address X>` and specify a _completely_ unrelated `<BTC address Y>`. When this happens, looking up rewards for `<STX address X>` will _NOT_ show rewards for `<BTC address Y>`. 

Clients, such as a PoX-capable-wallet, can be implemented such that all wallet initiated PoX locks are done using a BTC address derived from wallet's STX address, and can safely query for BTC rewards using that wallet address. However, nothing prevents a such a wallet user from using a different tool to perform a PoX lock for an unrelated BTC address, which the wallet will never be able to show BTC reward info for.

There also is _not_ an exhaustive 1-1 correlation between BTC rewards and STX blocks. This prevents certain features like an Explorer from showing BTC rewards on a STX block page, unless we are comfortable with some BTC reward blocks being essentially "hidden" because they do not have a corresponding STX block. See issue https://github.com/blockstack/stacks-blockchain-api/issues/332 for potential ways we can provide this capability.


kantai (approved):
LGTM!

kyranjamie:
Thanks Matt, this looks great

- What's the reason to have separate endpoints for rewards and total awards? Would be slightly easier to consume if it was just one
- Would be helpful to use a common format when describing amounts of currency, potentially even [Rossetta's schema](https://www.rosetta-api.org/docs/models/Amount.html), as to avoid satoshis getting confused with uSTX.

zone117x:
> * What's the reason to have separate endpoints for rewards and total awards? Would be slightly easier to consume if it was just one.

One is paginated. We haven't normally mixed paginated values with singletons in similar APIs. We could though as this one is relatively cheap. Should I add that?

> * Would be helpful to use a common format when describing amounts of currency, potentially even [Rossetta's schema](https://www.rosetta-api.org/docs/models/Amount.html), as to avoid satoshis getting confused with uSTX.

Yeah that seems helpful, should I add that to this PR or should we make that a separate issue?

kyranjamie:
Okay that makes sense, thanks. I'd be in favour of merging, it doesn't break any REST best practices as far as I'm aware? 🤔 

Your call as to whether to add to this PR or not. I'm not in a rush to consume this just yet, if that helps.

kyranjamie (approved):

PR #334, opened 2020-11-16
Fix mempool transaction disappear during re-orgs

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/333 - Disappearing Mempool TX

This was a bug in the sidecar's handling of mempool transactions. Transactions were not "re-inserted" into the mempool in the event that the block they were mined in became non-canonical.


kantai (approved):
LGTM!

PR #363, opened 2020-12-15
Namespace operations

noumanahmads:
## Description

Inserted the name and namespace data in the database which are extracted from the contract event. 
Following endpoints are implemented 

- /v1/namespaces/
- /v1/namespaces/:tld/names 

For details refer to [issue ](https://github.com/blockstack/stacks-blockchain-api/issues/317)

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

## Testing information
Test cases are added for the endpoints implemented. 
To run the test cases `npm run test:bns`

### DIFF
--- a/src/bns-helpers.ts
+++ b/src/bns-helpers.ts
@@ -28,6 +28,7 @@ export interface Attachment {
       name: string;
       namespace: string;
       tx_sender: Address;
+      op: string;
     };
   };
 }
@@ -52,6 +53,8 @@ export function parseNameRawValue(rawValue: string): Attachment {
     const name = nameCV.buffer.toString();
     const namespaceCV: BufferCV = metadataCV.data['namespace'] as BufferCV;
     const namespace = namespaceCV.buffer.toString();
+    const opCV: StringAsciiCV = metadataCV.data['op'] as StringAsciiCV;
+    const op = opCV.data;
     const addressCV: StandardPrincipalCV = metadataCV.data['tx-sender'] as StandardPrincipalCV;
     const address = addressCV.address;
 
@@ -62,6 +65,7 @@ export function parseNameRawValue(rawValue: string): Attachment {
           name: name,
           namespace: namespace,
           tx_sender: address,
+          op: op,
         },
       },
     };
@@ -72,7 +76,8 @@ export function parseNameRawValue(rawValue: string): Attachment {
 
 export function parseNamespaceRawValue(
   rawValue: string,
-  readyBlock: number
+  readyBlock: number,
+  txid: string
 ): DbBNSNamespace | undefined {
   const cl_val: ClarityValue = deserializeCV(hexToBuffer(rawValue));
   if (cl_val.type == ClarityType.Tuple) {
@@ -130,6 +135,7 @@ export function parseNamespaceRawValue(
       status: status,
       latest: true,
       buckets: buckets.toString(),
+      tx_id: txid,
     };
     return namespaceBNS;
   }
@@ -153,8 +159,8 @@ export async function parseContentHash(contentHash: string): Promise<string> {
     throw Error('Error: can not get content hash');
   }
   let content = '';
-  for (const char of result.attachment.content) {
-    content = content + char.toString(16);
+  for (const ascii of result.attachment.content) {
+    content = content + String.fromCharCode(ascii);
   }
 
   return content;
--- a/src/bns-helpers.ts
+++ b/src/bns-helpers.ts
@@ -1,0 +1,182 @@
+import {
+  deserializeCV,
+  ClarityType,
+  ClarityValue,
+  BufferCV,
+  StandardPrincipalCV,
+  TupleCV,
+  BufferReader,
+  Address,
+  IntCV,
+  addressToString,
+  StringAsciiCV,
+  SomeCV,
+  UIntCV,
+  ListCV,
+} from '@stacks/transactions';
+import { DbBNSNamespace } from './datastore/common';
+import { hexToBuffer } from './helpers';
+import BN = require('bn.js');
+import { CoreNodeParsedTxMessage } from './event-stream/core-node-message';
+import { TransactionPayloadTypeID } from './p2p/tx';
+import { StacksCoreRpcClient } from './core-rpc/client';
+
+export interface Attachment {
+  attachment: {
+    hash: string;
+    metadata: {
+      name: string;
+      namespace: string;
+      tx_sender: Address;
+    };
+    page_index: number;
+    position_in_page: string;
+  };
+}
+
+interface AttachmentValue {
+  attachment: {
+    content: number[];
+  };
+}
+
+export function parseNameRawValue(rawValue: string): Attachment {
+  const cl_val: ClarityValue = deserializeCV(hexToBuffer(rawValue));
+  if (cl_val.type == ClarityType.Tuple) {
+    const attachment = cl_val.data['attachment'] as TupleCV;
+
+    const hash: BufferCV = attachment.data['hash'] as BufferCV;
+    const contentHash = hash.buffer.toString('hex');
+
+    const metadataCV: TupleCV = attachment.data['metadata'] as TupleCV;
+
+    const nameCV: BufferCV = metadataCV.data['name'] as BufferCV;
+    const name = nameCV.buffer.toString();
+    const namespaceCV: BufferCV = metadataCV.data['namespace'] as BufferCV;
+    const namespace = namespaceCV.buffer.toString();
+    const addressCV: StandardPrincipalCV = metadataCV.data['tx-sender'] as StandardPrincipalCV;
+    const address = addressCV.address;
+
+    const page_indexCV: IntCV = metadataCV.data['namespace'] as IntCV;
+    const page_index = page_indexCV.value;
+    const position_in_pageCV: IntCV = metadataCV.data['namespace'] as IntCV;
+    const position_in_page = position_in_pageCV.value;
+
+    const result: Attachment = {
+      attachment: {
+        hash: contentHash,
+        metadata: {
+          name: name,
+          namespace: namespace,
+          tx_sender: address,
+        },
+        page_index: page_index,
+        position_in_page: position_in_page,
+      },
+    };
+    return result;
+  }
+  throw Error('Invalid clarity type');
+}
+
+export function parseNamespaceRawValue(
+  rawValue: string,
+  readyBlock: number
+): DbBNSNamespace | undefined {
+  const cl_val: ClarityValue = deserializeCV(hexToBuffer(rawValue));
+  if (cl_val.type == ClarityType.Tuple) {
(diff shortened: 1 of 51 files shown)


noumanahmads:
Adding some more tests to improve the test coverage.

noumanahmads:
@zone117x @lgalabru  
This PR is ready to review.

src/bns-helpers.ts
```diff
@@ -0,0 +1,182 @@
...
+  let result: AttachmentValue | undefined = undefined;
+  try {
+    result = await new StacksCoreRpcClient().fetchJson<AttachmentValue>(
+      `v2/attachments/${contentHash}`,
+      {
+        method: 'GET',
+        timeout: 10 * 1000, //10 seconds
+      }
+    );
+  } catch (error) {}
```
zone117x, line 159: Why hide the error here then throw below?

src/bns-helpers.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  if (result === undefined) {
+    throw Error('Error: can not get content hash');
+  }
+  let content = '';
+  for (const ascii of result.attachment.content) {
+    content = content + String.fromCharCode(ascii);
```
zone117x, line 166: This doesn't look right. Is `attachment.content` an octet array? If so, this should be parsed with something like:
```ts
const content = Buffer.from(result.attachment.content).toString('ascii');
````

However, an octet array response likely a bug in the core node. Binary content in JSON responses should be encoded as a hex string.

File: src/event-stream/event-server.ts [hunk 3]
```diff
@@ -190,6 +207,44 @@ async function handleClientMessage(msg: CoreNodeBlockMessage, db: DataStore): Pr
...
+              zonefile_hash: attachment.attachment.hash,
+              zonefile: attachmentValue,
+              latest: true,
+              tx_id: event.txid,
+              status: attachment.attachment.metadata.op,
+              index_block_hash: parsedMsg.index_block_hash,
+              canonical: true,
+            };
+            console.log('update names ', JSON.stringify(names));
+            await db.updateNames(names);
```
zone117x, line 234: Performing database updates here can cause data integrity issues. This should instead be something like `dbData.names.push(name)`, then the `await db.update(dbData)` function called at the end of this function should be performing these updates within its postgres transaction.

src/event-stream/event-server.ts
```diff
@@ -190,6 +207,44 @@ async function handleClientMessage(msg: CoreNodeBlockMessage, db: DataStore): Pr
...
+          } else if (functionName === namespaceReadyFunction) {
+            //event received for namespaces
+            const namespace: DbBNSNamespace | undefined = parseNamespaceRawValue(
+              event.contract_event.raw_value,
+              parsedMsg.block_height,
+              event.txid,
+              parsedMsg.index_block_hash
+            );
+            if (namespace != undefined) {
+              await db.updateNamespaces(namespace);
```
zone117x, line 244: Same thing here: should be `dbData.namespaces.push(namespace)`

src/datastore/postgres-store.ts
```diff
@@ -2665,6 +2711,176 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
     return { results };
   }
 
+  async updateNames(bnsName: DbBNSName) {
```
zone117x, line 2714: This function should accept a `client` as the first argument, i.e.
```ts
async updateName(client: ClientBase, bnsName: DbBNSName) { ... }
```
Then be called from the `async update(data: DataStoreUpdateData) { ... }` where it is already in a sql transaction.

src/datastore/postgres-store.ts
```diff
@@ -2665,6 +2711,176 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
+    } catch (error) {
+      logError(`Error performing PG update: ${error}`, error);
+      await client.query('ROLLBACK');
+      throw error;
+    } finally {
+      client.release();
+    }
+  }
+
+  async updateNamespaces(bnsNamespace: DbBNSNamespace) {
```
zone117x, line 2767: Same with this function -- should be called with an existing client connection

src/event-stream/event-server.ts
```diff
@@ -190,6 +207,44 @@ async function handleClientMessage(msg: CoreNodeBlockMessage, db: DataStore): Pr
...
         };
         dbTx.contractLogEvents.push(entry);
+        if (
+          event.contract_event.topic === printTopic &&
+          event.contract_event.contract_identifier === bnsContractIdentifier
+        ) {
+          const functionName = getFunctionName(event.txid, parsedMsg.parsed_transactions);
+          if (nameFunctions.includes(functionName)) {
+            const attachment = parseNameRawValue(event.contract_event.raw_value);
+            const attachmentValue = await parseContentHash(attachment.attachment.hash);
```
zone117x, line 217: Nit: this function should be named something like `fetchAttachmentContent`

zone117x: @lgalabru can you weigh in on the safety here? Performing a fetch here results in the following:
* Core node POSTs event with bns update to sidecar, waits for HTTP 200 response.
* Sidecar sees this bns data, then performs this GET `/v2/attachments/<hash>`  back to the core node server.
* Core node replies to the attachment request, sidecar finishes and responds to the initial event POST request.

So in a roundabout way, the core node ends up waiting on it's own http request and server response. Could this cause any race conditions?

src/bns-helpers.ts
lgalabru: Correct, `v2/attachment/:hash` will return binary contents.
Is the content-type `application/json`? That would be a bug

src/event-stream/event-server.ts
lgalabru: Yeah that feels weird. The attachment could be embedded in the json instead.

src/bns-helpers.ts
zone117x: The JSON response appears to contain something like:
```ts
interface Attachment {
  content: number[]; // octet array
}
```
as opposed to what I'd expect:
```ts
interface Attachment {
  content: string; // hex string
}
```

src/event-stream/event-server.ts
noumanahmads: Updated this as suggested.

src/event-stream/event-server.ts
noumanahmads: Updated this as suggested.

src/datastore/postgres-store.ts
noumanahmads: Updated this as suggested

src/event-stream/event-server.ts
noumanahmads: @zone117x 
Changed the function name here for now.

src/datastore/postgres-store.ts
noumanahmads: Added client as the first argument.

src/bns-helpers.ts
noumanahmads: For now, I have made the changes using content as an octet array.

src/bns-helpers.ts
noumanahmads: Updated this by throwing the error in the catch block.

src/datastore/common.ts
```diff
@@ -256,7 +256,11 @@ export interface DataStoreUpdateData {
     nftEvents: DbNftEvent[];
     contractLogEvents: DbSmartContractEvent[];
     smartContracts: DbSmartContract[];
+    names: DbBNSName[];
+    namespaces: DbBNSNamespace[];
   }[];
+  names: DbBNSName[];
+  namespaces: DbBNSNamespace[];
```
zone117x, line 263: Only one set of `names` and `namespaces` should be needed for this interface right? It looks like the ones within the `txs` scope makes the most sense in case we need to correlate name data with their associated tx later on.

zone117x:
@noumanahmads do you have any scripts I could use for manual end to end testing of namespace registration?

src/datastore/common.ts
noumanahmads: Yes, I have removed these extra variables and these you have mentioned above in this [commit](https://github.com/blockstack/stacks-blockchain-api/pull/363/commits/f2712fcb1dd8ba558cdcb0a38def2fc3861f4a5c).

noumanahmads:
> @noumanahmads do you have any scripts I could use for manual end to end testing of namespace registration? The tests currently don't cover the event stream interaction between the core-node and API.

@zone117x 
**Setup the environment**
1. start the miner using this command.
`env DYNAMIC_GENESIS_TIMESTAMP=1 cargo run local-leader.toml.default `
[local-leader.toml.default](https://gist.github.com/noumanahmads/96c9bc5e522bdcbab71f15035c2d4b41#file-local-leader-toml-default)
2. copy the b.yml and p.yml in stacks blockchain api and start these using this command
`docker-compose -f b.yml -f p.yml up`
[b.yml](https://gist.github.com/noumanahmads/96c9bc5e522bdcbab71f15035c2d4b41#file-b-yml)
[p.yml](https://gist.github.com/noumanahmads/96c9bc5e522bdcbab71f15035c2d4b41#file-p-yml)
3. start the node using this command
`env STACKS_EVENT_OBSERVER=localhost:3700 target/debug/stacks-node start --config=bns-local.conf`
[bns-local-conf](https://gist.github.com/noumanahmads/96c9bc5e522bdcbab71f15035c2d4b41#file-bns-local-conf)
4. run the stacks-blockchain-api using this 
`npm run dev`

**Testing script**
Use this script [hash.ts](https://gist.github.com/noumanahmads/96c9bc5e522bdcbab71f15035c2d4b41#file-hash-ts). We have to execute the functions one by one from line 348. 

I have pasted this file in stacks-blockchain-api project and added this `"bns": "ts-node hash.ts"` in package.json under scripts to execute this file. 

Uncomment the line on 348 i.e `nspreorder()`. Run the script using this command `npm run bns`.
After the success of this transaction, uncomment the next line and execute that. In this way we should proceed to `nameImport()` and `nameReady()` functions.

src/event-stream/event-server.ts
```diff
@@ -402,6 +455,11 @@ export async function startEventServer(opts: {
     }
   });
 
+  app.postAsync('/attachments/new', (req, res) => {
```
zone117x, line 458: Can we add a `// TODO: ...` comment here?

src/datastore/postgres-store.ts
```diff
@@ -2679,6 +2731,150 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
+        status,
+        latest,
+        tx_id,
+        canonical,
+        index_block_hash,
+      ]
+    );
+  }
+
+  async getNamespaceList() {
```
zone117x, line 2827: This `SELECT` and the couple additional below need a `WHERE canonical = true`. We can just add a `// TODO: fix canonical query` comment for now to merge

Issue #365, opened 2020-12-15
Permanent fix for the stx-mint event limit


zone117x:
In reference to PR https://github.com/blockstack/stacks-blockchain-api/pull/364

> A couple endpoints that return the stx-mint events for the genesis transaction are currently extremely resource intensive due to loading in ~330k at once. This is a temporary fix by adding a 200 limit to the amount of stx-asset events returned for a given transaction.

> A permanent fix is could be to simply not return events for these special massive genesis transactions, as I don't think there's a real need to display them. Alternatively, we can implement per-transaction event pagination. This will require some discussion.

PR #367, opened 2020-12-18
Add pagination for tx events #365

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/365

Implement pagination for transaction events. 

The endpoint `/extended/v1/tx/<txid>` can now be passed query parameters `event_offset` and `event_limit`. If unspecified, the default is `96`.

### DIFF
--- a/client/src/generated/apis/TransactionsApi.ts
+++ b/client/src/generated/apis/TransactionsApi.ts
@@ -33,6 +33,8 @@ export interface GetMempoolTransactionListRequest {
 
 export interface GetTransactionByIdRequest {
     txId: string;
+    eventOffset?: number;
+    eventLimit?: number;
 }
 
 export interface GetTransactionListRequest {
@@ -73,6 +75,8 @@ export interface TransactionsApiInterface {
      * Get a specific transaction by ID  `import type { Transaction } from \'@blockstack/stacks-blockchain-api-types\';` 
      * @summary Get transaction
      * @param {string} txId Hash of transaction
+     * @param {number} [eventOffset] The number of events to skip
+     * @param {number} [eventLimit] The numbers of events to return
      * @param {*} [options] Override http request option.
      * @throws {RequiredError}
      * @memberof TransactionsApiInterface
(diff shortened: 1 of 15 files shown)


kyranjamie (approved):

kyranjamie:
One place I am using events is to sum the total spend of a tx, which requring iterating events.

```ts
export function sumStxTxTotal(address: string, tx: Transaction, poxContractId?: string) {
  const dir = getStxTxDirection(address, tx);
  if (tx.tx_type === 'token_transfer') {
    return new BigNumber(tx.token_transfer.amount).plus(dir === 'sent' ? tx.fee_rate : 0);
  }
  if (tx.tx_type === 'coinbase' || tx.tx_type === 'poison_microblock') {
    return new BigNumber(tx.fee_rate);
  }

  if (isStackingTx(tx, poxContractId) && tx.tx_result?.repr) {
    // We get the amount stacked from the tx_result
    const matcher = /\(tuple \(lock-amount\su(\d+)/;
    const matches = matcher.exec(tx.tx_result.repr);
    if (matches && matches[1]) {
      return new BigNumber(matches[1], 10);
    }
  }

  const initialValue = new BigNumber(0);
  const sumEventTransferHandler = (prev: BigNumber, current: TransactionEvent) =>
    current.event_type === 'stx_asset' && current.asset.asset_event_type === 'transfer'
      ? new BigNumber(current.asset.amount || 0).plus(prev)
      : initialValue;
  return tx.events.reduce(sumEventTransferHandler, initialValue).plus(tx.fee_rate);
}
```

Is there anyway we can do this API side so the client doesn't have to iterate paginated results?

zone117x:
@kyranjamie Yes that looks doable. Can you create an issue with a new API endpoint that you think makes sense?

kyranjamie:
Noted here ☝🏼 

Ideally it won't be another request, though I fear this won't be possible.

zone117x:
Think we can merge this for now and work on https://github.com/blockstack/stacks-blockchain-api/issues/369 ?
I think a transaction with over 96 (default pagination size) stx transfer events is quite an edge-case right now

kyranjamie:
Yeah sure, will open an issue in wallet

PR #373, opened 2020-12-23
Send stx faucet requests to both miner and follower

zone117x:
Reference https://github.com/blockstack/stacks-blockchain-api/issues/359

* Performs nonce incrementing without the previously hardcoded limit.
* Broadcast transaction to both miner and follower if both are configured.


kantai (approved):
This looks good to me -- this should improve things for the faucet for now. There are still some bad cases where a transaction gets stuck in the network, but I think that this will be able to prevent most problems we're currently seeing.

PR #375, opened 2020-12-28
fix: miner reward event parse error

zone117x:
Fixed a bug with API getting stuck around block 101 on current Xenon testnet. A parse error in the miner reward event handler was stalling block processing.


kantai (approved):
LGTM!

And apologies for not catching this change in the event stream format

PR #388, opened 2021-01-07
feat: bns import script

noumanahmads:
## Description
Import script for BNS

For details refer to [issue](https://github.com/blockstack/stacks-blockchain-api/issues/317)

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

## Testing information
No

@zone117x  for review


fariedt:
This approach works for namespaces and names, but not for subdomains ([code](https://gist.github.com/fariedt/a86c12b4fd9e746f25903d14538cecb9)).  It's just too slow.  It takes 8+ seconds for every 1,000 subdomans it imports, and there are 2,178,704 subdomains.

Preprocessing -- I used a [simple script](https://gist.github.com/fariedt/6da18cf78c26acd3248cd5cffcb74273) to merge the `subdomains.csv` and `subdomain_zonefiles.txt` files, and then tried importing them.  Using node js's streams, it avoids loading everything into memory; the problem is with the db inserts.

I was able to work around that in the past with my script that wrote out a bunch of files that I could import with `\copy` in psql.  I will attempt to use [pg-copy-streams](https://github.com/brianc/node-pg-copy-streams) to see if I can safely import the (preprocessed) data.

src/importV1/index.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+  });
+
+  if (!bnsImport) return;
+
+  logger.info('legacy BNS data import started');
+
+  // check if the files we need can be read
+  try {
+    fs.accessSync(`${importDir}/chainstate.txt`, fs.constants.R_OK);
```
zone117x, line 166: Need's a `// TODO: validate contents with their .sha256 files`

src/index.ts
```diff
@@ -51,6 +52,9 @@ async function init(): Promise<void> {
     }
   }
   const promMiddleware = isProdEnv ? createPrometheusMiddleware() : undefined;
+  if ('pool' in db) {
```
zone117x, line 55: `// TODO: use "if (db instanceof PgDataStore)"`

src/importV1/index.ts
```diff
@@ -0,0 +1,192 @@
+import * as fs from 'fs';
```
zone117x, line 1: Nit: `// TODO: file naming consistency, should be 'src/import-v1/'`

zone117x:
Looks great! Couple comments for TODOs then good to merge

zone117x:
> This approach works for namespaces and names, but not for subdomains ([code](https://gist.github.com/fariedt/a86c12b4fd9e746f25903d14538cecb9)). It's just too slow. It takes 8+ seconds for every 1,000 subdomans it imports, and there are 2,178,704 subdomains.

@fariedt I had huge performance improvements by prepared statements and batching inserts. The code is fairly generalized, you should be able to re-use it.

See here: https://github.com/blockstack/stacks-blockchain-api/blob/338754b16f05f9a0e2a596fe47bf28254d262bcb/src/datastore/postgres-store.ts#L1789-L1827

src/importV1/index.ts
```diff
@@ -0,0 +1,192 @@
...
+        this.state = '';
+      }
+    } else if (line.startsWith('-----END')) {
+      if (this.state.startsWith('name') && this.rowCount > 0) {
+        logger.info(`${this.tag}: ${this.state} rows added: ${this.rowCount}`);
+      }
+      this.state = '';
+      this.rowCount = 0;
+    } else if (this.state != '') {
+      this.client = await this.db.pool.connect();
```
zone117x, line 68: Connecting and releasing a client each time probably slows this down quite a bit. It should use one client for the full duration of the import.

src/index.ts
faried: This is fixed.

src/importV1/index.ts
faried: Renamed.

src/importV1/index.ts
faried: I found it easier to batch up the subdomain objects in a stream reader function, and then perform a bulk insert.  I couldn't figure out how to write something like `batchIterateAsync`, but it works the same way, albeit not as cleanly.  I'll look into it some more.

It does a little over 2000 inserts/second right now.

PR #389, opened 2021-01-11
Feat/bns/names

noumanahmads:
## Description
Following endpoints are implemented:
- /names
- /names/:name
For details refer to [issue](https://github.com/blockstack/stacks-blockchain-api/issues/317)

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

## Testing information
Test cases are added for the endpoints implemented.
To run the test cases `npm run test:bns`

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review


src/api/routes/bns/names.ts
```diff
@@ -39,5 +42,39 @@ export function createBNSNamesRouter(db: DataStore): RouterWithAsync {
...
+      blockchain: bnsBlockchain,
+      expire_block: result.expire_block,
+      grace_period: result.grace_period,
+      last_txid: result.tx_id ? result.tx_id : '',
+      resolver: result.resolver,
+      status: result.status ? result.status : '',
+      zonefile: result.zonefile,
+      zonefile_hash: result.zonefile_hash,
+    };
+    const response = Object.fromEntries(Object.entries(nameInfo).filter(([_, v]) => v != null));
```
zone117x, line 75: `// TODO: loose null comparison hard to reason about`

src/datastore/postgres-store.ts
```diff
@@ -2882,7 +2882,9 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
       `
       SELECT *
       FROM names
-      WHERE name = $1
+      WHERE canonical = true
+      AND 
+      name = $1
```
zone117x, line 2887: `// TODO: latest = true`
right?

src/api/routes/bns/names.ts
noumanahmads: Here I am removing the keys which have null values in the `BNSGetNameInfoResponse` object. 
If you can suggest a better approach that would be great or should I add a comment for the logic. 
@zone117x

src/datastore/postgres-store.ts
noumanahmads: right, fixed this one.

src/api/routes/bns/names.ts
zone117x: There's a loose comparison `!=` against null. I would expect this should be disabled by the linter, because it's not always obvious what js values can be coerced into.

PR #395, opened 2021-01-15
feat: config to specifying a different core node for the /v2 proxy

zone117x:


.env
```diff
@@ -13,6 +13,9 @@ STACKS_BLOCKCHAIN_API_HOST=127.0.0.1
 STACKS_CORE_RPC_HOST=127.0.0.1
 STACKS_CORE_RPC_PORT=20443
 
+# STACKS_CORE_PROXY_HOST=127.0.0.1
```
CharlieC3, line 16: Just double checking but this ENV can accept a FQDN, right?

zone117x: Yep 👍

CharlieC3 (approved):
:shipit:

PR #398, opened 2021-01-16
fix: create mock tx from event for BTC tx's

### LINKED ISSUE #396: Bitcoin-broadcasted operation events should be handled
STX transfers can be performed via on-bitcoin transactions. These produce normal stacks transfer events, but associate them with a transaction receipt formatted like:

```
{
            "txid": "<bitcoin-txid>",
            "tx_index": 1,
            "status": "success",
            "raw_result": "0x...",
            "raw_tx": "0x00",
            "contract_abi": null,
}
```

An example transaction:

https://www.blockchain.com/btc/tx/c0263c14f689ae757290f90765a28314497f52bd22b8bcbf87a12c764dd9d9be

hstove:
I've ran this and was able to get past block 119. However, when fetching the BTC tx from `/v1/tx/:txid`, there was an error around how I was storing the `postConditions` buffer. I'm currently re-running with that fix to make sure that works.

Fixes #396 , which was halting progress on the api at block 119

### DIFF
--- a/src/event-stream/reader.ts
+++ b/src/event-stream/reader.ts
@@ -2,7 +2,6 @@ import {
   CoreNodeBlockMessage,
   CoreNodeMessageParsed,
   CoreNodeParsedTxMessage,
-  StxTransferEvent,
 } from './core-node-message';
 import {
   readTransaction,
@@ -10,10 +9,6 @@ import {
   RecipientPrincipalTypeId,
   Transaction,
   TransactionAuthTypeID,
-  SigHashMode,
-  TransactionPublicKeyEncoding,
-  TransactionAnchorMode,
-  TransactionPostConditionMode,
 } from '../p2p/tx';
 import { NotImplementedError } from '../errors';
 import { getEnumDescription, logger, logError } from '../helpers';
@@ -24,9 +19,6 @@ import {
   addressToString,
   AddressHashMode,
   BufferReader,
-  ChainID,
-  createSingleSigSpendingCondition,
-  createAddress,
 } from '@stacks/transactions';
 import { c32address } from 'c32check';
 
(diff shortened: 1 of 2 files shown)


src/event-stream/reader.ts
```diff
@@ -67,9 +75,53 @@ export function parseMessageTransactions(msg: CoreNodeBlockMessage): CoreNodeMes
...
+      try {
+        rawTx = readTransaction(bufferReader);
+        txSender = getTxSenderAddress(rawTx);
+        sponsorAddress = getTxSponsorAddress(rawTx);
+      } catch (error) {
+        logError(
+          `BTC transaction found. Recreating from event. TX: ${JSON.stringify(coreTx)}: ${error}`,
+          error
+        );
+        const event = msg.events[i] as StxTransferEvent;
```
lgalabru, line 90: should we instead check that we're looking at a "raw_tx = `0x00`" and only build that Tx if that's the case?

hstove: Yep, probably!

lgalabru (approved):

zone117x:
Thanks for the fix @hstove! I rebased with your cleanup commit, and fixed a bug where the stx events would sometimes be matched against the wrong tx. I tested locally up to block ~400.

src/event-stream/reader.ts
```diff
@@ -74,53 +112,30 @@ export function parseMessageTransactions(msg: CoreNodeBlockMessage): CoreNodeMes
...
       const txBuffer = Buffer.from(coreTx.raw_tx.substring(2), 'hex');
-      const bufferReader = BufferReader.fromBuffer(txBuffer);
-      let txSender: string;
-      let sponsorAddress = undefined;
       let rawTx: Transaction;
-      try {
+      let txSender: string;
+      let sponsorAddress: string | undefined;
+      if (coreTx.raw_tx === '0x00') {
+        const event = msg.events.find(event => event.txid === coreTx.txid);
```
zone117x, line 119: This was previously doing `event = msg.events[i]` where `i` was the tx index rather than event index.

Issue #402, opened 2021-01-18
Events missing from STX transfer transactions


kyranjamie:
STX token transfer events are appearing without any events. See these two example transactions:

https://explorer.stacks.co/txid/0x60c2c8c9386473c4621ec3dfa638eae49e15cc307e975df6e29d5bd95b918548?chain=mainnet
https://explorer.stacks.co/txid/0xa118e363c3db0214cef7d4da7ccc120f8df7c94aa48c2c2d2009cf3e6b296976?chain=mainnet

It is expected the API returns an events array containing at least the 1 token transfer event.

This issue was discovered [in relation to this user's deposit issues on Discord](https://discord.com/channels/621759717756370964/621759718192447502/800671101889151007). Unconfirmed from the user whether or not the funds were eventually deposited.

zone117x:
This turns out to be a pagination bug. The events are not missing, they are behind the next page https://stacks-node-api.mainnet.stacks.co/extended/v1/tx/0x60c2c8c9386473c4621ec3dfa638eae49e15cc307e975df6e29d5bd95b918548?event_offset=96

`event_index` appears to be using some absolute value within the block as opposed to scoped to a specific transaction, making the first "page" of events for some transactions appear empty.

Working on a fix -- in the meantime, integrators can page each request a few times as a workaround.

PR #405, opened 2021-01-19
fix: #402 tx events not showing due to pagination issue

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/402

Transaction event indexes are received scoped to block level rather than a transaction. E.g. if a block has 100 transactions, with one or more events per transaction, then transactions N+1 will have event indexes started _after_ 0. 

This resulted in some STX transfer transactions, which should only have 1 event, not showing the event because it was pushed 1 or more pages down when queried. 

This PR modifies the event server to re-calculate event indexes on a per-transaction level.


kyranjamie (approved):
Love some chained array methods

zone117x:
Map mutations as a side-effect. What am I doing, gross.. changing to an explicit for loop..

PR #407, opened 2021-01-20
Tx mempool address filters

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/380

```yaml

  /extended/v1/tx/mempool:
    get:
      summary: Get mempool transactions
      tags:
        - Transactions
      operationId: get_mempool_transaction_list
      description: Get all recently-broadcast mempool transactions
      parameters:
        - name: sender_address
          in: query
          description: Filter to only STX transfer transactions with this sender address.
          required: false
          schema:
            type: string
        - name: recipient_address
          in: query
          description: Filter to only STX transfer transactions with this recipient address.
          required: false
          schema:
            type: string
        - name: address
          in: query
          description: Filter to only show STX transfer transactions with this address as the recipient or sender.
          required: false
          schema:
            type: string
        - name: limit
          in: query
          description: max number of mempool transactions to fetch
          required: false
          schema:
            type: integer
        - name: offset
          in: query
          description: index of first mempool transaction to fetch
          required: false
          schema:
            type: integer
```

### DIFF
--- a/client/src/generated/apis/TransactionsApi.ts
+++ b/client/src/generated/apis/TransactionsApi.ts
@@ -27,6 +27,9 @@ import {
 } from '../models';
 
 export interface GetMempoolTransactionListRequest {
+    senderAddress?: string;
+    recipientAddress?: string;
+    address?: string;
     limit?: number;
     offset?: number;
 }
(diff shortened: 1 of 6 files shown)


kyranjamie (approved):

agraebe:
was this pushed live to the deployed docs? I don't see it yet

PR #412, opened 2021-01-24
feat: normalize express prometheus metrics with route parsing

zone117x:
Fixes https://github.com/blockstack/stacks-blockchain-api/issues/403

This converts the raw paths received by the `@promster/express` middleware into Express path template strings.

1. First, [`express-list-endpoints`](https://www.npmjs.com/package/express-list-endpoints) is used to enumerate all the application defined API route templates.
2. Then [`path-to-regexp`](https://www.npmjs.com/package/path-to-regexp) is used to map the Express route strings into a set of regexps.
3. The promster middleware runs the raw request URL paths through this set of regexes to determine the route template.

The resulting raw paths, e.g.
`/extended/v1/address/ST26DR4VGV507V1RZ1JNM7NN4K3DTGX810S62SBBR/stx?asdf=1234`
are now reported by their path template, .e.g.
`/extended/v1/address/:stx_address/stx `

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19642,6 +19642,11 @@
         "vary": "~1.1.2"
       },
       "dependencies": {
+        "path-to-regexp": {
+          "version": "0.1.7",
+          "resolved": "https://registry.npmjs.org/path-to-regexp/-/path-to-regexp-0.1.7.tgz",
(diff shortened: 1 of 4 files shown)


kyranjamie (approved):
image

PR #413, opened 2021-01-25
feat: env var override for /v2/pox min_amount_ustx

zone117x:
Feature to override the `min_amount_ustx` value returned by the proxied `/v2/pox` endpoint.

To enable, set the env var `V2_POX_MIN_AMOUNT_USTX` to the number that should be used to overwrite the original `min_amount_ustx` value from the stacks-node response.

For example, setting `V2_POX_MIN_AMOUNT_USTX=123` results in `/v2/pox` returning:
```json
{
    "contract_id": "SP000000000000000000002Q6VF78.pox",
    "first_burnchain_block_height": 666050,
    "min_amount_ustx": 123,    <--- SEE HERE
    "prepare_cycle_length": 100,
    "rejection_fraction": 25,
    "reward_cycle_id": 0,
    "reward_cycle_length": 2100,
    "rejection_votes_left_required": 40379486728792,
    "total_liquid_supply_ustx": 1009487168219817,
    "next_reward_cycle_in": 517
}
```


hstove (approved):
quite a hack indeed

CharlieC3 (approved):
:shipit:

PR #414, opened 2021-01-25
feat: handling for on-btc-chain stx-stacks operations

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/411

PR adds support for the on-btc-chain `stx-stacks` operation. A synthetic pox contract call tx is generated for the operation event, similar to the previous on-btc-chain stx-transfer operation.

This was tested by grabbing event data generated by this stacks-node integration test: https://github.com/blockstack/stacks-blockchain/commit/d163d5cf2c817d83d7d120f8f988d9abf0a684be and injecting it into the API while running in mocknet mode.

### DIFF
--- a/src/event-stream/event-server.ts
+++ b/src/event-stream/event-server.ts
@@ -103,6 +103,15 @@ async function handleClientMessage(
 ): Promise<void> {
   const parsedMsg = parseMessageTransactions(chainId, msg);
 
+  // Delete on-btc-chain transactions that failed, there is no useful data in them,
+  // and the db and API schemas can't currently fit these types of transactions.
+  for (let i = parsedMsg.transactions.length - 1; i >= 0; i--) {
+    if (!parsedMsg.parsed_transactions[i]) {
+      parsedMsg.parsed_transactions.splice(i, 1);
+      parsedMsg.transactions.splice(i, 1);
+    }
+  }
+
   const dbBlock: DbBlock = {
     canonical: true,
     block_hash: parsedMsg.block_hash,
--- a/src/event-stream/event-server.ts
+++ b/src/event-stream/event-server.ts
@@ -31,7 +31,7 @@ import {
 } from '../datastore/common';
 import { parseMessageTransactions, getTxSenderAddress, getTxSponsorAddress } from './reader';
 import { TransactionPayloadTypeID, readTransaction } from '../p2p/tx';
-import { BufferReader } from '@stacks/transactions';
+import { BufferReader, ChainID } from '@stacks/transactions';
 
 async function handleBurnBlockMessage(
   burnBlockMsg: CoreNodeBurnBlockMessage,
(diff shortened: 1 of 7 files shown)


src/event-stream/reader.ts
```diff
@@ -121,8 +203,16 @@ export function parseMessageTransactions(msg: CoreNodeBlockMessage): CoreNodeMes
           throw new Error(`Could not find txid for process BTC tx: ${JSON.stringify(msg)}`);
```
kantai, line 203: I think this shouldn't throw -- bitcoin operations could _fail_, but still be processed and result in a tx receipt (e.g., a stack-stx op that doesn't stack enough, or the caller doesn't have enough unlocked STX) -- these txs just wouldn't have events, though. I think the thing to do here would be to log some information and continue processing.

zone117x: Fixed. An empty/failed btc transaction like this can't currently be fit into the current db schema or API schemas, so they are simply ignored.

src/event-stream/reader.ts
```diff
@@ -66,12 +76,81 @@ export function getAddressFromPublicKeyHash(
...
+  const rewardCycleLength = chainId === ChainID.Mainnet ? 2100 : 50;
+  const lockPeriod = Math.floor((unlockBurnHeight - burnBlockHeight) / rewardCycleLength);
+  const senderAddress = createAddress(event.stx_lock_event.locked_address);
+  const poxAddress = createAddress(
+    chainId === ChainID.Mainnet ? 'SP000000000000000000002Q6VF78' : 'ST000000000000000000002AMW42H'
+  );
+
+  const clarityFnArgs: ClarityValue[] = [
+    lockAmount,
+    tupleCV({
```
kantai, line 101: Can you open an issue to track mocking a more accurate argument here -- this address should be the first output address of the bitcoin transaction. I realize that the `stacks-node` doesn't currently provide enough information to obtain this, so changes would be required there as well, but we should discuss in a new issue.

kantai (approved):
This looks good to me -- we should start an issue for generating a more accurate transaction for the StackStxOp though.

src/event-stream/reader.ts
zone117x: Created https://github.com/blockstack/stacks-blockchain-api/issues/415

src/event-stream/event-server.ts
```diff
@@ -96,8 +96,21 @@ async function handleMempoolTxsMessage(rawTxs: string[], db: DataStore): Promise
...
+  const parsedMsg = parseMessageTransactions(chainId, msg);
+
+  // Delete on-btc-chain transactions that failed, there is no useful data in them,
+  // and the db and API schemas can't currently fit these types of transactions.
+  for (let i = parsedMsg.transactions.length - 1; i >= 0; i--) {
+    if (!parsedMsg.parsed_transactions[i]) {
+      parsedMsg.parsed_transactions.splice(i, 1);
+      parsedMsg.transactions.splice(i, 1);
+    }
+  }
```
hstove, line 113: There is no way for these transactions to have an event, right? Because otherwise some downstream code will throw, and progress will halt.

hstove (approved):

src/event-stream/event-server.ts
zone117x: From Aaron earlier:
> I'm pretty sure that tx receipts will be generated even for bitcoin operations that fail (and therefore don't have events)

So probably okay? 🤷

PR #441, opened 2021-02-11
feat: new api endpoint to get inbound stx and send-many transfers w/ memos

hstove:
This PR exposes a new API endpoint, `/extended/v1/address/:address/inbound`. It pairs with the `send-many-memo` contract, and [the CLI](https://github.com/hstove/send-many-stx-cli) for that.

It basically looks for STX transfer tx's, as well as `stx-transfer` events from the hard-coded (configurable) contract. It merges those two into a unified, paginatable API.

The API payload is pretty sparse right now. I think we could get it to a point where the API response is almost identical to the `transactions` API. The goal here is to have something that exchanges can consume - so they probably need more data than this.

Feel free to make whatever changes you want - especially around the API naming, props, etc.


yknl:
Is there a particular reason that the transactions from the multiple send contract cannot be shown in the regular `/extended/v1/address/:address/transactions` endpoint? For wallet integrators it would be good to have one single endpoint for transactions.

zone117x:
@yknl `/extended/v1/address/:address/transactions` can be adapted to show the `send-many-memo ` contract-call transactions associated with the address -- but that will not be plug-and-play for integrators. They'd need to manually parse out the contract-call tx event logs to determine the set of stx transfers and associated memo strings.

yknl:
This sounds like a problem for integrators. If not everyone updates to this new endpoint, the `send-many-memo` function cannot be said to work properly because the receiving integrator might not have switched to the new endpoint which can read the memo. If it doesn't work reliably 100% of the time I think we'd have a hard time selling it as a working feature. Unless of course you can convince every integrator to immediately upgrade and use this new endpoint.

hstove:
@yknl I totally agree that it's a tricky issue. I really don't think we should monkey-patch the regular transactions API just to include these custom transfers with memos. I do think we should include transactions with asset transfers for the address in the transactions API, but even then, it requires custom code to handle a memo, which is really just a `print` event before a `stx-transfer` event.

If we need integrators to update their code (which they will), then why not just have them use this new endpoint, which is specifically for integrators to handle incoming deposits?

I agree though that the `send-many-memo` approach might not work. It's pretty messy. I just think that if we were going to do this contract-based approach, I'd prefer this new API. Maybe we just do a queue after all...

zone117x:
A queue has its own set of problems. For one, it doesn't actually solve the problem sending more than 25 transactions per block. Rather, it's only helper that still requires integrators to build a pool of funded addresses, use a new "latest-queued-nonce" endpoint, handle "pending-pending" transactions, handle a new class of errors for when queued transaction have mempool admission failures, etc. 

All things considered, I think this send-many-memo contract call is the best approach.

zone117x:
Latest commit adds support for `/extended/v1/address/:address/transactions`. Specifically, the endpoint now shows contract-call txs that have stx transfers associated with the given address. When an integrator comes across one of these transactions, it can simply grab the associated `stx_asset` event(s) from the event array to show the send/receive address pairs and amounts. Additionally, they can add support for extracting memos when the tx contract is `send-many-memo`.

hstove:
This looks good :+1:

So now integrators have two options:

- The `/stx_inbound` api, which is quite specialized but gets the job done for 'what incoming STX transfers (with memos) are there)
- The `/transactions` api, which is the same as before. Integrators should be given guidance that, in order to get memos, they need to parse the `print` event _before_ an incoming `stx_transfer` event.

yknl:
How much work would it be to add these txs to an endpoint that is similar to the current `/v1/block` endpoint? I know that some integrators are using this right now for transaction confirmations.

zone117x:
@yknl you do mean generating fake stx-transfer txs for all the transfers occur within a `send-many-memo` contract-call tx, and sticking them inside a block payload? That's essentially the only way to make it so integrators can do _nothing_ except change a URL. It would be hacky and inaccurate, and potentially lead to a handful of other issues that would likely make it not _actually_ plug-and-play for integrators using the existing /v1/block polling approach.

yknl:
Yes, can you explain how it would be inaccurate?

zone117x:
The transactions within the block would no longer be accurate, as in, they would not be what the real blockchain's block contains. We would have to generate fake stx-transfer _transactions_ for every stx-transfer _event_ that is emitted from the single contract-call tx. So you have something like a couple `send-many-memo` contract-call transactions, artificially exploded into hundreds of stx-transfer transactions, with fake/inaccurate txids, tx indexs, etc.

yknl:
> The transactions within the block would no longer be accurate, as in, they would not be what the real blockchain's block contains. We would have to generate fake stx-transfer _transactions_ for every stx-transfer _event_ that is emitted from the single contract-call tx. So you have something like a couple `send-many-memo` contract-call transactions, artificially exploded into hundreds of stx-transfer transactions, with fake/inaccurate txids, tx indexs, etc.
> 
> Tooling could easily then try to lookup one of these fake transactions using another endpoint, or try to match them with data from another endpoint or the core-node itself, and break completely because of the inaccuracies this would introduce.

Thanks, that makes sense.

yknl:
This is not strictly related to this PR but the explorer and wallets not showing incoming transactions from the `send-many` contract is a *very* jarring experience for users. It leaves average users no way of verifying they're received an anticipated transaction without performing some calculations and no way for the sender to prove the transaction was sent.

zone117x:
You're referring to mempool right? That's the second item on the TODO list for this PR (or a follow up PR). First one already done.

yknl:
I was referring to transaction history in general. Is that no longer the case?

zone117x:
You mean transaction history per address right? That was addressed here https://github.com/blockstack/stacks-blockchain-api/pull/441#issuecomment-778243488

yknl:
> You mean transaction history per address right? That was addressed here [#441 (comment)](https://github.com/blockstack/stacks-blockchain-api/pull/441#issuecomment-778243488)

Are there any explorer changes needed to get the transactions to show or will this fix it?

zone117x:
The transaction show, here's an example https://explorer.stacks.co/txid/0x96ba5783bda05cd093b9028d2473db2875f886036d608ebc0a78b1026e41b76f?chain=testnet

yknl:
Feedback from integrators is that for this to resolve their issues, they need a way to filter by block height or block hash. 

For example:
`extended/v1/address/<address>/transactions?blockHash=<blockhash>`

or 

`extended/v1/address/<address>/transactions?height=3733`

friedger:
This issue also applies to other native assets. 

At Boom, we don't see incoming nft transfers from contract calls neither. //cc @dantrevino

zone117x:
@yknl Commit https://github.com/blockstack/stacks-blockchain-api/pull/441/commits/89b8750962d2b338a2f435125dc7f204d9905dc1 adds an optional block `height` filter param, e.g. `/extended/v1/address/<address>/transactions?height=3733`

kyranjamie (approved):

PR #444, opened 2021-02-15
Include any tx type in the address filtered mempool result

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/438

Previously, when using an address filter for the `/extended/v1/tx/mempool` endpoint, only `stx-transfer` tx types would be returned. This decision was initially made because only `stx-transfer` types have anything like a known `recipient` before the tx is mined. This is in contrast to a `contract-call` tx type, where there _can_ be recipients in the events generated, but events are not available until the tx is actually mined (and the contract code is ran).

This change allows any tx type to be returned, but keep in mind that only the tx `sender` will be checked for all tx types that are not a `stx-transfer`.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -129,19 +129,19 @@ paths:
       parameters:
         - name: sender_address
           in: query
-          description: Filter to only STX transfer transactions with this sender address.
+          description: Filter to only return transactions with this sender address.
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):
10/10 Would review again

PR #450, opened 2021-02-16
feat: add dotenv-flow

hstove:
Adds [`dotenv-flow`](https://github.com/kerimdzhanov/dotenv-flow). Finally got fed up enough with editing `.env`. This lets you use a `.env.local` which will override anything in `.env`.


zone117x (approved):
Handy lib, thanks @hstove!

PR #462, opened 2021-02-23
Add get block by height endpoint

zone117x:
Implement a new endpoint `get block by height`, similar to the `get block by hash` endpoint. 

`GET /extended/v1/block/by_height/<number>`

### DIFF
--- a/src/api/routes/block.ts
+++ b/src/api/routes/block.ts
@@ -40,7 +40,7 @@ export function createBlockRouter(db: DataStore): RouterWithAsync {
   });
 
   router.getAsync('/by_height/:height', async (req, res) => {
-    const height = parseInt(req.params['height']);
+    const height = parseInt(req.params['height'], 10);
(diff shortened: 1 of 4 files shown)


src/api/routes/block.ts
```diff
@@ -36,14 +39,33 @@ export function createBlockRouter(db: DataStore): RouterWithAsync {
     res.json(response);
   });
 
+  router.getAsync('/by_height/:height', async (req, res) => {
+    const height = parseInt(req.params['height']);
```
kyranjamie, line 43: need to handle the radix?

kyranjamie (approved):

src/api/routes/block.ts
zone117x: I think it's okay as of a new-ish version of nodejs/ecmascript where the octal heuristic parsing was removed. But still, js so is sketch.. fixed

PR #466, opened 2021-02-26
Total STX supply endpoint

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/455

Implements a total STX supply endpoint, and another that is formatted the same as the legacy 1.0 API -- so integrators would only need to change the URL.

Examples:

`GET /extended/v1/total_supply`:
```json
{
  "unlocked_percent": "77.48",
  "total_stx": "1352464600.000000",
  "unlocked_stx": "1047848171.287772",
  "block_height": 5086
}
```

`GET /extended/v1/total_supply/legacy_format`:
```json
{
  "unlockedPercent": "77.48",
  "totalStacks": "1352464600.000000",
  "totalStacksFormatted": "1,352,464,600.000000",
  "unlockedSupply": "1047848171.287772",
  "unlockedSupplyFormatted": "1,047,848,171.287772",
  "blockHeight": "5086
}
```

### DIFF
--- a/src/api/routes/total-supply.ts
+++ b/src/api/routes/total-supply.ts
@@ -14,6 +14,7 @@ import {
 import {
   NetworkBlockTimesResponse,
   NetworkBlockTimeResponse,
+  GetTotalStxSupplyResponse,
 } from '@blockstack/stacks-blockchain-api-types';
 
 export function createTotalSupplyRouter(db: DataStore): RouterWithAsync {
@@ -59,7 +60,7 @@ export function createTotalSupplyRouter(db: DataStore): RouterWithAsync {
       }
     }
     const supply = await getStxSupplyInfo(atBlockHeight);
-    const result = {
+    const result: GetTotalStxSupplyResponse = {
       unlocked_percent: supply.unlockedPercent,
       total_stx: supply.totalStx,
       unlocked_stx: supply.unlockedStx,
--- a/src/api/routes/total-supply.ts
+++ b/src/api/routes/total-supply.ts
@@ -2,18 +2,9 @@ import * as express from 'express';
 import { addAsync, RouterWithAsync } from '@awaitjs/express';
 import BigNumber from 'bignumber.js';
 import { DataStore } from '../../datastore/common';
-import { validate } from '../validate';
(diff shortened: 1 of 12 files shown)


zone117x:
Note that I'm seeing a liquid STX discrepancy of ~11 when comparing the sql query result to the `total_liquid_supply_ustx` value at https://stacks-node-api.mainnet.stacks.co/v2/pox.

The v2/pox endpoint shows `1047801298.354442` while the sql query shows `1047801309.687772`. An ~11 STX difference is a really strange amount -- the kinds of bugs that I'm thinking could cause a discrepancy would result in a much higher amount.

@kantai any ideas what this could be?

src/api/routes/total-supply.ts
```diff
@@ -0,0 +1,92 @@
...
+      unlocked_stx: supply.unlockedStx,
+      block_height: supply.blockHeight,
+    };
+    res.json(result);
+  });
+
+  router.getAsync('/legacy_format', async (req, res) => {
+    let atBlockHeight: number | undefined;
+    if ('height' in req.query) {
+      atBlockHeight = parseInt(req.query['height'] as string, 10);
```
kyranjamie, line 66: If using BigNumber I'd avoid parseInt altogether

kyranjamie: `new BigNumber(parseInt(req.query['height']).toNumber()`?

kyranjamie (approved):

src/api/routes/total-supply.ts
zone117x: `parseInt` pretty much never throws, which allows a nice 400 error message after this. I'm not sure about BigNumber

src/helpers.ts
```diff
@@ -154,13 +155,25 @@ export function formatMapToObject<TKey extends string, TValue, TFormatted>(
   return obj;
 }
 
+export const TOTAL_STACKS = new BigNumber(1320000000)
+  .plus(322146 * 100 + 5 * 50000) // air drop
```
zone117x, line 159: This was copied from the legacy 1.0 API codebase. Is it still accurate?

diwakergupta:
@zone117x OK to ship this but let's open another issue to understand the discrepancy so we don't lose track.

PR #471, opened 2021-03-03
Feat/dropped mempool txs

zone117x:
Implements the dropped mempool tx event handling from https://github.com/blockstack/stacks-blockchain/pull/2479

Likely fixes https://github.com/blockstack/stacks-blockchain-api/issues/458

Transactions that previously were stuck at `tx_status: pending` or (in rare cases) `canonical: false`, can now return more accurate `dropped` statuses:

```ts
interface MempoolTxResponse {
  /** Replaced by a transaction with the same nonce, but a higher fee. */
  tx_status: 'dropped_replace_by_fee';
  /** Replaced by a transaction with the same nonce but in the canonical fork. */
  tx_status: 'dropped_replace_across_fork';
  /** The transaction is too expensive to include in a block. */
  tx_status: 'dropped_too_expensive';
  /** Transaction was dropped because it became stale. */
  tx_status: 'dropped_stale_garbage_collect';
}
```

Note that dropped mempool transactions will not be included in the `/extended/v1/tx/mempool` endpoint. A new endpoint is provided to retrieve dropped transactions -- the primary use-case for this endpoint is currently for diagnostic purpose:

#### `GET /extended/v1/tx/mempool/dropped`

```json
{
  "limit": 96,
  "offset": 0,
  "total": 5,
  "results": [
    {
      "tx_id": "0x8912000000000000000000000000000000000000000000000000000000000005",
      "tx_status": "dropped_stale_garbage_collect",
      "tx_type": "coinbase",
      "receipt_time": 1594307705,
      "receipt_time_iso": "2020-07-09T15:15:05.000Z",
      "nonce": 0,
      "fee_rate": "1234",
      "sender_address": "sender-addr",
      "sponsored": true,
      "sponsor_address": "sponsor-addr",
      "post_condition_mode": "allow",
      "coinbase_payload": {
        "data": "0x636f696e62617365206869"
      }
    },
    {
      "tx_id": "0x8912000000000000000000000000000000000000000000000000000000000004",
      "tx_status": "dropped_too_expensive",
      "tx_type": "coinbase",
      "receipt_time": 1594307704,
      "receipt_time_iso": "2020-07-09T15:15:04.000Z",
      "nonce": 0,
      "fee_rate": "1234",
      "sender_address": "sender-addr",
      "sponsored": true,
      "sponsor_address": "sponsor-addr",
      "post_condition_mode": "allow",
      "coinbase_payload": {
        "data": "0x636f696e62617365206869"
      }
    },
    {
      "tx_id": "0x8912000000000000000000000000000000000000000000000000000000000003",
      "tx_status": "dropped_replace_by_fee",
      "tx_type": "coinbase",
      "receipt_time": 1594307703,
      "receipt_time_iso": "2020-07-09T15:15:03.000Z",
      "nonce": 0,
      "fee_rate": "1234",
      "sender_address": "sender-addr",
      "sponsored": true,
      "sponsor_address": "sponsor-addr",
      "post_condition_mode": "allow",
      "coinbase_payload": {
        "data": "0x636f696e62617365206869"
      }
    },
    {
      "tx_id": "0x8912000000000000000000000000000000000000000000000000000000000001",
      "tx_status": "dropped_replace_across_fork",
      "tx_type": "coinbase",
      "receipt_time": 1594307702,
      "receipt_time_iso": "2020-07-09T15:15:02.000Z",
      "nonce": 0,
      "fee_rate": "1234",
      "sender_address": "sender-addr",
      "sponsored": true,
      "sponsor_address": "sponsor-addr",
      "post_condition_mode": "allow",
      "coinbase_payload": {
        "data": "0x636f696e62617365206869"
      }
    },
    {
      "tx_id": "0x8912000000000000000000000000000000000000000000000000000000000000",
      "tx_status": "dropped_replace_across_fork",
      "tx_type": "coinbase",
      "receipt_time": 1594307695,
      "receipt_time_iso": "2020-07-09T15:14:55.000Z",
      "nonce": 0,
      "fee_rate": "1234",
      "sender_address": "sender-addr",
      "sponsored": true,
      "sponsor_address": "sponsor-addr",
      "post_condition_mode": "allow",
      "coinbase_payload": {
        "data": "0x636f696e62617365206869"
      }
    }
  ]
}
```

### DIFF
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -46,6 +46,7 @@ import {
   ElementType,
   FoundOrNot,
   hexToBuffer,
+  logger,
   unixEpochToIso,
 } from '../../helpers';
 import { readClarityValueArray, readTransactionPostConditions } from '../../p2p/tx';
@@ -478,18 +479,6 @@ export interface GetTxWithEventsArgs extends GetTxArgs {
   eventOffset: number;
 }
 
-function isDroppedTx(tx: DbMempoolTx): boolean {
-  switch (tx.status) {
-    case DbTxStatus.DroppedReplaceByFee:
-    case DbTxStatus.DroppedReplaceAcrossFork:
-    case DbTxStatus.DroppedTooExpensive:
-    case DbTxStatus.DroppedStaleGarbageCollect:
-      return true;
-    default:
-      return false;
-  }
-}
-
 export async function getTxFromDataStore(
   db: DataStore,
   args: GetTxArgs | GetTxWithEventsArgs
@@ -500,24 +489,27 @@ export async function getTxFromDataStore(
   const txQuery = await db.getTx(args.txId);
   const mempoolTxQuery = await db.getMempoolTx({ txId: args.txId, includePruned: true });
 
-  // If mined.canonical = false && mempool.dropped = true then use mempool
-  if (
-    txQuery.found &&
-    mempoolTxQuery.found &&
-    !txQuery.result.canonical &&
-    isDroppedTx(mempoolTxQuery.result)
-  ) {
-    dbTx = mempoolTxQuery.result;
-  } else if (txQuery.found) {
+  // First, check the happy path: the tx is mined and in the canonical chain.
+  if (txQuery.found && txQuery.result.canonical) {
     dbTx = txQuery.result;
-  } else if (mempoolTxQuery.found) {
+  }
+  // Otherwise, if not mined or not canonical, check in the mempool.
+  else if (mempoolTxQuery.found) {
     dbTx = mempoolTxQuery.result;
-  } else {
+  }
+  // Fallback for a situation where the tx was only mined in a non-canonical chain, but somehow not in the mempool table.
+  else if (txQuery.found) {
+    logger.warn(`Tx only exists in a non-canonical chain, missing from mempool: ${args.txId}`);
+    dbTx = txQuery.result;
+  }
+  // Tx not found in db.
+  else {
     return { found: false };
   }
 
   // if tx is included in a block
   if ('tx_index' in dbTx) {
+    // if tx events are requested
     if ('eventLimit' in args) {
       const eventsQuery = await db.getTxEvents({
         txId: args.txId,
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -24,6 +24,8 @@ import {
   RosettaBlock,
   RosettaParentBlockIdentifier,
   TransactionEventStxLock,
+  TransactionStatus,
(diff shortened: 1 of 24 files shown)


src/api/controllers/db-controller.ts
```diff
@@ -468,34 +478,58 @@ export interface GetTxWithEventsArgs extends GetTxArgs {
   eventOffset: number;
 }
 
+function isDroppedTx(tx: DbMempoolTx): boolean {
```
kyranjamie, line 481: Can tell the compiler the result of this is a specific type of DbMempoolTx, with `val is XXX` return syntax

kyranjamie (approved):

src/api/controllers/db-controller.ts
```diff
@@ -468,34 +478,58 @@ export interface GetTxWithEventsArgs extends GetTxArgs {
...
+    !txQuery.result.canonical &&
+    isDroppedTx(mempoolTxQuery.result)
+  ) {
     dbTx = mempoolTxQuery.result;
-  } else {
-    const txQuery = await db.getTx(args.txId);
-    if (!txQuery.found) {
-      return { found: false };
-    }
+  } else if (txQuery.found) {
```
kantai, line 511: I think you want something like:

```
if ( txQuery.found && txQuery.result.canonical ) {
      dbTx = txQuery.result;
} else {
      dbTx = mempoolTxQuery.result;
}
```

Where you use the "mempool" tx if the tx hasn't been mined in a canonical block. However, this brings up a question-- what happens when a tx is remined in a fork? What would `getTx` return in that case?

zone117x: I refactored and simplified the conditions here. The only bit of weirdness is when a tx only exists as non-canonical, and missing from the mempool table completely. I'm not sure if a mempool event is guaranteed for every tx, for example when syncing from scratch. In this case, it just returns the mined tx but with the `canonical: false`. The explorer should already have code to handle this.

 I added a log warning when this happens, although it seems like this may be a rare but expected condition?

kantai: > I'm not sure if a mempool event is guaranteed for every tx, for example when syncing from scratch. In this case, it just returns the mined tx but with the canonical: false flag. 

Yep, that's exactly right.

> I added a log warning when this happens, although, it seems like this may be an expected (albeit rare) condition?

Yep

zone117x: Okay 👍, in that case, do you think this PR is ready to merge?

kantai (approved):

src/api/controllers/db-controller.ts
kantai: Yep, LGTM

PR #472, opened 2021-03-04
fix: update rosetta api with latest specs

rllola:
## Description

Update Rosetta endpoints to the latest specs 1.4.10 (https://www.rosetta-api.org/docs/welcome.html). The changes include minor fixes on the `construction` and `block` endpoints.

It also include an update in the tests regarding the RSV/VRS see PR https://github.com/blockstack/stacks-blockchain-api/pull/463

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No.

No.

## Testing information
Tests should be done using rosetta-cli.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


src/rosetta-helpers.ts
```diff
@@ -282,6 +283,15 @@ export function isDecimalsSupported(operations: RosettaOperation[]): boolean {
   return true;
 }
 
+export function getCurrencyData(): RosettaCurrency {
```
kyranjamie, line 286: Maybe this can be renamed to `getStxCurrencyMetadata()`, or something a little more specific. Reading the fn name elsewhere in the code I wasn't entirely sure what it returns.

.env
```diff
@@ -9,6 +9,7 @@ STACKS_CORE_EVENT_HOST=127.0.0.1
 
 STACKS_BLOCKCHAIN_API_PORT=3999
 STACKS_BLOCKCHAIN_API_HOST=127.0.0.1
+STACKS_BLOCKCHAIN_API_DB=pg
```
kyranjamie, line 12: Do we want to commit this?

kyranjamie (commented):
Thanks & LGTM, though @zone117x should give final sign off.

.env
rllola: Well ideally we should not commit `.env` file at all. See the recommendation from the npm package used https://github.com/motdotla/dotenv#should-i-commit-my-env-file

My proposition is to have `example.env` file that could be used as a template and just let it loudly fail if you don't have setup your environment variables.

What do you think ?

rllola: But yeah it was a mistake it should not be there. I just modified it to fit my need.

zone117x: I actually don't agree with the `dotenv` recommendations in regards to this repo. The `.env` file serves as a single source of truth reference for various env vars and the default values. The alternative is to essentially duplicate the contents of this file as JS code within some more obscure init area, then also duplicate the contents of this file in another readme doc somewhere -- introducing yet another area to maintain, that will inevitably end up with bugs or outdated info. 

The `example.env` approach could work, but then we'd have to add documentation for `rename example.env to .env`, which pretty much everyone would be required to do, and kinda defeats the purpose.

rllola: The `.env` file define environment-specific values and there is passwords (postgres and btc rpc). I think that adding to the documentation an instruction about properly setting your environment variables is a good thing.

rllola: I have removed it.

zone117x:
Thanks @rllola! Can you ensure the `test:rosetta` script passes? You can see the test logs in gh actions here: https://github.com/blockstack/stacks-blockchain-api/pull/472/checks?check_run_id=2040509652

zone117x (approved):
LGTM 👍

PR #478, opened 2021-03-08
chore: fix gh pages workflow

CharlieC3:
## Description

Fixes github auth issue from [here](https://github.com/blockstack/stacks-blockchain-api/runs/2057412650?check_suite_focus=true#step:7:36) when pushing up gh-pages docs. I had to abandon using Gulp to deploy the docs because it's doing something weird and causing authentication problems where there shouldn't be any.

This solution generates the types and docs the same way it was being created previously, but uses a different method of publishing to the `gh-pages` branch.

Tested and working: https://github.com/blockstack/stacks-blockchain-api/tree/gh-pages

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Checklist

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,12 +1,10 @@
 name: gh-pages
-env:
-  COMMIT_USER: Hiro DevOps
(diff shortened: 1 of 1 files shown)


zone117x (approved):
LGTM, thanks again @CharlieC3

PR #481, opened 2021-03-09
chore: add stale config

agraebe:
Adding stale config [from template](https://github.com/blockstack/.github/blob/master/.github/stale.yml)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 180
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 180
```
diwakergupta, line 2: We should consider putting a tighter bound for stale PRs, maybe 3 months? You can specify PR specific settings in a `pulls` section, see https://github.com/probot/stale#usage for details

agraebe: updated!

zone117x (approved):

PR #488, opened 2021-03-11
fix: support `nft_burn_event` and `ft_burn_event`

zone117x:
Fix https://github.com/blockstack/stacks-blockchain-api/issues/487


kantai (approved):
LGTM!

PR #491, opened 2021-03-11
fix: correct post-condition code string

zone117x:
Fixes https://github.com/blockstack/stacks-blockchain-api/issues/486

Fixes the bug seen here:
image
image

Ge => greater than or equal
Le => less than or equal

/ht @kantai, @aulneau, @jasperjansz


aulneau:
Nice, thanks for the quick fix!

kantai (approved):
LGTM!

PR #494, opened 2021-03-12
List of nfts  by address

asimm241:

## Description

This PR list all nfts owned by an address

For details refer to issue https://github.com/blockstack/stacks-blockchain-api/issues/227

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

## Testing information
npm run test:integration

## Checklist
@zone117x

### DIFF
--- a/docs/entities/nft-events/nft-event.schema.json
+++ b/docs/entities/nft-events/nft-event.schema.json
@@ -6,7 +6,10 @@
     "sender",
     "recipient",
     "asset_identifier",
-    "value"
+    "value",
+    "tx_id",
+    "block_height",
+    "value_repr"
   ],
   "properties": {
     "sender": {
@@ -19,6 +22,15 @@
       "type": "string"
     },
     "value": {
+      "type": "string"
+    },
+    "tx_id": {
+      "type": "string"
+    },
+    "block_height": {
+      "type": "number"
+    },
+    "value_repr": {
       "type": "string"
     }
   }
--- a/docs/entities/nft-events/nft-event.schema.json
+++ b/docs/entities/nft-events/nft-event.schema.json
@@ -1,0 +1,25 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "NftEvent",
+  "required": [
+    "sender",
+    "recipient",
+    "asset_identifier",
+    "value"
+  ],
+  "properties": {
+    "sender": {
+      "type": "string"
+    },
+    "recipient": {
+      "type": "string"
+    },
+    "asset_identifier": {
+      "type": "string"
+    },
+    "value": {
+      "type": "string"
+    }
+  }
+}
--- a/src/api/routes/address.ts
+++ b/src/api/routes/address.ts
@@ -230,5 +230,22 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
     }
   });
 
+  router.getAsync('/:stx_address/nf_id', async (req, res) => {
+    // get recent asset event associated with address
+    const stxAddress = req.params['stx_address'];
+    if (!isValidPrincipal(stxAddress)) {
+      return res.status(400).json({ error: `invalid STX address "${stxAddress}"` });
+    }
+
+    const limit = parseAssetsQueryLimit(req.query.limit ?? 20);
+    const offset = parsePagingQueryInput(req.query.offset ?? 0);
+    const response = await db.getAddressNFTEvent({
+      stxAddress,
+      limit,
+      offset,
+    });
+    res.json({ nft_events: response.results, total: response.total });
+  });
+
   return router;
 }
--- a/src/api/routes/address.ts
+++ b/src/api/routes/address.ts
@@ -3,7 +3,13 @@ import { addAsync, RouterWithAsync } from '@awaitjs/express';
 import * as Bluebird from 'bluebird';
 import { DataStore } from '../../datastore/common';
 import { parseLimitQuery, parsePagingQueryInput } from '../pagination';
-import { formatMapToObject, getSendManyContract, isValidPrincipal, logger } from '../../helpers';
+import {
+  bufferToHexPrefixString,
+  formatMapToObject,
+  getSendManyContract,
+  isValidPrincipal,
+  logger,
+} from '../../helpers';
 import { getTxFromDataStore, parseDbEvent } from '../controllers/db-controller';
 import {
   TransactionResults,
@@ -12,6 +18,7 @@ import {
   AddressStxBalanceResponse,
   AddressStxInboundListResponse,
   InboundStxTransfer,
+  AddressNftListResponse,
 } from '@blockstack/stacks-blockchain-api-types';
 import { ChainID } from '@stacks/transactions';
 
@@ -244,7 +251,17 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
       limit,
       offset,
     });
-    res.json({ nft_events: response.results, total: response.total });
+    const nft_events = response.results.map(row => ({
+      ...row,
+      value: bufferToHexPrefixString(row.value),
+    }));
+    const nftListResponse: AddressNftListResponse = {
+      nft_events: nft_events,
+      total: response.total,
+      limit: limit,
+      offset: offset,
+    };
+    res.json(nftListResponse);
   });
 
   return router;
--- a/src/api/routes/address.ts
+++ b/src/api/routes/address.ts
@@ -20,7 +20,7 @@ import {
   InboundStxTransfer,
   AddressNftListResponse,
 } from '@blockstack/stacks-blockchain-api-types';
-import { ChainID } from '@stacks/transactions';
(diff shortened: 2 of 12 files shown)


src/api/routes/address.ts
```diff
@@ -230,5 +230,22 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
     }
   });
 
+  router.getAsync('/:stx_address/nf_id', async (req, res) => {
```
hstove, line 233: Any thoughts on a better name for this route? It seems that `nft_events` might be more appropriate

src/tests/api-tests.ts
```diff
@@ -2627,6 +2627,101 @@ describe('api tests', () => {
...
+      tx_index: 1,
+      block_height: 1,
+      asset_identifier: 'some-asset',
+      value: Buffer.from([0]),
+      recipient: addr2,
+      sender: 'none',
+    };
+    await db.updateNftEvent(client, stxTx, nftEvent2);
+
+    const result1 = await supertest(api.server).get(`/extended/v1/address/${addr2}/nft_events`);
```
zone117x, line 2709: Can this test that `addr1` no longer owns the nft?

src/api/routes/address.ts
```diff
@@ -230,5 +230,22 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
     }
   });
 
+  router.getAsync('/:stx_address/nft_events', async (req, res) => {
```
zone117x, line 233: Can you update `docs/openapi.yaml` with this new endpoint?

src/api/routes/address.ts
```diff
@@ -230,5 +230,22 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
...
+    }
+
+    const limit = parseAssetsQueryLimit(req.query.limit ?? 20);
+    const offset = parsePagingQueryInput(req.query.offset ?? 0);
+    const response = await db.getAddressNFTEvent({
+      stxAddress,
+      limit,
+      offset,
+    });
+    res.json({ nft_events: response.results, total: response.total });
```
zone117x, line 247: Looks like this is returning a type containing `value: Buffer`? That needs to be encoded as a hex string.

src/api/routes/address.ts
asimm241: yeah, I was waiting for it to be finalized.

src/datastore/postgres-store.ts
```diff
@@ -2936,6 +2937,42 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
+    limit: number;
+    offset: number;
+  }): Promise<{ results: AddressNftEventIdentifier[]; total: number }> {
+    return this.query(async client => {
+      const result = await client.query<AddressNftEventIdentifier & { count: string }>(
+        `
+        WITH lastTransfers AS (
+          SELECT MAX(block_height) as max_block_height, asset_identifier, value
+          FROM nft_events
+          WHERE canonical = true AND (sender = $1 OR recipient = $1)
```
zone117x, line 2951: What is the purpose of matching against `sender` here?

src/tests/api-tests.ts
asimm241: I don't understand, how do I test that?

zone117x: This test simulates an nft sent to `addr1`, then transferred to `addr2`. Fetching `/extended/v1/address/${addr1}/nft_events` again here should show that `addr1` no longer has the nft.

src/datastore/postgres-store.ts
asimm241: to incorporate the `Burn` type of NFTs. shouldn't it?

asimm241: You're right It should only check against the recipient.

src/tests/api-tests.ts
asimm241: _I wasn't going for it. I was just checking different total and limit._
is it supposed to remove `addr1` as an owner when another event with `addr2` and the same `asset_identifier` inserted in the database? 
Currently, db maintain both entries.

zone117x: That's alright, ownership over time is a good thing to test for. If you increment the `block_height` in the second event, then only `addr2` should end up owning the nft.

src/datastore/postgres-store.ts
asimm241: does this check make sense for `Transfer` type NFTs?

src/tests/api-tests.ts
asimm241: I have updated the query to check for ownership.

src/api/routes/address.ts
```diff
@@ -230,5 +237,32 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
...
+
+    const limit = parseAssetsQueryLimit(req.query.limit ?? 20);
+    const offset = parsePagingQueryInput(req.query.offset ?? 0);
+    const response = await db.getAddressNFTEvent({
+      stxAddress,
+      limit,
+      offset,
+    });
+    const nft_events = response.results.map(row => ({
+      ...row,
```
zone117x, line 255: Nit: can this object be explicitly defined rather than the spread operator? If extra properties end up in that row object in the future, they can unintentionally end up being sent to the client.

src/datastore/postgres-store.ts
```diff
@@ -2936,6 +2937,48 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
+          GROUP BY asset_identifier, value
+        ),
+        lastNFTTransfers AS (
+          SELECT MAX(block_height) as max_block_height, asset_identifier, value
+          FROM lastAddressTransfers FULL JOIN nft_events USING (asset_identifier, value)
+          WHERE canonical = true AND asset_identifier = lastAddressTransfers.asset_identifier
+          GROUP BY asset_identifier, value
+        )
+        SELECT sender, recipient, asset_identifier, value, COUNT(*) OVER() AS count
+        FROM lastNFTTransfers FULL JOIN nft_events USING (asset_identifier, value)
```
zone117x, line 2961: Looks like we'll need a multicolumn Index on `(asset_identifier, value)` for this query. You can just add 
```ts
pgm.createIndex('nft_events', ['asset_identifier', 'value']);
```
to `src/migrations/1588261750265_nft_events.ts`

src/datastore/postgres-store.ts
```diff
@@ -2936,6 +2937,48 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
 
+  async getAddressNFTEvent(args: {
+    stxAddress: string;
+    limit: number;
+    offset: number;
+  }): Promise<{ results: AddressNftEventIdentifier[]; total: number }> {
+    return this.query(async client => {
+      const result = await client.query<AddressNftEventIdentifier & { count: string }>(
+        `
+        WITH lastAddressTransfers AS (
```
zone117x, line 2948: It looks like this query can return incorrect NFT ownership if an NFT's ownership is transferred more than once per block, for example in different transactions within a block or if a single transaction moved the NFT multiple times. 

To fix this the query needs to get the latest by `block_height DESC, tx_index DESC, event_index DESC` as opposed to only `block_height`. 

The multicolumn filtering can be done using a `DISTINCT ON(...) ... GROUP BY ...`, for example:

```sql
WITH address_transfers AS (
  SELECT asset_identifier, value, sender, recipient, block_height, tx_index, event_index
  FROM nft_events
  WHERE canonical = true AND recipient = $1
),
last_nft_transfers AS (
  SELECT DISTINCT ON(asset_identifier, value) asset_identifier, value
  FROM nft_events
  WHERE canonical = true
  ORDER BY asset_identifier, value, block_height DESC, tx_index DESC, event_index DESC
)
SELECT sender, recipient, asset_identifier, value, COUNT(*) OVER() AS count
FROM address_transfers INNER JOIN last_nft_transfers USING (asset_identifier, value)
ORDER BY block_height DESC, tx_index DESC, event_index DESC
LIMIT $2 OFFSET $3
```

docs/entities/nft-events/nft-event.schema.json
```diff
@@ -0,0 +1,25 @@
+{
```
zone117x, line 1: @hstove @friedger is this API payload adequate for the uses you have in mind? It could also return the transaction ID, block height, etc if needed.

hstove: It's not bad, and would be valuable to ship in current form. But it would be much better to include txid, block height, timestamp, etc.

What does `value` return for different types? The value could be any Clarity type. I'm not sure what data we have, but including things like `raw_value`, `value_repr`, etc could be good.

zone117x (approved):
This is great, thanks @asimm241!

docs/entities/nft-events/nft-event.schema.json
friedger: I was late to the party :-) I have sent a new PR #507 to streamline clarity value representations

Issue #428, opened 2021-02-03
Implement get raw tx endpoint


zone117x:
Raw txs are already stored but current inaccessible via the API.
An endpoint like `/extended/v1/tx/:txid/raw` should return the raw transaction data.

PR #495, opened 2021-03-12
Get Raw Transaction Endpoint Implementation

noumanahmads:
## Description

Get Raw transaction from `tx_id`.
New endpoint `/:tx_id/raw` added. 

For details refer to issue #428 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

Yes, this endpoint should be documented somewhere. 

## Testing information
npm run test

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review

### DIFF
--- a/docs/api/transaction/get-raw-transaction.example.json
+++ b/docs/api/transaction/get-raw-transaction.example.json
@@ -1,0 +1,3 @@
+{ 
+  "raw_tx": "test-raw-tx" 
+}
--- a/docs/api/transaction/get-raw-transaction.example.json
+++ b/docs/api/transaction/get-raw-transaction.example.json
@@ -1,3 +1,3 @@
 { 
-  "raw_tx": "test-raw-tx" 
+  "raw_tx": "0x1234" 
 }
--- a/docs/api/transaction/get-raw-transaction.schema.json
+++ b/docs/api/transaction/get-raw-transaction.schema.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "GET raw transaction",
+  "title": "GetRawTransactionResult",
+  "type": "object",
+  "required": ["raw_tx"],
+  "properties": {
+    "raw_tx": {
+      "type":"string",
+      "description": "raw tx details"
+    }
+  }
+}
--- a/docs/api/transaction/get-raw-transaction.schema.json
+++ b/docs/api/transaction/get-raw-transaction.schema.json
@@ -7,7 +7,7 @@
   "properties": {
     "raw_tx": {
       "type":"string",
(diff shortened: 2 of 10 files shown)


src/datastore/postgres-store.ts
```diff
@@ -306,6 +306,10 @@ interface TransferQueryResult {
   amount: string;
 }
 
+export interface RawTxQueryResult {
+  raw_tx: string;
```
zone117x, line 310: This is stored as a `bytea` column type, which means this return value should be `raw_tx: Buffer`?

src/datastore/postgres-store.ts
```diff
@@ -2936,6 +2940,28 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
+        ORDER BY canonical DESC, block_height DESC
+        LIMIT 1
+        `,
+        [hexToBuffer(txId)]
+      );
+      if (result.rowCount === 0) {
+        return { found: false } as const;
+      }
+      const queryResult: RawTxQueryResult = {
+        raw_tx: result.rows[0].raw_tx.toString(),
```
zone117x, line 2959: `bufferToHexPrefixString` should be used to convert the `raw_tx` Buffer to a hex string.

docs/api/transaction/get-raw-transaction.example.json
```diff
@@ -0,0 +1,3 @@
+{ 
+  "raw_tx": "test-raw-tx"
```
zone117x, line 2: `"raw_tx: "0x1234"` would be more inline with the other examples

docs/api/transaction/get-raw-transaction.schema.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "GET raw transaction",
+  "title": "GetRawTransactionResult",
+  "type": "object",
+  "required": ["raw_tx"],
+  "properties": {
+    "raw_tx": {
+      "type":"string",
+      "description": "raw tx details"
```
zone117x, line 10: A more accurate description would be something like "A hex encoded serialized transaction"

docs/api/transaction/get-raw-transaction.example.json
noumanahmads: done.

docs/api/transaction/get-raw-transaction.schema.json
noumanahmads: Updated description.

zone117x (approved):
LGTM. Thanks @noumanahmads!

PR #501, opened 2021-03-16
Ability to specify cache-control headers for v2 proxied paths

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/467

Cache-control headers can be set using the `.proxy-cache-control.json` file. 

The file contains a set of regex paths with cache-control header values.

Example that makes `/v2/info` cache for 300 seconds, and `/v2/pox` cache for 150 seconds:

```json
{
  "paths": {
    "^/v2/info(.*)": "public, max-age=300",
    "^/v2/pox(.*)": "public, max-age=150"
  }
}
```

When matching routes the first matching entry is used.

This file is also `watched` during runtime, so that changes can be made while the application is running, and they will be automatically picked up.


zone117x:
@CharlieC3 does this setup work well for devops?

CharlieC3 (approved):
This works for me. I like that we don't need to redeploy the API to make changes. Good thinking!

zone117x:
@kantai does this approach make sense to you, considering the stacks-node path configuration, and cloudflare caching behavior?

kantai (approved):
This approach looks great to me!

PR #502, opened 2021-03-16
Condense all v2 routes into a single prom metric bucket

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/403

Make all `v2/*` routes report as a single metric.

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -169,6 +169,22 @@ export async function startApiServer(datastore: DataStore, chainId: ChainID): Pr
 
   // Manual route definitions for the /v2/ proxied endpoints
   routes.push({
+    path: '/v2/pox',
+    regexp: /^\/v2\/pox(.*)/,
+  });
+  routes.push({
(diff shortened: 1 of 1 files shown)


CharlieC3:
See my comment [here](https://github.com/blockstack/stacks-blockchain-api/issues/403#issuecomment-800530857)

wileyj:
would it be possible to separate out a few of the endpoints that we could use for canaries?
paths like `/v2/info` and `/v2/pox` would be very helpful to have individual metrics for to measure latency - these directly affect things like the wallet.

zone117x:
I added those and the couple other common endpoints I've seen

wileyj:
> I added those and the couple other common endpoints I've seen

d'oh, not sure how i didn't see that - not like there's a ton of code to look over. thanks!

wileyj (approved):
shipit

CharlieC3 (approved):
Thanks for adding the other buckets! :shipit:

Issue #497, opened 2021-03-14
Improve write retry logic


CharlieC3:
**Describe the bug**
If the database is briefly unavailable (1-5 minutes), writes may cause a `ECONNREFUSED` or `ETIMEDOUT` error. Instead of retrying the same failed write, the API appears to just try writing the next events coming in from the follower.

**To Reproduce**
1. Run the API, follower, and postgres DB locally
2. Cut the connection to the postgres db either by restart or some other method
3. Have the follower emit an event while the postgres db is still down
4. postgres db comes back up
5. Have the follower emit another event

**Expected behavior**
The API should ideally place all pending writes in a FIFO queue and continuously retry writing the event next in line until it succeeds.

**Console log**
[Testnet - stacks-node-api.txt](https://github.com/blockstack/stacks-blockchain-api/files/6137340/Testnet.-.stacks-node-api.txt)

**Additional context**
API version: `0.52.3`

PR #503, opened 2021-03-17
fix: handle transient postgres connection errors #497

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/497

Transient pg connection errors are not handled by the postgres lib's connection pool. This is somewhat unexpected, see https://github.com/brianc/node-postgres/issues/1789.

This PR adds connection try/catch retry logic for the errors that occur when the postgres service is restarting.

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -332,10 +332,41 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
   }
 
   /**
+   * Creates a postgres pool client connection. If the connection fails due to a transient error, it is retried until successful.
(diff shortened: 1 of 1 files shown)


CharlieC3:
@zone117x Does this handle `ETIMEDOUT` PGDB errors as well? I saw some of those errors appear next to `ECONNREFUSED` errors in the log file I uploaded to the original issue.

zone117x:
I just added handling for `ETIMEDOUT`, although I wasn't able to reproduce this one locally, so we'll just have to assume that nodejs doesn't do anything odd/inconsistent between with this error code vs `ECONNREFUSED`

CharlieC3 (approved):
Thanks for the quick turnaround on this issue!

Issue #464, opened 2021-02-23
Paginate tx events in same format as other paginated routes


aulneau:
There is no total events returned with the events key in a transaction. It'd be nice if it was there to help know if there are more events to fetch for creating a paginated "load more" component.

noumanahmads:
I am working on this.

kyranjamie:
This will be a breaking change, right?

noumanahmads:
I am updating this [method](https://github.com/blockstack/stacks-blockchain-api/blob/98404b4f08aff542bfbe5c167a30aa65f34800ae/src/datastore/postgres-store.ts#L1698).
Here offset and limit is implemented in multiple queries like this :
```
 SELECT
          event_index, tx_id, tx_index, block_height, canonical, locked_amount, unlock_height, locked_address
        FROM stx_lock_events
        WHERE tx_id = $1 AND index_block_hash = $2 AND event_index BETWEEN $3 AND $4
```
```
SELECT
          event_index, tx_id, tx_index, block_height, canonical, asset_event_type_id, sender, recipient, amount
        FROM stx_events
        WHERE tx_id = $1 AND index_block_hash = $2 AND event_index BETWEEN $3 AND $4
```
```
SELECT
          event_index, tx_id, tx_index, block_height, canonical, asset_event_type_id, sender, recipient, asset_identifier, amount
        FROM ft_events
        WHERE tx_id = $1 AND index_block_hash = $2 AND event_index BETWEEN $3 AND $4
```
```
SELECT
          event_index, tx_id, tx_index, block_height, canonical, asset_event_type_id, sender, recipient, asset_identifier, value
        FROM nft_events
        WHERE tx_id = $1 AND index_block_hash = $2 AND event_index BETWEEN $3 AND $4
```
```
SELECT
          event_index, tx_id, tx_index, block_height, canonical, contract_identifier, topic, value
        FROM contract_logs
        WHERE tx_id = $1 AND index_block_hash = $2 AND event_index BETWEEN $3 AND $4
```

How event index is implemented in these tables? 
For example if `event_index` is 1 in table `stx_events` then it shouldn't exist in any other tables `ft_events`, `nft_events`, `contract_logs`. Is this the case here?

zone117x:
@kyranjamie this is just adding a `total` field to the response, no breaking change. 

[...]

PR #513, opened 2021-03-19
Fix/paginate tx events

noumanahmads:
## Description

Paginate events in the transaction.

For details refer to issue #464

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No. 

## Testing information
npm run test

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review

### DIFF
--- a/docs/entities/transactions/abstract-transaction.schema.json
+++ b/docs/entities/transactions/abstract-transaction.schema.json
@@ -89,11 +89,29 @@
     "post_condition_mode": {
       "$ref": "../post-conditions/post-condition-mode.schema.json"
     },
-    "events": {
-      "type": "array",
-      "description": "List of transaction events",
-      "items": {
-        "$ref": "../transaction-events/transaction-event.schema.json"
+    "events" : {
+      "type": "object",
+      "description": "Events of transaction",
+      "properties": {
+        "limit": {
+          "type": "integer",
+          "description": "The number of events to return"
+        },
+        "offset": {
+          "type": "integer",
+          "description": "The number to events to skip"
+        },
+        "total": {
+          "type": "integer",
+          "description": "The number of blocks available"
+        },
+        "results": {
+          "type": "array",
+          "items": {
+            "description": "List of transaction events",
+            "$ref": "../transaction-events/transaction-event.schema.json"
+          }
+        }
       }
     }
   }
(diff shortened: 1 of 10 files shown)


docs/entities/transactions/abstract-transaction.schema.json
```diff
@@ -89,11 +89,29 @@
     "post_condition_mode": {
       "$ref": "../post-conditions/post-condition-mode.schema.json"
     },
-    "events": {
-      "type": "array",
-      "description": "List of transaction events",
-      "items": {
-        "$ref": "../transaction-events/transaction-event.schema.json"
+    "events" : {
```
zone117x, line 92: This is a significant breaking change for anyone using this API response. I think a good middle-ground is to include `event_count` as a top-level property, and to leave this property as in (we wouldn't get the returned limit and offset values, but I don't think those are very important).

@kyranjamie @aulneau

kyranjamie: Yes, this is necessary otherwise all previous wallet builds will break.

docs/index.d.ts
```diff
@@ -1816,7 +1828,7 @@ export interface RosettaSignature {
  */
 export interface SigningPayload {
   /**
-   * The network-specific address of the account that should sign the payload.
+   * [DEPRECATED by account_identifier in v1.4.4] The network-specific address of the account that should sign the payload.
```
zone117x, line 1831: This seems like a rebase problem?

docs/index.d.ts
```diff
@@ -1254,7 +1254,19 @@ export interface NftEvent {
...
+  value: {
+    /**
+     * Hex string representing the identifier of the NFT
+     */
+    hex: string;
+    /**
+     * Readable string of the NFT identifier
+     */
+    repr: string;
+  };
```
zone117x, line 1269: Same here, rebase problem?

docs/index.d.ts
noumanahmads: Yes, my branch was rebased with the master. I did some schema updates and ran `npm run build`. These changes came from that. 
If I run `npm run build` in the master branch, it is also giving me these changes.

zone117x: Ah see, thanks for the clarification.

zone117x (approved):
LGTM!

PR #522, opened 2021-03-25
Graceful shutdown

zone117x:
Fixes a race condition during shutdown while an event is being processed. A sql transaction commit can be completed without the HTTP 200 response making it to the stacks-node, which would result in the stacks-node re-playing this same event once the API is back online. 

Also fixes a handful of other small issues:
* The HTTP server init was not handling the async errors like `EADDRINUSE` after calling `server.listen()`.
* The websocket server lib wrapping the http server object was swallowing (at least) the `error` event, it has been further de-coupled from the http server.
* Improved error message for `The configured STACKS_CHAIN_ID does not match`.
* Improved error stack traces by installing `source-map-support`.

The `npm run start` script has unexpected signals passing behavior (like `SIGINT`). The dockerfile was updated to use `node ./lib/index.js`. 
See https://github.com/npm/npm/issues/4603 for 6 years of this causing problems in projects 🤦


src/api/routes/ws-rpc.ts
```diff
@@ -78,23 +81,23 @@ export function createWsRpcRouter(db: DataStore, server: http.Server): WebSocket
       let rpcReqs = Array.isArray(parsedRpcReq) ? parsedRpcReq : [parsedRpcReq];
 
       // Ignore client notifications, spec dictates server should never respond to these.
-      rpcReqs = rpcReqs.filter(req => req.type !== RpcStatusType.notification);
+      rpcReqs = rpcReqs.filter(req => req.type !== 'notification');
 
       const responses: JsonRpc[] = rpcReqs.map(rpcReq => {
         switch (rpcReq.type) {
-          case RpcStatusType.request:
+          case 'request':
```
kyranjamie, line 88: Why killing the string enum?

kyranjamie (approved):

src/api/routes/ws-rpc.ts
zone117x: It was throwing with `RpcStatusType` being undefined. I'm not sure how it ever worked. Maybe ts-node doesn't compile the const enum when debugging now, not really sure.

PR #525, opened 2021-03-26
chore: add .nvmrc

marcosc90:
## Description

This PR adds `.nvmrc` for those who use `nvm` to quickly switch node versions between projects

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No changes to production code.

## Testing information

run `nvm use` and node version should switch to `14.16.0`


zone117x (commented):
I didn't know this existed! Super useful, thank you. 
Do you know how strict this is, and if it can be set to something like `v14.*` ?

marcosc90:
> Do you know how strict this is, and if it can be set to something like `v14.*` ?

We can set it to `14` and should use the highest `14.x` that you have installed via NVM.

zone117x:
Do you think we should use `v14` or `v14.16`?

marcosc90:
I prefer a specific version unless we know for sure that it works on all 14.x.x.

zone117x:
I think it would be safe to allow any patch version, unless nodejs breaks semver it should work?
Scoping it to a specific minor version makes the most sense to me. We should also update any dockerfiles like https://github.com/blockstack/stacks-blockchain-api/blob/b6fcf1fc6696d6ab97193018d0a619d0ca4e4a5b/Dockerfile#L1 to a more specific version as well.

marcosc90:
> I think it would be safe to allow any patch version, unless nodejs breaks semver it should work?

Agree, we can set `v14.16`, so any `v14.16.z` would work

zone117x:
Can you update with `v14.16` and then this looks good to merge 👍

zone117x (approved):
LGTM, thanks @marcosc90!

PR #528, opened 2021-03-29
fix: use `suggested_fee` info in rosetta

rllola:

## Description

We now use `suggested_fee` to calculate the fee required for the transaction. We don't need the `fee` operation neither in order to create a transaction.

See https://github.com/blockstack/stacks-blockchain-api/issues/505

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

Yes. The api will return an error if `fee` operation is added when creating an transaction.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/docs/entities/rosetta/rosetta-construction-options.schema.json
+++ b/docs/entities/rosetta/rosetta-construction-options.schema.json
@@ -23,7 +23,7 @@
     },
     "amount": {
       "type": "string",
-      "description": "Amount to be transfeered."
+      "description": "Amount to be transfered."
     },
     "symbol": {
       "type": "string",
@@ -52,6 +52,10 @@
     "fee": {
       "type": "string",
       "description": "Fee for this transaction"
+    },
+    "size": {
+      "type": "number",
+      "description": "Transaction approximative size (used to calculate total fee)."
     }
   }
 }
--- a/docs/entities/rosetta/rosetta-construction-options.schema.json
+++ b/docs/entities/rosetta/rosetta-construction-options.schema.json
@@ -54,7 +54,7 @@
       "description": "Fee for this transaction"
     },
     "size": {
-      "type": "number",
+      "type": "integer",
       "description": "Transaction approximative size (used to calculate total fee)."
     }
   }
--- a/src/api/routes/rosetta/construction.ts
+++ b/src/api/routes/rosetta/construction.ts
@@ -18,6 +18,8 @@ import {
   RosettaConstructionPayloadResponse,
   RosettaConstructionCombineRequest,
   RosettaConstructionCombineResponse,
+  RosettaAmount,
+  RosettaCurrency,
 } from '@blockstack/stacks-blockchain-api-types';
 import {
   createMessageSignature,
@@ -115,6 +117,7 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
     }
 
     const operations: RosettaOperation[] = req.body.operations;
+    const tokenTransferOptions: UnsignedTokenTransferOptions;
 
     // We are only supporting transfer, we should have operations length = 3
     if (operations.length != 3) {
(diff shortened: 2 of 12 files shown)


docs/entities/rosetta/rosetta-construction-options.schema.json
```diff
@@ -52,6 +52,10 @@
     "fee": {
       "type": "string",
       "description": "Fee for this transaction"
+    },
+    "size": {
+      "type": "number",
```
kyranjamie, line 57: Should this be `integer`?

src/api/routes/rosetta/construction.ts
```diff
@@ -165,6 +167,24 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
...
 
+    // dummy transaction to calculate size
+    const tokenTransferOptions: UnsignedTokenTransferOptions = {
+      recipient: options.token_transfer_recipient_address as string,
+      amount: new BN(options.amount as string),
+      // We don't know the fee yet but need a placeholder
+      fee: new BN(0),
+      // placeholder public key
+      publicKey: '000000000000000000000000000000000000000000000000000000000000000000',
+      network: GetStacksTestnetNetwork(),
```
kyranjamie, line 178: Seems unconventional to uppercase this method. Can it be renamed to `getStacksTestnetNetwork` like the rest?

src/api/routes/rosetta/construction.ts
```diff
@@ -248,12 +275,25 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
...
 
+    // Getting fee info
+    const feeInfo = await new StacksCoreRpcClient().getEstimatedTransferFee();
+    const currency: RosettaCurrency = {
+      symbol: RosettaConstants.symbol,
+      decimals: RosettaConstants.decimals,
+    };
+
+    const fee: RosettaAmount = {
+      value: (+feeInfo * txSize).toString(),
```
kyranjamie, line 286: Might this need to use a safe math lib?

docs/entities/rosetta/rosetta-construction-options.schema.json
rllola: Yep. Probably.

src/api/routes/rosetta/construction.ts
rllola: Yes should be ok to do that.

That make me think. Should we not use the correct network info ? We can detect using chain id info right ?

src/api/routes/rosetta/construction.ts
rllola: Yeah I am not sure about that. Typescript should avoid problems but not quite sure.

Will blockstack allow negative fee ? I can test for that. There is also a `max_fee` value that is used to verify we won't spend too much on fee.

Also another question, does transfer transaction have a fix size ?

docs/entities/rosetta/rosetta-construction-options.schema.json
rllola: Does it actually change index.d.ts ? I changed locally but I have the same file generated.

kyranjamie: It won't change the resulting type (TypeScript doesn't differentiate numbers and "integers"), but it would be more accurate were to use these specs for tests. `1.23` would fail with `integer` type

src/api/routes/rosetta/construction.ts
```diff
@@ -165,6 +167,24 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
       }
     }
 
+    // dummy transaction to calculate size
+    const tokenTransferOptions: UnsignedTokenTransferOptions = {
```
kyranjamie, line 171:
```suggestion
    // dummy transaction to calculate size
    const dummyTokenTransferTx: UnsignedTokenTransferOptions = {
```

Could rename this so it's clear it's not needed as a real tx

src/api/routes/rosetta/construction.ts
kyranjamie: I don't think it matters if this tx is only to calculate the size?

src/api/routes/rosetta/construction.ts
kyranjamie: TypeScript doesn't change anything at runtime, so this could be prone to floating point imprecision 0.1 * 0.2

Negative fee should definitely fail.

I believe transfers txs do have a fixed size, but only because the `@stacks/transaction` lib pads the memo field with empty bytes

zone117x: Yeah, we should just use native `BigInt` here.

src/api/routes/rosetta/construction.ts
zone117x: Yep, the chainID should be used to determine whether testnet or mainnet is used here 👍

rllola: Yeah this tx is only used to calculate the size but in `/payload` it also calls `GetStacksTestnetNetwork` regardless of the network defined. I am going to create a dedicated PR for this.

rllola:
I have made the requested changes expect for the testnet/mainnet network which I will do in another PR.

zone117x (approved):

PR #530, opened 2021-03-30
Fix/readme defaults

kyranjamie:
Adds info to readme about the Java runtime environment dependency.

Sets default chain id to testnet.


.env
```diff
@@ -22,7 +22,7 @@ STACKS_CORE_RPC_PORT=20443
 # STACKS_FAUCET_NODE_PORT=<port number>
 
 ## configure the chainID/networkID; testnet: 0x80000000, mainnet: 0x00000001
-STACKS_CHAIN_ID=0x00000001
+STACKS_CHAIN_ID=0x80000000
```
zone117x, line 25: I'm wondering if it might make more sense to set this env var in the relevant npm scripts? Otherwise, this change will fix the chain/network ID error when running in the dev/testnet environments, but introduce the error in mainnet environments. 

Alternatively, we can keep this change here, but ensure the `mainnet` variable is set in the dockerfiles?

kyranjamie: I added this as I figured it's a more sensible default, but strictly speaking, neither should really be defined. Can we remove it altogether?

zone117x: We could, as long as the dockerfiles specify them. But it would mean all current methods of starting the API (either in mainnet/testnet/mocknet), would immediately throw an error that this env var is missing. I think it makes more sense for the npm scripts that are most likely to be used to start in testnet/mocknet should ensure this is set.

agraebe:
@zone117x to follow up. this PR might be redundant.

readme.md
```diff
@@ -38,6 +35,10 @@ Check to see if the server started successfully by visiting http://localhost:399
 
 ## Local Development
 
+### Requirements
+
+An API client is generated automatically with [OpenAPI Generator](https://github.com/OpenAPITools/openapi-generator). This has a dependency on [Java RE](https://www.java.com/en/download/).
```
zone117x, line 40: In the develop branch, there root proj no longer a direct dependency on the client lib, so the JRE so no longer necessary unless specifically trying to re-build the client lib. Should we keep this change in?

.env
zone117x: I think changing this is too disruptive without also changing a bunch of related docker images and probably some docs. Can you revert it?

readme.md
```diff
@@ -14,20 +14,17 @@ Ensure Docker is installed, then run the command:
...
 docker run -p 3999:3999 blockstack/stacks-blockchain-api-standalone mocknet
 ```
 
-
 Once the blockchain has synced with network, the API will be available at:
 [http://localhost:3999](http://localhost:3999)
 
 ## Development quick start
 
-First, ensure Docker is installed on your machine.
```
zone117x, line 29: Docker still required :)

zone117x:
@kyranjamie closing this for now, I think the major issue(s) that originally prompted this PR are no longer a problem. Let me know if you feel otherwise!

PR #545, opened 2021-04-13
feat: implement offline mode

### LINKED ISSUE #544: [Rosetta] Offline Construction API
### Context:
Rosetta-Api requires Construction API to be able to be run in a [completely offline environment](https://www.rosetta-api.org/docs/construction_api_introduction.html#completely-offline). This is a hard requirement from Coinbase for security reasons. Production transaction testing cannot commence without it.

As far as I can tell, the Stacks-API implements stacks `/v2` routes and `/rosetta/v1` routes in the same nodejs server

### Requirements
- to be able to run Construction API only in an offline environment
- `/construction/metadata` endpoint is the exception
- ideally an environment variable to toggle "online" vs "offline" modes

### Examples:
See [cardano-rosetta](https://github.com/input-output-hk/cardano-rosetta#mode)
See [rosetta-bitcoin](https://github.com/coinbase/rosetta-bitcoin/blob/7b5393dca11490da30594c211dc3b716900e4ac0/configuration/configuration.go#L41-L43)

asimm241:
## Description

This PR implements an offline mode for `stacks-blockchain-api`. This offline mode required by Rosetta's construction endpoints. In this mode, there will be no requests to node and database.  Only those API's which can work offline will work and nothing else. 

Closes #544 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Testing information
npm run test:integration:rosetta

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review


src/index.ts
```diff
@@ -51,61 +57,66 @@ async function getCoreChainID(): Promise<ChainID> {
...
-    case 'pg':
-    case undefined: {
-      db = await PgDataStore.connect();
-      break;
-    }
-    default: {
-      throw new Error(
-        `Invalid STACKS_BLOCKCHAIN_API_DB option: "${process.env['STACKS_BLOCKCHAIN_API_DB']}"`
-      );
+  const mode = process.env['MODE'];
```
zone117x, line 60: The env var should have an application prefix, for example `STACKS_API_x`. Also since this is a boolean, it should probably be something like `STACKS_API_OFFLINE_MODE`, then change the check to:
`if ('STACKS_API_OFFLINE_MODE' in process.env) { // run in offline mode }`

src/datastore/offline-dummy-store.ts
```diff
@@ -0,0 +1,284 @@
+import { EventEmitter } from 'events';
```
zone117x, line 1: This is a lot of code that requires maintaining as the `DataStore` interface evolves. Something like `db = {} as DataStore;` would have the same effect.

Alternatively, a js `proxy` could be used to throw a helpful error if a function is accidentally called. Something like:
```ts
export const OfflineDummyStore: DataStore = new Proxy(new EventEmitter() as DataStore, {
  get(target: any, propKey) {
    if (propKey in target) return target[propKey];
    return function () {
      throw new Error(
        `Cannot call function on the Dummy datastore. Check if the application is running in offline mode.`
      );
    };
  },
});
```

zone117x (commented):
Can you add the a commented out `STACKS_API_OFFLINE_MODE` to the `.env` file, to serve as a reference? Like
```
# Used to run the API without a stacks-node or postgres connection. Only the Rosetta <???> endpoints are supported in this mode
# STACKS_API_OFFLINE_MODE=1
```

We should probably also mention it in the readme for deploying. The readme could use a lot more usage documentation in general, but adding this would be a good start.

.env
```diff
@@ -36,3 +36,15 @@ MAINNET_SEND_MANY_CONTRACT_ID=SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.send-man
 
 # Directory containing Stacks 1.0 BNS data extracted from https://storage.googleapis.com/blockstack-v1-migration-data/export-data.tar.gz
 # BNS_IMPORT_DIR=/extracted/export-data-dir/
+
+# Used to run the API without a stacks-node or postgres connection. Only the Rosetta 
+# /network/list
```
zone117x, line 41: Do you think it might be more helpful for this (and the readme) to include a link to https://www.rosetta-api.org/docs/node_deployment.html#offline-mode-endpoints rather than listing the endpoints?

asimm241: yeah, that is more elegant.  :)

zone117x (approved):

PR #546, opened 2021-04-13
chore: stacking feature; event processing;

rllola:

## Description

* Rosetta ocnstruction api now support building stacking operation
* Fix the missing events that change the balance for Data api in rosetta side
* Get Network based on chainID (by default it used to be testnet)

See old PR https://github.com/blockstack/stacks-blockchain-api/pull/543

## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Other

No

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -1,44 +1,43 @@
 import {
-  serializeCV,
-  deserializeCV,
-  ClarityAbi,
   abiFunctionToString,
-  getTypeString,
-  cvToString,
   BufferReader,
+  ClarityAbi,
+  cvToString,
+  deserializeCV,
+  getTypeString,
+  serializeCV,
 } from '@stacks/transactions';
 
 import {
-  Transaction,
-  SmartContractTransaction,
-  ContractCallTransaction,
-  TransactionEvent,
   Block,
-  TransactionType,
-  TransactionEventSmartContractLog,
-  TransactionEventStxAsset,
-  TransactionEventFungibleAsset,
-  TransactionEventNonFungibleAsset,
+  ContractCallTransaction,
   MempoolTransaction,
-  RosettaTransaction,
+  MempoolTransactionStatus,
   RosettaBlock,
   RosettaParentBlockIdentifier,
+  RosettaTransaction,
+  SmartContractTransaction,
+  Transaction,
+  TransactionEvent,
+  TransactionEventFungibleAsset,
+  TransactionEventNonFungibleAsset,
+  TransactionEventSmartContractLog,
+  TransactionEventStxAsset,
   TransactionEventStxLock,
   TransactionStatus,
-  MempoolTransactionStatus,
+  TransactionType,
 } from '@blockstack/stacks-blockchain-api-types';
 
 import {
   DataStore,
+  DbAssetEventTypeId,
+  DbBlock,
+  DbEvent,
+  DbEventTypeId,
+  DbMempoolTx,
+  DbTx,
   DbTxStatus,
   DbTxTypeId,
-  DbEventTypeId,
-  DbAssetEventTypeId,
-  DbStxEvent,
-  DbEvent,
-  DbTx,
-  DbMempoolTx,
-  DbBlock,
 } from '../../datastore/common';
 import {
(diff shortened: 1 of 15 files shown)


asimm241 (approved):
LGTM!

asimm241 (changes requested):
Please fix the failing checks though.

rllola:
@asimm241 Done!

src/api/controllers/db-controller.ts
```diff
@@ -51,7 +50,9 @@ import {
 } from '../../helpers';
 import { readClarityValueArray, readTransactionPostConditions } from '../../p2p/tx';
 import { serializePostCondition, serializePostConditionMode } from '../serializers/post-conditions';
-import { getOperations } from '../../rosetta-helpers';
+import { getMinerOperations, getOperations, processEvents } from '../../rosetta-helpers';
+
+const LIMIT_EVENTS_QUERY = 500000;
```
zone117x, line 55: Can this be moved into the function where it's used?

src/api/routes/rosetta/construction.ts
```diff
@@ -226,47 +263,69 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
...
+        }
+        break;
+      case 'stacking':
+        // Getting stacking info
+        const poxInfo = await new StacksCoreRpcClient().getPox();
+        const coreInfo = await new StacksCoreRpcClient().getInfo();
+        const contractInfo = poxInfo.contract_id.split('.');
+        options.contract_address = contractInfo[0];
+        options.contract_name = contractInfo[1];
+        // Adding 3 blocks to provide a buffer for transaction to confirm
```
zone117x, line 287: Can you explain a bit on how this number was chosen?

src/api/controllers/db-controller.ts
```diff
@@ -345,20 +359,41 @@ export async function getBlockFromDataStore({
...
-    const operations = getOperations(tx);
-    return {
+
+  const transactions: RosettaTransaction[] = [];
+
+  for (const tx of txsQuery.result) {
+    const events = await db.getTxEvents({
+      txId: tx.tx_id,
+      indexBlockHash: indexBlockHash,
+      limit: LIMIT_EVENTS_QUERY,
```
zone117x, line 385: Has this been tested with the genesis block (height 1)? There's 300k+ stx-events in the genesis block. The postgres query to fetch all of these (`SELECT * from stx_events where block_height = 1`) takes ~10 seconds in our infrastructure, and my guess is that the rosetta endpoint response size would be several megabytes.

raynaudoe: We did the rosetta-cli test on the entire mainnet network, but starting from index 2 because as you say the response of block 1 was quite heavy due to all the mint events.

zone117x: Should this function have special handling for block 1 to ignore events? This would be a dos vulnerability on the public API.

raynaudoe: I can surely skip searching for events on block 1, also adding a reasonable limit for the query ?. Would `limit: 5000` be a good upper limit for the events amount in one block ?

src/api/routes/rosetta/construction.ts
rllola: Sure. I have found it in you documentation step-by-step guide : https://docs.blockstack.org/understand-stacks/integrate-stacking#step-5-lock-stx-to-stack

Could it be an issue ?

src/api/controllers/db-controller.ts
zone117x: Yep that sounds okay. The main thing we need is to skip the event querying from the genesis block.

raynaudoe: ok, done!

src/api/routes/rosetta/construction.ts
zone117x: I don't have the context on how that number was originally decided, but I'm assuming there is a good reason for it 👍

zone117x (approved):
This looks great, thanks everyone!

PR #547, opened 2021-04-13
feat: address transactions with stx transfers endpoint

### LINKED ISSUE #369: Include STX spend from all transaction types
Related https://github.com/blockstack/stacks-blockchain-api/pull/367

In the wallet a list of transactions for a given address is displayed. The UI annotates the amount of STX inbound, or outbound, for each transaction.

In the case of a `contract_call` transactions, the list of events must be iterated in order to sum the total (of `stx_asset` `transfer` types).

Ideally, it'd be great if the `ContractCallTransaction` could include an additional property that calculates this, so the client doesn't have to make many additional requests.

zone117x:
Closes #369

Implements a new endpoint `/extended/v1/address/{principal}/transactions_with_transfers`. This is similar to the `/extended/v1/address/{principal}/transactions` endpoint, but adds STX transfer data to each transaction entry.

This allows the explorer and wallets to easily show STX transfer amounts for tx types other than simple stx-transfer txs. For example:
* The genesis block transaction for addresses in the Stacks 1.0 import.
* Coinbase transactions where STX are minted for addresses with an unlocking schedule.
* Contract-call transactions, for example the `send-many` contract used by exchanges.

Here's the interface for the items returned in the response:
https://github.com/blockstack/stacks-blockchain-api/blob/e8becd488bddc99da32c6306a5320c4772d16645/docs/index.d.ts#L1085-L1112

And here's some sample example for `/extended/v1/address/SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE/transactions_with_transfers/`:
https://github.com/blockstack/stacks-blockchain-api/blob/e8becd488bddc99da32c6306a5320c4772d16645/docs/api/address/get-address-transactions-with-transfers.example.json#L1-L107

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -53,6 +53,7 @@ import {
   DbBnsSubdomain,
   DbConfigState,
   DbTokenOfferingLocked,
+  DbTxWithStxTransfers,
 } from './common';
 import {
   TokenOfferingLocked,
@@ -3028,7 +3029,7 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
     height?: number;
   }): Promise<{ results: DbTx[]; total: number }> {
     return this.query(async client => {
-      const queryParams: any[] = [stxAddress, limit, offset];
+      const queryParams: (string | number)[] = [stxAddress, limit, offset];
       if (height !== undefined) {
         queryParams.push(height);
       }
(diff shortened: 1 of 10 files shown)


src/datastore/postgres-store.ts
```diff
@@ -3008,6 +3009,132 @@ export class PgDataStore extends (EventEmitter as { new (): DataStoreEventEmitte
...
+    offset,
+    height,
+  }: {
+    stxAddress: string;
+    limit: number;
+    offset: number;
+    height?: number;
+  }): Promise<{ results: DbTxWithStxTransfers[]; total: number }> {
+    return this.queryTx(async client => {
+      const queryParams: any[] = [stxAddress, limit, offset];
```
kyranjamie, line 3024:
```suggestion
      const queryParams: [string, number, number] = [stxAddress, limit, offset];
```
?

kyranjamie (approved):

PR #552, opened 2021-04-15
fix: handle invalid BTC addresses in import

hstove:
On the latest API instance, we noticed an issue where existing BNS names were not imported. Digging in, it looks like the issue was caused by an invalid BTC address (`1AafnuydZLQJ1phQZd7MeYLbcWA6Ewwqp5`). We run `b58toC32` which will throw if the address is invalid. I believe this was causing the whole import to fail. Invalid subdomain addresses are "normal", because they're off-chain.

This adds a catch block for invalid conversions, and instead the original value (the BTC address) is used as the `owner`. I've tested this with the full import, and it seems to be working.

### DIFF
--- a/src/import-v1/index.ts
+++ b/src/import-v1/index.ts
@@ -204,6 +204,15 @@ class SubdomainZonefileParser extends stream.Transform {
   }
 }
 
+// Convert a BTC address to STX, otherwise return the original value if invalid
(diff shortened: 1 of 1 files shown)


asimm241 (approved):
LGTM

zone117x (approved):
Thanks @hstove!

PR #556, opened 2021-04-19
Fix/bundle genesis data

zone117x:
Fixes https://github.com/blockstack/stacks-blockchain-api/issues/393

The chainstate.txt is now committed to the repo, and the locked STX data from v1 is automatically imported when ran in production mode.

Optimized the import process for token vesting. Previously took around 20-30 seconds on my machine, now takes less than 2 seconds.

The committed chainstate.txt is gzipped to only ~8MB. Uses NodeJS's native zlib to decompress, with essentially zero overhead.


src/import-v1/c32helper.ts
```diff
@@ -0,0 +1,102 @@
+// The b58ToC32 function from https://github.com/blockstack/c32check, optimized for speed in NodeJS.
```
asimm241, line 1: is there a need for separate test cases for this?

asimm241 (approved):
other than my comment above , LGTM.

src/import-v1/c32helper.ts
zone117x: Tests added

asimm241 (approved):
@zone117x LGTM!

noumanahmads (approved):
@zone117x Looks good to me.

PR #559, opened 2021-04-20
feat: make the v2 proxy cache control file configurable via env var #519

### LINKED ISSUE #519: Make cache-control file name/path customizable
When running the API in k8s, I'm unable to override the default `.proxy-cache-control.json` file without also deleting the other files in the same directory. This is just how K8s works unfortunately.

I'd like the file name + path customizable via environment variable. When the env var is set, the API should look to the given path + filename for the cache control JSON file. This would allow me to mount the cache control JSON file in an empty directory and not the `/app` directory so I can preserve the other application files.

There is a workaround, but it would disable automatic updates to the file, meaning the pod would have to be restarted to get the update (rendering the chokidar usage pretty useless in this scenario).

zone117x:
Closes #519


CharlieC3 (approved):
:shipit: Thank you!

PR #560, opened 2021-04-20
Fix redirect url issue

asimm241:
## Description
BNS's `v1/names/:fqn` API redirects to incorrect URL. This PR fixes that. 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/routes/bns/names.ts
+++ b/src/api/routes/bns/names.ts
@@ -63,7 +63,7 @@ export function createBnsNamesRouter(db: DataStore): RouterWithAsync {
         const namePart = name.split('.').slice(1).join('.');
         const resolverResult = await db.getSubdomainResolver({ name: namePart });
         if (resolverResult.found) {
-          res.redirect(resolverResult.result + req.url);
+          res.redirect(`${resolverResult.result}/v1/names${req.url}`);
           next();
         }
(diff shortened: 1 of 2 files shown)


src/api/routes/bns/names.ts
```diff
@@ -63,7 +63,7 @@ export function createBnsNamesRouter(db: DataStore): RouterWithAsync {
         const namePart = name.split('.').slice(1).join('.');
         const resolverResult = await db.getSubdomainResolver({ name: namePart });
         if (resolverResult.found) {
-          res.redirect(resolverResult.result + req.url);
+          res.redirect(`${resolverResult.result}/v1/names${req.url}`);
           next();
```
zone117x, line 67: Seems like this should `return` after `next()`?

asimm241: is there a need? idk. I'll add it just to be sure.

zone117x: If you don't then it will fall through to the `res.status(404)` below, and silently error I think. Not a big deal, but better to avoid it

zone117x (approved):

PR #561, opened 2021-04-21
fix: distinction between total locked and total unlocked token offering

zone117x:
Return both `total_locked` and `total_unlocked` in the account balance `token_offering_locked` response object, based off current block height.

Related to https://github.com/blockstack/stacks-blockchain-api/issues/515


asimm241 (approved):

noumanahmads (approved):
Looks good to me.

PR #563, opened 2021-04-21
Updating rosetta dockerfile

wileyj:
- Refactoring to use runit vs custom script
- clone the latest releases vs static versions

## Description
This change refactors the rosetta Dockerfile to use [runit](http://smarden.org/runit/) as a scheduler vs a customized shell script. This change was motivated by discussions with another team trying to use this Dockerfile, and realizing the static version was very out of date. Further work was added to standardize the Dockerfile, so there weren't things like custom users/scripts (particularly in the case of postgresql). 

The *builder* images are also refactored to more closely resemble the default Dockerfiles, as well as cloning the latest release for each of `stacks-blockchain-api` and `stacks-blockchain` repos. 

Other changes are more use of ARG and ENV variables to create the scripts, i.e.
```
printf '#!/bin/sh
exec svlogd -tt %s/postgresql' ${STACKS_LOG_DIR} > ${STACKS_SVC_DIR}/postgresql/log/run
```

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

### DIFF
--- a/stx-rosetta.Dockerfile
+++ b/stx-rosetta.Dockerfile
@@ -1,147 +1,124 @@
-### Build blockstack-core-sidecar API
-FROM node:lts-buster as build
 
-ARG API_TAG=v0.29.4
-
-RUN apt-get -y update && apt-get -y install openjdk-11-jre-headless
-
+#######################################################################
+## Build the stacks-blockchain-api
+FROM node:lts-buster as stacks-blockchain-api-build
+ARG STACKS_API_REPO=blockstack/stacks-blockchain-api
+ENV STACKS_API_REPO=${STACKS_API_REPO}
 WORKDIR /app
-
-RUN git clone -b $API_TAG --depth 1 https://github.com/blockstack/stacks-blockchain-api.git .
+RUN apt-get update -y \
+    && apt-get install -y \
+        curl \
+        jq \
+        openjdk-11-jre-headless
+RUN export API_TAG=$(curl -sL https://api.github.com/repos/${STACKS_API_REPO}/releases/latest | jq .tag_name | tr -d '"') \
+    && git clone -b ${API_TAG} --depth 1 https://github.com/${STACKS_API_REPO} .
 RUN echo "GIT_TAG=$(git tag --points-at HEAD)" >> .env
-RUN npm install && npm run build && npm prune --production
+RUN npm config set unsafe-perm true
+RUN npm install \
+    && npm run build \
+    && npm prune --production
 
-### Build stacks-node binary
-
-FROM rust:stretch as stacks-node-build
-
-ARG STACKS_TAG=2.0.5
-
-RUN mkdir -p /src /stacks
+#######################################################################
+## Build the stacks-blockchain
+FROM rust:buster as stacks-blockchain-build
+ARG STACKS_REPO=blockstack/stacks-blockchain
+ENV STACKS_REPO=${STACKS_REPO}
 WORKDIR /src
-RUN git clone -b $STACKS_TAG --depth 1 https://github.com/blockstack/stacks-blockchain.git .
-RUN rustup target add x86_64-unknown-linux-gnu
-RUN cargo build --release --workspace=./ --target x86_64-unknown-linux-gnu
-RUN cp -R /src/target/x86_64-unknown-linux-gnu/release/. /stacks
+RUN apt-get update -y \
+    && apt-get install -y \
+        curl \
+        jq
+RUN export STACKS_TAG=$(curl -sL https://api.github.com/repos/${STACKS_REPO}/releases/latest | jq .tag_name | tr -d '"') \
+    && git clone -b ${STACKS_TAG} --depth 1 https://github.com/${STACKS_REPO} .
+RUN cd testnet/stacks-node \
+    && cargo build --features monitoring_prom,slog_json --release
+RUN mkdir -p /out \
+    && cp /src/target/release/stacks-node /out
 
-### Fetch stacks-node binary
-
-### Begin building base image
-FROM ubuntu:focal
-
+#######################################################################
+## Build the final image with all components from build stages
+FROM debian:buster
 ARG STACKS_NETWORK=testnet
-
-SHELL ["/bin/bash", "-c"]
-
-### Install utils
-RUN apt-get update
-RUN apt-get install -y sudo curl pslist
-
-### Set noninteractive apt-get
-RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
-
-### Storage goes in /data; see https://www.rosetta-api.org/docs/standard_storage_location.html
-RUN mkdir -p /data
-
-### Install nodejs
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
-RUN apt-get install -y nodejs
-
-### stacky user ###
-# see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
-RUN useradd -l -u 33333 -G sudo -md /data/stacky -s /bin/bash -p stacky stacky \
-    # passwordless sudo for users in the 'sudo' group
-    && sed -i.bkp -e 's/%sudo\s\+ALL=(ALL\(:ALL\)\?)\s\+ALL/%sudo ALL=NOPASSWD:ALL/g' /etc/sudoers
-ENV HOME=/data/stacky
-WORKDIR $HOME
-USER stacky
-RUN sudo chown -R stacky:stacky $HOME
-RUN mkdir /data/stacky/.bashrc.d
-
-### Node.js
-RUN node -e 'console.log("Node.js runs")'
-
-### Setup stacks-node
-COPY --from=stacks-node-build /stacks/stacks-node stacks-node/
-ENV PATH="$PATH:$HOME/stacks-node"
-
-### Setup stacks-blockchain-api
-COPY --from=build /app stacks-blockchain-api
-
-#### Copy stacks-node mocknet config
-RUN cp stacks-blockchain-api/stacks-blockchain/*.toml .
-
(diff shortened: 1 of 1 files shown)


wileyj:
ping @zone117x

stx-rosetta.Dockerfile
```diff
@@ -1,147 +1,124 @@
...
-  rkill -9 $stacks_api_pid\n\
-  pg_stop\n\
-  rm -rf $PGDATA\n\
-  sleep 5\n\
-done\n\
-' >> run.sh && chmod +x run.sh
-
+ENV MAINNET_STACKS_CHAIN_ID=0x00000001
+ENV TESTNET_STACKS_CHAIN_ID=0x80000000
+ENV V2_POX_MIN_AMOUNT_USTX=90000000260
```
CharlieC3, line 69: I think this is handled automatically now. @zone117x can you confirm?

stx-rosetta.Dockerfile
```diff
@@ -1,147 +1,124 @@
...
-
-ARG API_TAG=v0.29.4
-
-RUN apt-get -y update && apt-get -y install openjdk-11-jre-headless
 
+#######################################################################
+## Build the stacks-blockchain-api
+FROM node:lts-buster as stacks-blockchain-api-build
+ARG STACKS_API_REPO=blockstack/stacks-blockchain-api
+ENV STACKS_API_REPO=${STACKS_API_REPO}
```
CharlieC3, line 6: I think this ENV var can be removed. `${STACKS_API_REPO}` will just use the `ARG` like it's an ENV var. The ENV var directive should be used when the env variable needs to be available in the resulting image.

stx-rosetta.Dockerfile
CharlieC3, line 69: Can we combine all of these `ENV` directives into one directive? It'll result in fewer layers in the image and decrease its size. Also if any of them aren't used in the final image's runtime, you can just use their `ARG` counterpart due to the same reason in my other comment.

stx-rosetta.Dockerfile
```diff
@@ -1,147 +1,124 @@
...
+    && apt-get clean \
+    && rm -rf /var/cache/apt/* /var/lib/apt/lists/* /tmp/* ${STACKS_SVC_DIR}/getty*
+COPY --from=stacks-blockchain-build /out ${STACKS_BLOCKCHAIN_DIR}
+COPY --from=stacks-blockchain-api-build /app ${STACKS_BLOCKCHAIN_API_DIR}
+RUN cp ${STACKS_BLOCKCHAIN_API_DIR}/stacks-blockchain/Stacks-mocknet.toml ${STACKS_BLOCKCHAIN_DIR}/Stacks-testnet.toml \
+    && cp ${STACKS_BLOCKCHAIN_API_DIR}/stacks-blockchain/Stacks-mocknet.toml  ${STACKS_BLOCKCHAIN_DIR}/Stacks-mocknet.toml 
+
+###################################
+##  runit service files
+RUN printf '#!/bin/sh\nexec 2>&1\n[ ! -d %s ] && mkdir -p %s && chown -R postgres:postgres %s && gosu postgres /usr/lib/postgresql/%s/bin/pg_ctl init -D %s\nexec gosu postgres /usr/lib/postgresql/%s/bin/postmaster -D %s' ${PG_DATA} ${PG_DATA} ${PG_DATA} ${PG_VERSION} ${PG_DATA} ${PG_VERSION} ${PG_DATA} > ${STACKS_SVC_DIR}/postgresql/run \
```
CharlieC3, line 103: Have you considered committing the ruinit service files to this repo, and `COPY`ing them into the image? Might be easier to maintain and read that way.

stx-rosetta.Dockerfile
```diff
@@ -1,147 +1,124 @@
...
-  pg_stop\n\
-  rm -rf $PGDATA\n\
-  sleep 5\n\
-done\n\
-' >> run.sh && chmod +x run.sh
-
+ENV MAINNET_STACKS_CHAIN_ID=0x00000001
+ENV TESTNET_STACKS_CHAIN_ID=0x80000000
+ENV V2_POX_MIN_AMOUNT_USTX=90000000260
+RUN apt-get update \
```
CharlieC3, line 70: Combining this `RUN` command with the following three will further decrease the number of layers and size of the resulting image.

stx-rosetta.Dockerfile
wileyj: unless the repo name changes, in which case this is a single change or a build-arg vs multiple changes in the file

stx-rosetta.Dockerfile
wileyj: Yes, but it would also be incredibly hard to read. the goal here isn't a small image

stx-rosetta.Dockerfile
wileyj: Possible, sure - but changes to the repo would require a full rebuild since the cache layers would no longer be used. 
Since there are many parts of this process, the goal was to create an image that can be built using cache layers as quickly as is feasible.

stx-rosetta.Dockerfile
wileyj: Same as above - any change in the process/dockerfile would invalidate the cache layers and cause a longer rebuild. 
Can they all be combined? yes - iniitially that's what i was doing. but any small change required a full rebuild vs updating a single layer or two.

stx-rosetta.Dockerfile
CharlieC3: I'm saying you can delete just this ENV var and the image will work the same as it did before. There's no benefit to declaring an ARG then setting an ENV var to it if the env var won't be used in the final image because it's redundant. If the repo name changes, you'd update the ARG default value.

stx-rosetta.Dockerfile
CharlieC3: That's fine if you want to keep it that way, I don't find it difficult to read with one directive. Also it seems `ENV` directives don't create additional layers anymore either which is nice.

stx-rosetta.Dockerfile
CharlieC3: > but changes to the repo would require a full rebuild since the cache layers would no longer be used.

If you `COPY` only those ruinit files, changes to the repo wouldn't require a full rebuild. That only happens when you write `COPY . .` for example. Either way will work the same, it's just difficult to read or maintain the scripts when they're embedded.

stx-rosetta.Dockerfile
CharlieC3: > any change in the process/dockerfile would invalidate the cache layers and cause a longer rebuild.

Only true if you `COPY` the dockerfile into the image. For `RUN` directives the docker engine only compares the text of the old `RUN` vs the text of the new `RUN` to determine if it can be reused from cache. These are only minor suggestions, so feel free to do what you like.

wileyj:
ping @zone117x

zone117x:
Thanks @wileyj this looks good to me. Do you think we should merge this to master now, or wait until we merge API develop branch to master?

stx-rosetta.Dockerfile
zone117x: Yes, this env var is no longer used anywhere.

wileyj:
> Thanks @wileyj this looks good to me. Do you think we should merge this to master now, or wait until we merge API develop branch to master?

Up to you! I've shared the file from my branch with the folks who needed it, so there's no rush - pinging just to get it off my plate.

zone117x:
> Up to you! I've shared the file from my branch with the folks who needed it, so there's no rush - pinging just to get it off my plate.

Alright, there could be some changes required once develop branch is ready, and that's expected this week. So I think holding off until then makes sense.

wileyj:
> > Up to you! I've shared the file from my branch with the folks who needed it, so there's no rush - pinging just to get it off my plate.
> 
> Alright, there could be some changes required once develop branch is ready, and that's expected this week. So I think holding off until then makes sense.

👍 changed the base branch to `develop` as well.

wileyj:
ping @zone117x

zone117x:
@wileyj the only concern I have about this PR is the move away from pinned versions (of the stacks-node repo at least). 
See this discussion for more context on versioning issues: https://github.com/blockstack/stacks-blockchain-api/discussions/567

I do think this dockerfile change is a step in the right direction, because the hardcoded versions tend to go stale. 

Should we merge this PR with the possibility that this dockerfile could build images with stacks-api and stacks-node versions that are incompatible with each other (when the release cadences don't match up)?

wileyj:
> #567

Interesting...I don't have any opinion directly on that discussion, but also -  nothing prevents us from setting **specific** versions in the Dockerfile vs trying to dynamically retrieve the version through github api. 

The reason I made the Dockerfile this way was to prevent having to manually update the version numbers each time a 
 new release is cut for either repo (stacks-blockchain or stacks-blockchain-api). If that's not preferred, I can roll back that idea to hardcode the versions. 

Based on the discussion you linked, is there a pair of versions you'd like to start with initially?
i.e.
https://github.com/blockstack/stacks-blockchain-api/releases/tag/v0.58.0
https://github.com/blockstack/stacks-blockchain/releases/tag/2.0.11.0.0


Once I have the specific versions I can update the Dockerfile and we can get this off of our plate. 
The caveat being we'll need to ensure we keep the file up to date with current versions.

zone117x:
> Based on the discussion you linked, is there a pair of versions you'd like to start with initially?
> i.e.
> https://github.com/blockstack/stacks-blockchain-api/releases/tag/v0.58.0
> https://github.com/blockstack/stacks-blockchain/releases/tag/2.0.11.0.0

Yes these would be good versions to set the dockerfile to for now 👍 

> The caveat being we'll need to ensure we keep the file up to date with current versions.

Yeah, this is definitely a pain, hopefully something we can figure out in #567 eventually.

zone117x (approved):

PR #566, opened 2021-04-26
Fix empty subdomains

### LINKED ISSUE #564: When resolving registrar URL, the result is the URL of the API itself
The latest issue around subdomain registrations is that, when you try and resolve an unregistered subdomain, the request fails because of an infinite redirect.

URL to try: https://stacks-node-api.stacks.co/v1/names/some_unregistered_name.id.stx

I believe this is because the [DB query](https://github.com/blockstack/stacks-blockchain-api/blob/7f8886024a4caf0d57a825dc2b526e6b94c4c1ef/src/datastore/postgres-store.ts#L3945) for `getSubdomainResolver` is returning the URL of the API - not the registrar.

asimm241:
## Description
There were empty subdomains being stored in the subdomains table.  This PR fixes that. 

Closes #564.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review

### DIFF
--- a/src/datastore/common.ts
+++ b/src/datastore/common.ts
@@ -423,7 +423,7 @@ export interface DbTokenOfferingLocked {
 
 export interface DataStore extends DataStoreEventEmitter {
   getSubdomainResolver(name: { name: string }): Promise<FoundOrNot<string>>;
-  getUnresolvedSubdomain(txId: string, indexBlockHash: string): Promise<FoundOrNot<DbBnsSubdomain>>;
+  getNameCanonical(txId: string, indexBlockHash: string): Promise<FoundOrNot<boolean>>;
   getBlock(blockHash: string): Promise<FoundOrNot<DbBlock>>;
   getBlockByHeight(block_height: number): Promise<FoundOrNot<DbBlock>>;
   getCurrentBlock(): Promise<FoundOrNot<DbBlock>>;
--- a/src/datastore/memory-store.ts
+++ b/src/datastore/memory-store.ts
@@ -95,10 +95,7 @@ export class MemoryDataStore extends (EventEmitter as { new (): DataStoreEventEm
     });
   }
 
-  getUnresolvedSubdomain(
-    txId: string,
-    indexBlockHash: string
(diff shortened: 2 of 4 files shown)


zone117x (commented):
I think this change need paired with a change to the subdomain attachment handling here: https://github.com/blockstack/stacks-blockchain-api/blob/072ac20f132ddc300af3a4d3858ab070196087a0/src/event-stream/event-server.ts#L476-L483

This is only responsible for determining if the subdomain should be marked as canonical (detect if the tx was forked before the attachment was received). 
I think it can be replaced with something like
```sql
SELECT canonical FROM names
WHERE tx_id = {attachment.tx_id}
AND index_block_hash = {attachment.index_block_hash}
```

markmhx:
Does this fully resolve the following issue? https://github.com/blockstack/stacks-blockchain-api/issues/564

asimm241:
> @asimm241 Does this fully resolve the following issue? #564
> 
> @zone117x might this go out with the API release today or do you prefer we release against testnet first?

yes.

asimm241:
> I think this change need paired with a change to the subdomain attachment handling here:
> 
> https://github.com/blockstack/stacks-blockchain-api/blob/072ac20f132ddc300af3a4d3858ab070196087a0/src/event-stream/event-server.ts#L476-L483
> 
> This is only responsible for determining if the subdomain should be marked as canonical (detect if the tx was forked before the attachment was received).
> I think it can be replaced with something like
> 
> ```sql
> SELECT canonical FROM names
> WHERE tx_id = {attachment.tx_id}
> AND index_block_hash = {attachment.index_block_hash}
> ```

can we add a test case for this forking?

zone117x:
@asimm241 Yes we should add a test to mark the name as non-canonical and ensure a subsequent subdomain ends up marked as non-canonical. In the meantime, this PR looks good to merge into develop.

zone117x (approved):

PR #576, opened 2021-05-04
Feat/tokens metadata

asimm241:
## Description
This PR adds the support for Fungible token metadata and Non Fungible token metadata.  It checks the deployed contract's compliance with [sip-9](https://github.com/stacksgov/sips/blob/friedger-main/sips/sip-009/sip-009-nft-standard.md) and [sip-10](https://github.com/hstove/sips/blob/feat/sip-10-ft/sips/sip-010/sip-010-fungible-token-standard.md).

It checks the standards by comparing contract abi. If it is compliance, It fetches data from the contract by making read-only call to: 

#### For FT:
- get-name
- get-token-uri
- get-symbol
- get-decimals 

#### For NFT:

- get-last-token-id
- get-token-uri

Using the token uri it fetches the metdata from the web, and store the relevent data in the database. 

It exposes two APIs
1. `v1/tokens/:contractId/ft/metadata`
2. `v1/tokens/:contractId/nft/metadata`

 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

@zone117x do we need to add documentation

## Testing information
`npm run test:integration:tokens` 
`npm run test:token`

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


zone117x (commented):
So far so good. I think next steps are ingesting token metadata from the event server.

docs/index.d.ts
```diff
@@ -1021,6 +1021,44 @@ export interface RosettaNetworkStatusResponse {
...
+  /**
+   * A URI pointing to a resource with mime type image/* representing the asset to which this token represents. The API may provide a URI to a cached resource, dependending on configuration. Otherwise, this can be the same value as the canonical image URI.
+   */
+  image_uri: string;
+  /**
+   * The original image URI specified by the contract. A URI pointing to a resource with mime type image/* representing the asset to which this token represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive.
+   */
+  image_canonical_uri: string;
+}
+
```
aulneau, line 1061: I'm curious, where did this schema originate? it doesn't follow the standards that have been worked on: 
https://github.com/stacksgov/sips/blob/friedger-main/sips/sip-009/sip-009-nft-standard.md
https://github.com/stacksgov/sips/blob/hstove-feat/sip-10-ft/sips/sip-010/sip-010-fungible-token-standard.md

agraebe:
@asimm241 there is a team (Swapr) that would rely on the completion of this PR by the end of May. would this timeline work for you? 

cc @aulneau @andresgalante

aulneau:
> @asimm241 there is a team (Swapr) that would rely on the completion of this PR by the end of May. would this timeline work for you?
> 
> cc @aulneau @andresgalante

Expanding on this, this PR would enable our wallets to support the FT ([SIP 010](https://github.com/stacksgov/sips/blob/hstove-feat/sip-10-ft/sips/sip-010/sip-010-fungible-token-standard.md)) and NFT ([SIP 009](https://github.com/stacksgov/sips/blob/friedger-main/sips/sip-009/sip-009-nft-standard.md)) standards in that we'd be able to properly fetch metadata associated with a given token. This would allow us to display the ticker, name, decimals (for fungible tokens) and the asset image. We have some options on how we could do this ourselves via contract calls in the wallets, but an API based approach would be much better in the long term.

psq:
I have the same questions as @aulneau, and not clear how this would get used (i.e. queried) to help wallets and explorers.   for example, there is no event generated when a new sip-011 token is created (besides a contract deployment), so how would the api node know that it needs to call a bunch of functions on a contract to retrieve things like symbol, name, decimals, and metadata (which requires 2 rpc call), and the sip-010 (and sip-009) meta data schema is different (albeit "compatible", so far).

And with the experience building `swapr`, the metadata probably should not have a fixed schema, or at a minimum allow for storing/retrieving extra top level elements.

asimm241:
> @asimm241 there is a team (Swapr) that would rely on the completion of this PR by the end of May. would this timeline work for you?
> 
> cc @aulneau @andresgalante

I think so.

docs/index.d.ts
zone117x: Which part @aulneau? The canonical image URI was my suggestion to factor in how an API instance should expose a resource like an image to clients.

zone117x:
> there is no event generated when a new sip-011 token is created (besides a contract deployment), so how would the api node know that it needs to call a bunch of functions on a contract to retrieve things like symbol, name, decimals, and metadata (which requires 2 rpc call)

The API can check the ABI in the contract-deploy tx handler. If the ABI "contains" the expected sip-11 or sip-09 elements it can be passed to another component responsible for retrieving the initial metadata. 

> and the sip-010 (and sip-009) meta data schema is different (albeit "compatible", so far).

The API can/will expose differences here where necessary?

> And with the experience building `swapr`, the metadata probably should not have a fixed schema, or at a minimum allow for storing/retrieving extra top level elements.

Can you elaborate or give some examples here?

psq:
> The API can check the ABI in the contract-deploy tx handler. If the ABI "contains" the expected sip-11 or sip-09 elements it can be passed to another component responsible for retrieving the initial metadata.

That will not work consistently, see for example https://github.com/tokensoft/tokensoft_token_stacks/blob/main/contracts/tokensoft-token.clar#L110-L119, where the meta data url can be changed at will, and of course, the meta data file itself can be changed even more often (not that would know any token that rebranded, ever, 😉 )

I'm also considering doing something similar where I have some generic swapr token that are pre-deployed (simplifies the ui), but gets configured with the correct data, when creating a new pair for example.

> Can you elaborate or give some examples here?

I'm experimenting with adding one more image format (explicitly vector) for one, and for a second, swapr will need some extra meta data to help generate the proper post conditions (find the token name, and some token may require special case, so we need something fairly fluid that would store whatever is in the metadata file (as long as it is compliant with the base schema).


Rather than storing information in the database at deployment, it may be better to consider querying the token functions on a regular basis (less than daily after the first time?), so you can provide some kind of caching, but can also guarantee it won't get stale.

zone117x:
> where the meta data url can be changed at will, and of course, the meta data file itself can be changed even more often

This was discussed in the original RFC that requested this API feature. For now we are considering the values as static. If there's a strong desire to have dynamic metadata then I'd suggest an amendment to SIP-10 and SIP-09 to specify a contract print event indicating when metadata has been changed.

psq:
I just checked the tokens from Arkadiko, and they are also using a variable for the token uri, so this seems pretty universal amongst the major projects working on fungible tokens, at least so far, that they assume their metadata is dynamic.  Seems correct to me.

zone117x:
I'd still suggest a new SIP/amendment that specifies contract events should be emitted when metadata is updated. The API could deal with TTLs / cache-invalidation for this potentially in the future, but I would not expect that to be an initial feature.

docs/index.d.ts
aulneau: ideally for the FT schema these fields would be available:

- `symbol`
- `decimals`

aulneau:
@zone117x question for you, is it possible scope this PR such that in addition to fetching the token_uri data, it would be able to expose the `decimals` and `symbol` that are defined in the traits? We need this for the wallets, and apps like `swapr` would depend on this, too.

psq:
currently, swapr is using `call-read` on each token to gather each data point, which is reasonably fast for a handful of tokens, but won't scale very well, so having one endpoint to gather all the data for a token (name, symbol, decimals, uri, metadata pointed to by uri is not `none`).  For the LP tokens, I need to gather more data, but that's too specialized for a generic api).

src/datastore/memory-store.ts
```diff
@@ -604,4 +605,11 @@ export class MemoryDataStore extends (EventEmitter as { new (): DataStoreEventEm
   ): Promise<FoundOrNot<AddressTokenOfferingLocked>> {
     throw new Error('Method not implemented');
   }
+
+  getftMetadata(contractId: string): Promise<FoundOrNot<DbAssetMetadata>> {
```
kyranjamie, line 609:
```suggestion
  getFtMetadata(contractId: string): Promise<FoundOrNot<DbAssetMetadata>> {
```

src/datastore/common.ts
```diff
@@ -612,6 +619,9 @@ export interface DataStore extends DataStoreEventEmitter {
     address: string,
     blockHeight: number
   ): Promise<FoundOrNot<AddressTokenOfferingLocked>>;
+
+  getftMetadata(contractId: string): Promise<FoundOrNot<DbAssetMetadata>>;
```
kyranjamie, line 623:
```suggestion
  getFtMetadata(contractId: string): Promise<FoundOrNot<DbAssetMetadata>>;
```

docs/api/tokens/fungible-token.schema.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "fungible-token-metadata",
+  "title": "FTMetadataResponse",
```
kyranjamie, line 4: I would suggest we follow full naming conventions here. We can later create an alias type to a shortened name

```suggestion
  "title": "FungibleTokenMetadataResponse",
```

docs/api/tokens/non-fungible-token.schema.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "non-fungible-token-metadata",
+  "title": "NFTMetadataResponse",
```
kyranjamie, line 4:
```suggestion
  "title": "NonFungibleTokenMetadataResponse",
```

andresgalante:
@asimm241 I need some help to deblur the next steps on this PR:

1. We need an answer to @aulneau's question [here](https://github.com/blockstack/stacks-blockchain-api/pull/576#issuecomment-839966727) and
2. You mentioned that you think this can be done before the end of the month, what the level of certainty?

@zone117x If this requires an amendment to SIP-10 and SIP-09, should we involved the authors? Is it a blocker to move on with this PR?

zone117x:
> is it possible scope this PR such that in addition to fetching the token_uri data, it would be able to expose the decimals and symbol that are defined in the traits? 

Yes. When SIP-09/SIP-10 contracts are deployed, the API should retrieve and store these values.

> If this requires an amendment to SIP-10 and SIP-09, should we involved the authors? Is it a blocker to move on with this PR?

Supporting dynamic metadata changes is not a blocker for this PR, it's a feature request. In order for the API to support that feature, it involves SIP amendments.

psq:
> Supporting dynamic metadata changes is not a blocker for this PR,

sure

> it's a feature request.

aren't you creating more work if we know this is not going work for most tokens, as opposed to try to make something that works out of the gate?

> In order for the API to support that feature, it involves SIP amendments.

can't seem be convinced this is the solution.  No event will be generated if the metadata json file is updated, and that metadata file was purposefully created as something outside the contract so it could be modified easily, while the contract can not.  So in order to not forever present stale data, the api node would need to query that file on a regular basis anyway (weekly or less is probably fine, it is probably not going to change that often...), and if you need to do this, you might as well get the up to date uri value before checking the file content for changes.

zone117x:
> aren't you creating more work if we know this is not going work for most tokens, as opposed to try to make something that works out of the gate?

I'm not sure if that's true or meaningful with the current sample size of tokens. 

> can't seem be convinced this is the solution. No event will be generated if the metadata json file is updated, and that metadata file was purposefully created as something outside the contract so it could be modified easily, while the contract can not. So in order to not forever present stale data, the api node would need to query that file on a regular basis anyway (weekly or less is probably fine, it is probably not going to change that often...), and if you need to do this, you might as well get the up to date uri value before checking the file content for changes.

A contract-call can be made to signal that metadata should be updated. If you're aiming to have remote JSON file rapidly changing such that an "expiry" event via contract-call is too much too ask for, then setting an arbitrary TTL of a week or a month or whatever won't solve that problem anyway. 

The clear solution here is for the API to be notified when token metadata is updated, for example via contract events. You're asking the API to implement a kind of statefulness that does not yet exist in the codebase. Different API instances can easily end up with different views/responses for the same requested resource. Additionally, attempting to cache with absolutely zero indication of a TTL will result in either scalability/wastefulness problems on one end, or long-expired data on the other end. Trying to solve that by picking an interval "in the middle" like a week doesn't solve that problem, rather it just introduces a bit of both problems.

That said, I'm not absolutely against implementing a generic cache refreshing mechanism like you're suggesting, but it _is_ a kludge and should not be treated as the solution for dynamically updating metadata.

psq:
Thank you for the dialogue, @zone117x, it helps better understand where you are coming from and what your concerns are.  You make valid points.

Getting a bit late for me here, so I'll sleep on it and keep on the conversation tomorrow.

asimm241:
@aulneau @kyranjamie @psq can you guys review this PR. Meanwhile, I'm working on test cases!

aulneau (approved):

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,397 @@
...
+    const symbolCV = this.checkAndParseString(contractCallSymbol);
+    const decimalsCV = this.checkAndParseUintCV(contractCallDecimals);
+
+    if (uriCV) {
+      const metadata = await this.getMetadataFromUri(uriCV.data);
+
+      const fungibleTokenMetadata: DbFungibleTokenMetadata = {
+        name: nameCV ? nameCV.data : metadata.name, //prefer the on-chain name
+        description: metadata.description,
+        image_uri: metadata.imageUrl,
```
psq, line 223: where does `imageUrl` comes from?  the json itself will include just `image` for the uri of the graphic file to use.

psq:
> @aulneau @kyranjamie @psq can you guys review this PR. Meanwhile, I'm working on test cases!

except that question I have about the name of the key used to retrieve the image uri for FTs, this should cover most of the on chain data.

However, more work will be required later on to better support the external metadata, as you can not assume that either the uri or the content of the json are immutable.  The whole point of having this file off chain was for it to be changed easily.

src/event-stream/tokens-contract-handler.ts
asimm241: I'm using this as a ref https://ipfs.io/ipfs/QmPAg1mjxcEQPPtqsLoEcauVedaeMH81WXDPvPx3VC5zUz. It has `imageURL` key.

asimm241: am I using the right value?

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,397 @@
...
+              return false;
+            }
+          });
+        }
+        return true;
+      }
+      return false;
+    });
+    return found != undefined;
+  }
```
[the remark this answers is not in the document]
asimm241: This method actually checks if the given fun is in the list of functions, for this to be true the arguments of the functions should also match. 

I think you're right about line 352. Let me see that.

src/event-stream/tokens-contract-handler.ts
psq: see the FT sip schema (yeah, not so readable, but you can paste the schema into https://extendsclass.com/json-schema-validator.html to get an an example file that follows the schema), but here's the image key https://github.com/stacksgov/sips/pull/5/files#diff-b83747873c039e89a2dbcc593f0934ce59aaab9f58c8cacaefe4c01f6f2bdc95R143

some example files that follow the schema: https://swapr.finance/tokens/xusd.json, https://swapr.finance/tokens/token-soft-xbtc.json (you can ignore the `vector` key, but this just shows people can extend the original schema, and you can ignore these.

that's for FTs, not following the NFT SIP as closely, but there is a new SIP to describe this, so maybe not ready.  Others may have more precise information.

asimm241: It seems we have to treat FT and NFT token uri differently. It would be very helpful if anybody has concrete information regarding it.

psq: I'm only familiar with the FT side of things, so maybe @friedger could chime in?

andresgalante:
Is this code deployed so we can test it on the wallet? If not, can we deploy it on regtest?

src/api/init.ts
```diff
@@ -172,6 +173,17 @@ export async function startApiServer(datastore: DataStore, chainId: ChainID): Pr
     })()
   );
 
+  //Setup routes for token metadata
+  app.use(
+    '/v1',
```
zone117x, line 178: Any reason why this isn't defined under the `/extended/v1` route?

src/api/routes/tokenes/tokens.ts
```diff
@@ -0,0 +1,57 @@
+import { addAsync, RouterWithAsync } from '@awaitjs/express';
```
zone117x, line 1: Typo in the directory name `tokenes` -> `tokens`

src/api/init.ts
asimm241: I wasn't aware of the convention. let me change it.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,417 @@
...
+        await this.storeNFTMetadata(nonFungibleTokenMetadata);
+      }
+    } catch (error) {
+      logger.error('erro: error handling NFT contract', error);
+    }
+  }
+
+  /**helpng method for creating http url */
+  private makeHostedUrl(uri: string): string {
+    if (uri.includes('http')) return uri;
```
zone117x, line 279: We should try to use the native `URL` API rather than another lib `uri-js`. And we should explicitly check for the supported protocols (http, https, ipfs). 
Then this code would look something like 
```ts
const parsedUri = new URL(uri);
if (parsedUri.protocol === 'http:' || parsedUri.protocol === 'https:') return uri;
if (parsedUri.protocol === 'ipfs:') { .... }
throw new Error(`Unsupported uri protocol: ${uri}`);
```

src/event-stream/tokens-contract-handler.ts
friedger: There is no specification for the meta data that should be returned by get-token-uri.
I usually recommend the meta data standard by open sea https://docs.opensea.io/docs/metadata-standards. Risidio are reviewing their current meta data schema.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,417 @@
...
+    const parsedURI = URI.parse(uri);
+    return `${PUBLIC_IPFS}/${parsedURI.host}${parsedURI.path}`;
+  }
+
+  async makeApiCall<Type>(url: string): Promise<Type> {
+    const result = await fetch(url);
```
zone117x, line 285: Is the stacks-node guaranteed to respond to this `/v2/...` request while its event dispatcher is waiting on the API for a response to the `/new_burn_block` event request, which this path is currently holding up? cc @kantai

jcnelson: No -- the relayer and p2p thread both stalled while the event dispatcher is transmitting data to the event observer.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  /**
+   *Store ft metadata to db
+   */
+  private async storeFTMetadata(ft_metadata: DbFungibleTokenMetadata) {
+    try {
+      await this.db.updateFtMetadata(ft_metadata);
+    } catch (error) {
+      logger.error(`Error: error occured while updating FT metadata`);
```
zone117x, line 336: Should throw here otherwise error is swallowed rather than handled more appropriately in a function higher up in the call graph.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  /**
+   *store NFT Metadata to db
+   */
+  private async storeNFTMetadata(nft_metadata: DbNonFungibleTokenMetadata) {
+    try {
+      await this.db.updateNFtMetadata(nft_metadata);
+    } catch (error) {
+      logger.error(`Error: error occured while updating NFT metadata`);
```
zone117x, line 347: Same here, should throw

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  /**
+   * check if the fun  exist in the function list
+   * @param fun - function to be found
+   * @param functionList - list of functions
+   * @returns - true if function is in the list false otherwise
+   */
+  private findFunction(fun: ClarityAbiFunction, functionList: ClarityAbiFunction[]): boolean {
+    const found = functionList.find(standardFunction => {
+      if (standardFunction.name != fun.name || standardFunction.args.length != fun.args.length)
```
zone117x, line 367: Should use strict equality `!==`. We should get the eslint rule fixed for this (in a later PR)

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,397 @@
...
+        }
+        return true;
+      }
+      return false;
+    });
+    return found != undefined;
+  }
+
+  private checkAndParseUintCV(responseCV: ClarityValue): UIntCV | undefined {
+    if (responseCV.type == ClarityType.ResponseOk && responseCV.value.type == ClarityType.UInt) {
```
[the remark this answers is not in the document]
zone117x: Looks like this is still an issue

src/migrations/1621511823381_nft-metadata.ts
```diff
@@ -0,0 +1,44 @@
...
+      type: 'string', 
+      notNull: true, 
+    }
+    });
+
+    pgm.createIndex('nft_metadata', 'name');
+    pgm.createIndex('nft_metadata', 'description');
+    pgm.createIndex('nft_metadata', 'image_uri');
+    pgm.createIndex('nft_metadata', 'image_canonical_uri');
+    pgm.createIndex('nft_metadata', 'contract_id');
```
zone117x, line 38: An index should only be needed on `name` and `contract_id` here

src/migrations/1621511832113_ft-metadata.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+    });
+
+    pgm.createIndex('ft_metadata', 'name');
+    pgm.createIndex('ft_metadata', 'description');
+    pgm.createIndex('ft_metadata', 'image_uri');
+    pgm.createIndex('ft_metadata', 'image_canonical_uri');
+    pgm.createIndex('ft_metadata', 'symbol');
+    pgm.createIndex('ft_metadata', 'decimals');
+    pgm.createIndex('ft_metadata', 'contract_id');
```
zone117x, line 48: Probably only index on `symbol`, `name`, and `contract_id`. We probably won't ever be querying by `description` for example.

package.json
```diff
@@ -10,10 +10,12 @@
     "test": "cross-env NODE_ENV=development jest --config ./jest.config.js --coverage --runInBand",
     "test:rosetta": "cross-env NODE_ENV=development jest --config ./jest.config.rosetta.js --coverage --runInBand",
     "test:bns": "cross-env NODE_ENV=development jest --config ./jest.config.bns.js --coverage --runInBand",
+    "test:tokens": "cross-env NODE_ENV=development jest --config ./jest.config.tokens.js --coverage --runInBand",
```
zone117x, line 13: This new test suite should be added to CI in `.github/workflows/stacks-blockchain-api.yml`

src/datastore/postgres-store.ts
```diff
@@ -4032,6 +4034,104 @@ export class PgDataStore
...
+      name,
+      description,
+      image_uri,
+      image_canonical_uri,
+      contract_id,
+      symbol,
+      decimals,
+    } = ftMetadata;
+    return await this.queryTx(async client => {
+      const result = await client.query(
```
zone117x, line 4108: There are forking scenarios that aren't handled by the queries against the `ft_metadata` and `nft_metadata` tables. 

We can probably ignore the consequences for now, but we should open a new issue for `Handle re-orgs for the ft/nft metadata tables`

src/event-stream/event-server.ts
```diff
@@ -236,6 +237,17 @@ async function handleClientMessage(
         abi: JSON.stringify(tx.core_tx.contract_abi),
         canonical: true,
       });
+      //check if this contract uses fungible/non fungible tokens
+      if (tx.core_tx.contract_abi && hasTokens(tx.core_tx.contract_abi)) {
```
zone117x, line 241: This should also check that `tx.core_tx.status === 'success'`

src/event-stream/event-server.ts
```diff
@@ -236,6 +237,17 @@ async function handleClientMessage(
...
+        //TODO start it in a seperate thread
+        await new TokensContractHandler(
```
zone117x, line 243: It looks like this could deadlock the event handler. If the stacks-node is waiting for an http response to the `/new_block` event request, then the stacks-node will likely not respond to the `/v2/...` requests made during this call.

We should setup a promise queue that is set to evaluate 1 of these at a time. The lib `p-queue` is useful for that, and already being used, see example here:
https://github.com/blockstack/stacks-blockchain-api/blob/9ec64ad13403a7e82e6aa5d3648a17cb3133c809/src/event-stream/event-server.ts#L532

src/datastore/postgres-store.ts
asimm241: https://github.com/blockstack/stacks-blockchain-api/issues/639

src/event-stream/tokens-contract-handler.ts
asimm241: isn't NFT immutable?

src/event-stream/tokens-contract-handler.ts
```diff
@@ -302,9 +301,16 @@ export class TokensContractHandler {
...
-    return `${PUBLIC_IPFS}/${parsedURI.host}${parsedURI.path}`;
+    const parsedUri = new URL(uri);
+    if (parsedUri.protocol === 'http:' || parsedUri.protocol === 'https:') return uri;
+    if (parsedUri.protocol === 'ipfs:')
+      return `${PUBLIC_IPFS}/${parsedUri.host}${parsedUri.pathname}`;
+
+    if (parsedUri.protocol === 'ipns:')
+      return `${PUBLIC_IPFS}/${parsedUri.host}${parsedUri.pathname}`;
+
+    console.log('handle nft metadata url', uri);
```
zone117x, line 312: `logger.warn` instead?

src/api/routes/tokens/tokens.ts
```diff
@@ -0,0 +1,67 @@
+import { addAsync, RouterWithAsync } from '@awaitjs/express';
+import * as express from 'express';
+import { DataStore } from '../../../datastore/common';
+import {
+  FungibleTokenMetadataResponse,
+  NonFungibleTokenMetadataResponse,
+} from '@stacks/stacks-blockchain-api-types';
+
+export function createTokenRouter(db: DataStore): RouterWithAsync {
```
zone117x, line 9: This should include a route to enumerate the list of tokens, similar to `/extended/v1/tx` and `/extended/v1/block`.

src/api/init.ts
```diff
@@ -130,6 +131,7 @@ export async function startApiServer(datastore: DataStore, chainId: ChainID): Pr
       router.use('/debug', createDebugRouter(datastore));
       router.use('/status', (req, res) => res.status(200).json({ status: 'ready' }));
       router.use('/faucets', createFaucetRouter(datastore));
+      router.use('/tokens', createTokenRouter(datastore));
```
zone117x, line 134: Looks like these new routes are missing OpenAPI documentation at `/docs/openapi.yaml`

zone117x:
@asimm241 @blockstack/ux-team This has been deployed to https://stacks-node-api-tokens.stacks.co/extended/v1/tx for testing purposes

src/event-stream/tokens-contract-handler.ts
```diff
@@ -0,0 +1,505 @@
...
+      }
+      return uri;
+    }
+    const fetchableUrl = this.getFetchableUrl(uri);
+    return fetchableUrl.toString();
+  }
+
+  async performFetch<Type>(url: string): Promise<Type> {
+    const MAX_PAYLOAD_SIZE = 1_000_000; // 1 megabyte
+    const result = await fetch(url, { size: MAX_PAYLOAD_SIZE });
```
zone117x, line 347: @asimm241 I saw this ddos vulnerability (making the API server OOM by fetching gigs of data) and patched it here, but need a unit test, probably needs to be a fetch mock

src/tests-tokens/tokens-metadata-tests.ts
```diff
@@ -0,0 +1,280 @@
...
+      'nft-trait',
+      pKey,
+      'src/tests-tokens/test-contracts/nft-trait.clar',
+      api
+    );
+    const tx = await standByForTx(contract.txId);
+    if (tx.status != 1) logger.error('contract deploy error', tx);
+
+    // TODO: these example contracts need the 3rd party `get-token-uri` urls fetch-mocked
+    // so these tests pass if/when those resources become unavailable or change.
```
zone117x, line 209: @asimm241 I added a comment here regarding unit test dependencies on 3rd party resources. These tests can break at any moment, we need to implement fetch mocking to make these durable.

CharlieC3:
> I haven't had a chance to inspect the logs, @CharlieC3 would you be able to find any errors containing Error processing token contract?

@zone117x Found them, logs can be downloaded here:
[Mainnet - stacks-node-api-tokens - token contract errors.csv](https://github.com/blockstack/stacks-blockchain-api/files/6809127/Mainnet.-.stacks-node-api-tokens.-.token.contract.errors.csv)

andresgalante:
@aulneau @kyranjamie Would you be able to review the last changes on this PR and make sure that it has all the features we need? This is a call for the last chance to ask for them :)

src/tests-tokens/tokens-metadata-tests.ts
```diff
@@ -334,6 +323,17 @@ describe('api tests', () => {
     expect(query1.body.results.length).toStrictEqual(20);
   });
 
+  test('large payload test', () => {
+    //mock the response
+    const rawdata = fs.readFileSync('src/tests-tokens/test-data/large-size-test-data.json');
```
zone117x, line 328: We probably don't want to commit large files like this to the repo, especially if we increased the limit to 5MB or something. Can you just generate this data?

asimm241: Neat idea, I think I can.

Redwarrio:
Hello word

src/tests-tokens/tokens-metadata-tests.ts
zone117x: Thanks!

Redwarrio (approved):

zone117x:
> This has been deployed to https://stacks-node-api-tokens.stacks.co/extended/v1/tx for testing purposes.
> 
> The db shows zero entries for fungible token metadata, so something seems wrong here. In the meantime, @asimm241 we should work on adding a new sql table for the token processor queue, and make sure we store error result status and info in it.
> 
> The db shows 4 entries for non-fungible token metadata:
> 
> * https://stacks-node-api-tokens.stacks.co/extended/v1/tokens/SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.friedger-pool-nft/nft/metadata
> * https://stacks-node-api-tokens.stacks.co/extended/v1/tokens/SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.beeple/nft/metadata
> * https://stacks-node-api-tokens.stacks.co/extended/v1/tokens/SP1JSH2FPE8BWNTP228YZ1AZZ0HE0064PS6RXRAY4.beeple/nft/metadata
> * https://stacks-node-api-tokens.stacks.co/extended/v1/tokens/SP1K1A1PMGW2ZJCNF46NWZWHG8TS1D23EGH1KNK60.friedger-pool-nft/nft/metadata

@asimm241 have you had a chance to look at this issue, where no FTs were ingested for some reason? Charlie was able to find some logs:

> Found them, logs can be downloaded here:
> [Mainnet - stacks-node-api-tokens - token contract errors.csv](https://github.com/blockstack/stacks-blockchain-api/files/6809127/Mainnet.-.stacks-node-api-tokens.-.token.contract.errors.csv)

docs/entities/tokens/non-fungible-token.schema.json
```diff
@@ -0,0 +1,36 @@
...
+  "$id": "non-fungible-token-metadata",
+  "title": "NonFungibleTokenMetadata",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "token_uri",
+    "name",
+    "description",
+    "image_uri",
+    "image_canonical_uri"
```
zone117x, line 12: @asimm241 can we also include the contract principal and txid that deployed the contract in both the NFT and FT payloads?

asimm241: sure

agraebe:
looks like the latest version has deployment errors. this whole PR will not be prioritized until the Rosetta updates are completed.

@zone117x will review the deployment errors

pstan26:
This PR not being merged is blocking MiamiCoin icon from being shown in the web wallet. Super important to have the brand consistent in the most used wallet rn.  link to the GH issue trying to get MIA icon included in web wallet here: https://github.com/blockstack/stacks-wallet-web/issues/1472

zone117x:
Status update for this PR:

When syncing from scratch, processing token metadata works pretty well for SIP-010 compliant FTs. 

NFTs aren't working so well -- the code needs updated to use [draft SIP-012](https://github.com/stacksgov/sips/pull/18) which defines NFT class metadata. Draft SIP-012 is still WIP, and after looking through various NFT contracts on mainnet, it looks like there's a few different ways authors have been defining metadata, examples:
```
https://explorer.stacks.co/txid/0x3f85e8d2a5453179438ed00e466e38b6c4acb3d2cefa579d20389b84602a12e7?chain=mainnet
(define-read-only (get-nft-meta)
  (ok {name: "Friedger Pool", uri: "https://pool.friedger.de/nft.png", mime-type: "image/png"}))
(define-read-only (get-token-uri (token-id uint))
  (ok (some "https://pool.friedger.de/nft.json")))

https://explorer.stacks.co/txid/0x7246cfb663ddd90041781919961a131938d34345879d708ed7d855039c70b84d?chain=mainnet
(define-read-only (get-nft-meta)
  (ok (some {name: "Clarity Developer OG", uri: "https://bafybeif4p2ukltj5eofwriclz4ru3p7izitprrs7a2rjhtp6qat673wagu.ipfs.dweb.link/", mime-type: "video/webm"})))

https://explorer.stacks.co/txid/0x5d1704291538dbeb201161f9f052f79281bea2a944f8c22e4ebdf67005d2b429?chain=mainnet
(define-read-only (get-nft-meta)
  (ok {name: "First Ever Hiro Hackathon Winner", uri: "https://ipfs.io/ipfs/QmXTeyFsiCGsB6st5FTbvc78RkXG8hKo5SxtCN8cv8ity3/hiro-first-hackathon.webm", mime-type: "video/webm",
        hash: "e4ce55dd89113b846c4d43f7018f66e57e7b12d6af960e2a143f91457d27e1f0"}))

https://explorer.stacks.co/txid/0xe865963227144aec75259f9c1f8e67b2d2b5764a29ba88a5794a700818659748?chain=mainnet
(define-read-only (get-nft-meta)
  {uri: "https://boom-nft-41369b66-36da-4442-be60-fff6d755b065.s3.amazonaws.com/24762181-0ba6-4c5b-9065-1c874fb334d2.svg", name: "Boomboxes", mime-type: "image/svg+xml"})

https://explorer.stacks.co/txid/0x10555a156fbd5931f2b33e843221ce73e3fd661127c64d840c3f8392dbd6887c?chain=mainnet
(define-read-only (get-nft-meta)
  (ok {name: "Friedger Pool", uri: "https://pool.friedger.de/nft.webp", mime-type: "image/webp"}))

https://explorer.stacks.co/txid/0x8f0393265a8a4102701ef6240b03022c567cf67f04e472e077709fe975424921?chain=mainnet
(define-read-only (get-nft-meta)
  (ok (some {name: "beeple", uri: "https://ipfsgateway.makersplace.com/ipfs/QmZ15eQX8FPjfrtdX3QYbrhZxJpbLpvDpsgb2p3VEH8Bqq", mime-type: "image/jpeg"})))
(define-read-only (get-token-uri (token-id uint))
  (ok (some "ipfs://ipfs/QmPAg1mjxcEQPPtqsLoEcauVedaeMH81WXDPvPx3VC5zUz")))

https://explorer.stacks.co/txid/0x8a8b1bd754664715c3280cb41fdf566c741789f200efa74681b79e4409e9800a?chain=mainnet
(define-read-only (get-nft-meta)
  (ok (some {name: "beeple", uri: "https://ipfsgateway.makersplace.com/ipfs/QmZ15eQX8FPjfrtdX3QYbrhZxJpbLpvDpsgb2p3VEH8Bqq", mime-type: "image/jpeg"})))


https://explorer.stacks.co/txid/0x5d8f968111be4556a13d8d0d25f39d998a749a982d673ae6ba7180ee7ccf1f62?chain=mainnet
(define-read-only (get-token-uri (id uint))
  (ok (some "https://cloudflare-ipfs.com/ipfs/bafkreidmfmsjiqunj4tkln57ozw2yhazufqv63j5zohhsqchwol6lznsfy")))
```

The NFT metadata processing code needs updated to handle the 2 or 3 common ways metadata is defined.


However, the bigger problem is that token metadata will _not_ resolve if the API is bootstrapped with an event-replay-log import, which is now the primary method used for our API deployments.

The solution is to "queue" detected token contract txs in postgres, and delay the process of resolving their metadata until after the API has full connectivity to the stacks-node.

I'm currently working on addressing both of the above issues.

CharlieC3:
Just coming up to speed here. After discussing with @zone117x, I've learned this PR introduces some new challenges and intricacies which will make the API much more difficult to maintain for future changes.

This may warrant a larger discussion, but it's becoming clear that the API is growing rapidly and taking on multiple roles beyond acting as an interface for the stacks-node and data on the blockchain. Ideally anything more complex than that would be extracted into its own dedicated microservice. Otherwise there are certain factors at play which would make the API a single point of failure, and lower the service's ability to be resilient.

wileyj:
@CharlieC3 
After looking through this PR, I want to add some concerns I have as this relates to the API itself. 
In my opinion, the API should remain a "simple" service that performs a "simple" task, but does it well. 
As we add more functionality and newer features, we add a lot more complexity and introduce functions that may impact the API performing it's tasks well. 

I would advocate using the so-called [microservice](https://en.wikipedia.org/wiki/Microservices) approach, where the API does what it does today (and does it well!), and as we need to add new features that aren't related to the **core** functionality of the API - they're added as HTTP requests to a separate service. 

for example, using @zone117x last comment, the endpoint `/extended/v1/tokens/ST3CK642B6119EVC6CT550PW5EZZ1AJW6608HK60A.citycoin-token/ft/metadata` here would now be an HTTP request made by the API to a new lightweight service, which would return the json to the API, which would then return data to the client. 

This would also solve some of the issues that were raised in that comment, the `event-replay-log import` function which is used today. 

I realize we're coming into this discussion late, but I do feel strongly about the microservice approach here - we should keep the API simple, and let it continue to do what it does well.

docs/entities/tokens/fungible-token.schema.example.json
```diff
@@ -0,0 +1,9 @@
+{
+  "token_uri": "https://heystack.xyz/token-metadata.json",
+  "name": "Heystack",
+  "description": "Heystack is a SIP-010-compliant fungible token on the Stacks Blockchain, used on the Heystack app",
+  "image_uri": "https://heystack.xyz/assets/Stacks128w.png",
+  "image_canonical_uri": "https://heystack.xyz/assets/Stacks128w.png",
+  "tx_id": "0xef2ac1126e16f46843228b1dk4830e19eb7599129e4jf392cab9e65ae83a45c0",
+  "sender_address": "ST399W7Z9WS0GMSNQGJGME5JAENKN56D65VGMGKGA"
```
rafaelcr, line 8: Add `symbol` and `decimals` here so it matches the schema

docs/api/tokens/get-fungible-tokens-metadata-list.example.schema.json
```diff
@@ -0,0 +1,16 @@
...
+  "total": 500,
+  "results": [
+    {
+      "token_uri": "https://heystack.xyz/token-metadata.json",
+      "name": "Heystack",
+      "description": "Heystack is a SIP-010-compliant fungible token on the Stacks Blockchain, used on the Heystack app",
+      "image_uri": "https://heystack.xyz/assets/Stacks128w.png",
+      "image_canonical_uri": "https://heystack.xyz/assets/Stacks128w.png",
+      "symbol": "stx",
+      "decimals": 5
```
rafaelcr, line 13: Add `tx_id` and `sender_address` examples here so it matches the schema

docs/entities/tokens/non-fungible-token.schema.example.json
```diff
@@ -0,0 +1,11 @@
+{
+  "token_uri": "https://pool.friedger.de/nft.json",
+  "name": "Friedger Pool",
+  "description": "Enjoying the stacking pool.",
+  "image_uri": "https://pool.friedger.de/nft.webp",
+  "image_canonical_uri": "https://pool.friedger.de/nft.webp",
+  "symbol": "stx",
+  "decimals": 5,
```
rafaelcr, line 8: `symbol` and `decimals` don't belong in this example

src/index.ts
```diff
@@ -131,6 +133,11 @@ async function init(): Promise<void> {
     monitorCoreRpcConnection().catch(error => {
       logger.error(`Error monitoring RPC connection: ${error}`, error);
     });
+
+    // TODO: register an exit-handler for the tokenProcessorQueue instance
+    const tokenMetadataProcessor = new TokensProcessorQueue(db, configuredChainID);
```
rafaelcr, line 138: You could use the newly added shutdown config handlers here for this TODO

rafaelcr (changes requested):

src/index.ts
zone117x: Good idea, done!

rafaelcr (approved):
LGTM 👍

zone117x:
Deployment testing has completed and this PR is ready to merge. I'm going to make one last minor change to make this feature opt-in for API deployments using and env var, so that entities running a private instance aren't unintentionally exposing themselves to various security concerns by fetching from user-defined URLs.

markmhx:
@zone117x Does this fully supportive NFT metadata in addition to FT metadata?

andresgalante:
@markmhx it doesn't support NFTs

markmhx:
What's with these commits then? 🤔 

image

andresgalante:
🤔 good catch

zone117x:
> @zone117x Does this fully support NFT metadata in addition to FT metadata?

No. NFT `class` metadata has some limited support, but will likely not be enabled in production deployment because it's not standardized. Individual NFT metadata (i.e. `CryptoKitty token num. 1395 metadata` vs `CryptoKitty metadata`) is not supported at all.

markmhx:
@zone117x mind creating an issue to capture the work that remains to flesh out NFT support? Is this next on your or @asimm241's docket, or not prioritized quite yet?

zone117x:
@markmhx see new issue https://github.com/blockstack/stacks-blockchain-api/issues/742 for completing NFT class metadata support -- however, the last I heard was that NFT class support wasn't very important, and that individual NFT token metadata support is what we want, which requires some discussion. I'm setting up a meeting with a few of the stakeholders. 

In the meantime, it would be helpful if UX could start a doc or gh issue/discussion to capture initial requirements?

markmhx:
Sounds good. On the UX side, we're looking to display individual names, individual images and class names for particular NFTs as seen here: https://github.com/blockstack/stacks-wallet-web/issues/1028#issuecomment-789822555

Please do indicate whether @jasperjansz needs to update these designs given limitations to what data can be made available.

unclemantis:
I have in .env the following:

STACKS_API_ENABLE_NFT_METADATA=1

When I connect to:

http://localhost:3999/extended/v1/tokens/nft/metadata

I get the following error:

  "error": "FT metadata processing is not enabled on this server"

What gives? Little help here. Thanks guys / gals!

PR #592, opened 2021-05-10
Fix zone file parsing 

### LINKED ISSUE #582: /v1/names/*/zonefile endpoint returning empty zone file
**Describe the bug**
The `/v1/names/*/zonefile` and `/v1/names/*/zonefile/*` API endpoints return an empty zone file at times. The [docs](https://blockstack.github.io/stacks-blockchain-api/#operation/fetch_zone_file) state:

>Fetch a user’s raw zone file. This only works for RFC-compliant zone files. This method returns an error for names that have non-standard zone files.

But I do not get an error. Instead, I get a HTTP 200 OK when I [query the endpoint](https://stacks-node-api.mainnet.stacks.co/v1/names/test.btc/zonefile/b07cad4ddd53c09342d6030154e16d7cad90fef5):

```JSON
{"zonefile":""}
```

The zone file is definitely on Atlas, as evidenced [here](https://stacks-node-api.mainnet.stacks.co/v2/attachments/b07cad4ddd53c09342d6030154e16d7cad90fef5):
```JSON
"244f524947494e20746573742e6274632e0a2454544c20333630300a0a746573742e6274632e09494e0941093136312e33352e3232382e36310a746573742e6274632e09494e0954585409224564697420746573742033220a"
```

When decoded:

```
$ORIGIN test.btc.
$TTL 3600

test.btc.	IN	A	161.35.228.61
test.btc.	IN	TXT	"Edit test 3"

```

Here is one that does work:

- [Zonefile endpoint](https://stacks-node-api.mainnet.stacks.co/v1/names/test.btc/zonefile/943f469cf471a9b15efc3bef06119a020b0d4542)
- [Attachment endpoint](https://stacks-node-api.mainnet.stacks.co/v2/attachments/943f469cf471a9b15efc3bef06119a020b0d4542)

**To Reproduce**
Not sure.

**Expected behavior**
For the endpoint to return the zone file.

**Additional context**

Is it something in Blockstack's [custom version of `zone-file`](https://github.com/blockstack/zone-file-js)?

https://github.com/blockstack/stacks-blockchain-api/blob/37bc66c6b5f9e8a687026810d9b32f2ae104787d/src/event-stream/event-server.ts#L66
https://github.com/blockstack/stacks-blockchain-api/blob/37bc66c6b5f9e8a687026810d9b32f2ae104787d/src/event-stream/event-server.ts#L471 

I use the original [`zone-file` package](https://www.npmjs.com/package/zone-file) and it parses the zone file above successfully:

```JS
> parseZoneFile(zone)
{
  '$origin': 'test.btc.',
  '$ttl': 3600,
  a: [ { name: 'test.btc.', ip: '161.35.228.61' } ],
  txt: [ { name: 'test.btc.', txt: 'Edit test 3' } ]
}
```

It is [impossible to write a clean lexer for DNS zone files](https://www.nlnetlabs.nl/documentation/nsd/grammar-for-dns-zone-files/), according to the author of NSD.

asimm241:
## Description
This PR fixes the issue described here #582 , It handles the zone files which are incorrect with missing Uri or subdomain properties.

Fixes #582 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review

### DIFF
--- a/src/event-stream/event-server.ts
+++ b/src/event-stream/event-server.ts
@@ -484,11 +484,9 @@ async function handleNewAttachmentMessage(msg: CoreNodeAttachmentMessage[], db: 
           // case for subdomain
           const subdomains: DbBnsSubdomain[] = [];
           for (let i = 0; i < zoneFileTxt.length; i++) {
-            if (!zoneFileContents.uri) {
-              throw new Error(`zone file contents missing URI: ${zonefile}`);
-            }
             const zoneFile = zoneFileTxt[i];
             const parsedTxt = parseZoneFileTxt(zoneFile.txt);
+            if (Object.entries(parsedTxt).find(e => e[1] === '') != undefined) continue; //if tx has any empty field , skip it
             const subdomain: DbBnsSubdomain = {
               name: name.concat('.', namespace),
               namespace_id: namespace,
@@ -504,7 +502,7 @@ async function handleNewAttachmentMessage(msg: CoreNodeAttachmentMessage[], db: 
               parent_zonefile_index: 0, //TODO need to figure out this field
               block_height: Number.parseInt(attachment.block_height, 10),
               zonefile_offset: 1,
-              resolver: parseResolver(zoneFileContents.uri),
+              resolver: zoneFileContents.uri ? parseResolver(zoneFileContents.uri) : '',
               atch_resolved: true,
             };
             subdomains.push(subdomain);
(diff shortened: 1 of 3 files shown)


src/event-stream/event-server.ts
```diff
@@ -484,11 +484,9 @@ async function handleNewAttachmentMessage(msg: CoreNodeAttachmentMessage[], db:
           // case for subdomain
           const subdomains: DbBnsSubdomain[] = [];
           for (let i = 0; i < zoneFileTxt.length; i++) {
-            if (!zoneFileContents.uri) {
-              throw new Error(`zone file contents missing URI: ${zonefile}`);
-            }
             const zoneFile = zoneFileTxt[i];
             const parsedTxt = parseZoneFileTxt(zoneFile.txt);
+            if (Object.entries(parsedTxt).find(e => e[1] === '') != undefined) continue; //if tx has any empty field , skip it
```
zone117x, line 489: Can this be a more explicit check, even if it ends up being more verbose? Hard to tell exactly what can trigger this condition.

PR #602, opened 2021-05-13
Feature/rosetta delegate stx

rllola:
Support for `delegate-stx` contract call.

Still need unit test.

### DIFF
--- a/src/api/routes/rosetta/construction.ts
+++ b/src/api/routes/rosetta/construction.ts
@@ -45,6 +45,9 @@ import {
   standardPrincipalCV,
   noneCV,
 } from '@stacks/transactions';
+import {
+  decodeBtcAddress,
+} from '@stacks/stacking';
 import * as express from 'express';
 import { StacksCoreRpcClient } from '../../../core-rpc/client';
 import { DataStore, DbBlock } from '../../../datastore/common';
@@ -190,12 +193,12 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
       case 'stacking': {
         // dummy transaction to calculate size
         const dummyPoxAddress = '1Xik14zRm29UsyS6DjhYg4iZeZqsDa8D3';
-        const { version, hash } = btcAddress.fromBase58Check(dummyPoxAddress);
-        const versionBuffer = bufferCV(new BN(version, 10).toBuffer());
-        const hashbytes = bufferCV(hash);
+        const { hashMode, data } = decodeBtcAddress(dummyPoxAddress);
+        const hashModeBuffer = bufferCV(new BN(hashMode, 10).toArrayLike(Buffer));
+        const hashbytes = bufferCV(data);
         const poxAddressCV = tupleCV({
           hashbytes,
-          version: versionBuffer,
+          version: hashModeBuffer,
         });
         if (!options.amount) {
           res.status(500).json(RosettaErrors[RosettaErrorsTypes.invalidOperation]);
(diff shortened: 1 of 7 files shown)


src/api/routes/rosetta/construction.ts
```diff
@@ -604,6 +643,60 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
...
+          res.status(500).json(RosettaErrors[RosettaErrorsTypes.invalidPublicKey]);
+          return;
+        }
+        const { version, hash } = btcAddress.fromBase58Check(poxBTCAddress);
+        const versionBuffer = bufferCV(new BN(version, 10).toBuffer());
+        const hashbytes = bufferCV(hash);
+        const poxAddressCV = tupleCV({
+          hashbytes,
+          version: versionBuffer,
+        });
```
zone117x, line 662: I'm not sure this works. Shouldn't the pox btc address be created with the `@stacks/stacking` lib which will convert the `version` byte?
cc @kyranjamie

kyranjamie: In the Stacks Wallet we don't let users define a BTC address, as this is managed by the pool. Would be a good idea to use the stacking lib if we can.

I'd also recommend these `case` statements match the `functionName` being called:

- `delegate-stx`
- `stack-stx`

zone117x: @rllola the above code for `stack-stx` is also broken. It also needs to convert the btc address to a pox address with the correct version byte. See how it's done in the `@stacks/stacking` lib: https://github.com/blockstack/stacks.js/blob/ce27b48c31b3101b0faf2360fa07655f0ec35fe1/packages/stacking/src/index.ts#L488-L511

rllola: @zone117x I don't see the difference between what is in the PR and what you linked ? The pox btc address in converted to a pox address.

rllola: Ah I think I got it. It is a the `decodeBtcAddress` function that should be used ? Instead of getting the btc address version

zone117x: The code here is incorrectly using the `version` byte from returned from `bitcoinjs.fromBase58Check`. The code in the `@stacks/stacking` lib converts the `version` byte to a Stacks address hash mode byte.

rllola: It should be fixed now.

rllola:
I am not quite sure what prettier is complaining about. VSC is complaining on this line https://github.com/Zondax/stacks-blockchain-api/blob/feature/rosetta-delegate-stx/src/tests-rosetta/api.ts#L69

src/tests-rosetta/api.ts
```diff
@@ -2132,6 +2132,52 @@ describe('Rosetta API', () => {
...
+    expect(JSON.parse(result.text)).toHaveProperty('metadata');
+    expect(JSON.parse(result.text)).toHaveProperty('suggested_fee');
+    expect(JSON.parse(result.text).metadata).toHaveProperty('contract_address');
+    expect(JSON.parse(result.text).metadata).toHaveProperty('contract_name');
+    expect(JSON.parse(result.text).metadata).toHaveProperty('burn_block_height');
+    expect(JSON.parse(result.text).metadata).toHaveProperty('delegate_to');
+    expect(JSON.parse(result.text).suggested_fee[0].value).toBe('260');
+  });
+
+
```
zone117x, line 2180: No, the lint issue is from this extra line. Just run `npm run lint:fix`

rllola: Fixed it. Sorry my VSC plugin was buggin.

zone117x (approved):

PR #606, opened 2021-05-18
chore: add initial overview.md

zone117x:
Project overview doc


kyranjamie (approved):
Which ascii charting tool did you use? I want to create fun, charts and diagrams too.

PR #611, opened 2021-05-23
Re-play emitted events

zone117x:
Implemented record, export, and import (playback) of data from a stacks-node's event stream. Solves the issue described in https://github.com/blockstack/stacks-blockchain/issues/2604 /cc @CharlieC3 

> Currently the stacks-node is only able to emit events live as they happen. This poses a problem in the scenario where the stacks-blockchain-api needs to be upgraded and its database cannot be migrated to a new schema.

> The only way to perform this upgrade is to wipe the stacks-blockchain-api's database. However, the stacks-blockchain-api is not able to request or query for the old events from previous blocks in this scenario, so the stacks-node also has to be restarted from an empty state in order to emit events from past blocks as it syncs with the network.

The feature can be used via program args. For example, if there are breaking changes in the API's sql schema, like adding a new column which requires event's to be re-played, the following steps could be ran:

1. Export event data to disk with the `export-events` command:
   ```
   node ./lib/index.js export-events --file /tmp/stacks-node-events.tsv
   ```
2. Update to the new stacks-blockchain-api version.
3. Perform the event playback using the `import-events` command:
   ```
   node ./lib/index.js import-events --file /tmp/stacks-node-events.tsv
   ```

Alternatively, instead of performing the `export-events` command in step 1, an environmental variable can be used to stream events to a file as they are received while the application is running. To enable this feature, set `STACKS_EXPORT_EVENTS_FILE` to the file path where events should be appended. Example:
```
STACKS_EXPORT_EVENTS_FILE=/tmp/stacks-events.tsv
```

The `import-events` process fully simulates a stacks-node event-emitter: it boots up the API's event observer http server on an ephemeral/internal port, then submits the event http POST requests exactly how they came in originally.

The process involves separate steps right now, but it could be adapted to be part of the regular automatic migration process.


src/event-stream/event-server.ts
```diff
@@ -844,5 +848,11 @@ export async function startEventServer(opts: {
...
   logger.info(`Event observer listening at: http://${addrStr}`);
 
-  return server;
+  const closeFn = async () => {
+    await new Promise<void>((resolve, reject) => {
+      logger.info('Closing event observer server...');
+      server.close(error => (error ? reject(error) : resolve()));
+    });
+  };
+  return Object.assign(server, { closeAsync: closeFn });
```
kyranjamie, line 857:
```suggestion
  return { ...server, ...{ closeAsync: closeFn } };
```
Avoids possible unwanted mutation of `server`.

CharlieC3:
1. I can see us using the export feature while the API is receiving events from the `stacks-node`. Can this somehow be configured to run the `export` on every new event written to the DB? (Incrementally would be ideal) That way it could help guarantee that the exported file is always up-to-date with what's in the DB, and avoid the scenario where we run the export in production, and another event comes through before we're able to stop the service.

2. I think the k8s pod could be reconfigured to first run the `import` when it comes up before starting the main API service, so that may be easier to implement in our infra.

3. Do you happen to know how long the `export` and `import` processes take from start to finish on mainnet or an equally-long chain?

src/event-stream/event-server.ts
zone117x: Normally would 100% agree, however, in this case copying the enumerable object properties from a very stateful and complex class instance like node's `http.Server` into a new object, would likely result in a very broken result object. An intentional mutation here is ok for this specific case. The best option is probably just returning `{ server: server, closeAsync: closeFn }`.

zone117x:
> Do you happen to know how long the export and import processes take from start to finish on mainnet or an equally-long chain?

Note that the length of the chain isn't the only factor -- the _uptime_ of the API instance has an impact on the number of event requests. For example, when syncing mainnet from scratch right now, the stacks-node will emit very few `/new_mempool_txs`, `/drop_mempool_tx`, and `/new_microblocks` events, because these are intended to be emitted when the network is live at chain tip. 

However, now that I'm thinking about it, the API is capable of rehydrating the entire correct db state with only `/new_block` and `/attachments/new` events. Something to keep in mind if we need to optimize this process later..


Here are the results on my machine with a recently fully synced mainnet db. 

Exporting:
```
$ time node lib/index.js export-events --file /tmp/events-export.tsv
Export event data to file: /tmp/events-export.tsv
Export started...
Export successful.
{"message":"Shutting down... received beforeExit.","level":"warn","timestamp":"2021-05-25T09:57:37.045Z"}
{"message":"App shutdown successful.","level":"info","timestamp":"2021-05-25T09:57:37.046Z"}

real	0m6.004s
user	0m2.240s
sys	0m0.596s

$ ls -lh /tmp/events-export.tsv
297M /tmp/events-export.tsv
```

Importing:
```
$ time node lib/index.js import-events --file /tmp/events-export.tsv --wipe-db
{"message":"Event observer listening at: http://127.0.0.1:50649","level":"info","timestamp":"2021-05-25T10:18:11.048Z"}
Importing raw event requests into temporary table...
Streaming raw event requests from temporary table...
<status updates>
Event import and playback successful.
{"message":"Shutting down... received beforeExit.","level":"warn","timestamp":"2021-05-25T11:55:19.042Z"}

real	7m10.821s
user	2m12.337s
sys	0m27.633s
```

zone117x:
> Can this somehow be configured to run the export on every new event written to the DB? (Incrementally would be ideal)

Done! Use this env var to enable the feature:
```
# Enable to have stacks-node events streamed to a file while the application is running
STACKS_EXPORT_EVENTS_FILE=/tmp/stacks-events.tsv
```

CharlieC3 (approved):
:shipit:

PR #612, opened 2021-05-24
Feat/microblocks

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/355

Implements API support for processing microblock data from the stacks-node event stream. Microblock are now stored as first-class objects in the db, and have some initial API endpoints to fetch them (`/extended/v1/microblock` -- similar to the `/extended/v1/block` endpoint). See https://stacks-blockchain-api-git-feat-microblocks-blockstack.vercel.app/#tag/Microblocks

Micro-forks are handled in the API in a similar way to anchor-block forks. A new `microblock_canonical` flag is used in the database similar to the anchor-block `canonical` flag.

Microblock data is also associated with each transaction, and returned in all existing tx API endpoints. These are new properties on the response objects, and _very_ close to being fully backwards compatible with the current API deployment.

Almost all the API endpoints now support an optional `?unanchored` query parameter. When left out, the responses do not include unconfirmed microblock data (transactions, address STX balances, etc).

The following properties on a transaction response object can be empty while the transaction's microblock is unconfirmed (waiting to be included in a real/anchor-block): `burn_block_time`, `burn_block_time_iso`, and `block_hash`. The `is_unanchored: boolean` property can be used to explicitly check if a tx object is unconfirmed and thus missing those properties. 

While a a microblock transaction is unconfirmed, clients are recommended to use/display the properties `microblock_hash`, `parent_block_hash`, and/or `parent_burn_block_time` as an alternative to the `block_hash` and `burn_block_time_iso`.

The transaction response schema with the new microblock properties can be seen here: https://stacks-blockchain-api-git-feat-microblocks-blockstack.vercel.app/#operation/get_transaction_by_id

The block response schemas also have new microblock related properties, for example an array of microblock hashes that were streamed off of the block and an array of microblock hashes that were confirmed by the block. See: https://stacks-blockchain-api-git-feat-microblocks-blockstack.vercel.app/#operation/get_block_by_hash

Additionally, this PR introduces a new account nonce endpoint. It provides a recommended nonce based off inspecting the mempool and microblock transactions. It also detects nonce gaps that tend to cause txs to get stuck in the mempool. See: https://stacks-blockchain-api-git-feat-microblocks-blockstack.vercel.app/#operation/get_account_nonces

# Testing
New microblocks-enabled API instances have been deploy:
* __mainnet__: https://stacks-node-api-microblocks.stacks.co/v2/info
* __testnet__: https://stacks-node-api-microblocks.testnet.stacks.co/v2/info
* __regest__: https://stacks-node-api-microblocks.regtest.stacks.co/v2/info

Note: mainnet miners haven't accepted microblocks since block height 7534, this is something the blockchain-core team is looking into.

# Deployment
~This must be ran against a stacks-node version that has a handful of microblock changes to the event emitter, see issue https://github.com/blockstack/stacks-blockchain/issues/2668 and PR https://github.com/blockstack/stacks-blockchain/pull/2653. The image `stacks-blockchain:fix-microblock-events` has these changes.~

This branch should be ran with stacks-blockchain 2.0.11.2.0-rc1.

# TODO - blockers
* [x] All endpoints should require an `?unanchored=true` param in order to include unconfirmed microblock data. This opt-in is necessary to prevent possibly breaking integrations that expect the `block_hash` property to exist in transaction response objects.
  * Done: https://github.com/blockstack/stacks-blockchain-api/pull/629
* [x] Include `parent_burn_block_{hash, time, height}` once available from core, see https://github.com/blockstack/stacks-blockchain/issues/2662.
  * Done: commit https://github.com/blockstack/stacks-blockchain-api/pull/629/commits/0953af2a1c359cfb58637be2c05d82640a948a79 in PR https://github.com/blockstack/stacks-blockchain-api/pull/629
# TODO - non-blockers
* [ ] Add microblocks and unconfirmed microblock-tx emitting to the [`DataStoreEventEmitter`](https://github.com/blockstack/stacks-blockchain-api/blob/cba848585f2d372345aca9a35ae984b5dd7ab6d3/src/datastore/common.ts#L284-L292) along with equivalent new [socket.io subscription topic](https://github.com/blockstack/stacks-blockchain-api/blob/ba2e4ed86cfea1eed3d6a4ab47b1c55c10fa4de5/src/api/routes/socket-io.ts#L16) for microblocks.
* [ ] `tx_index` is likely no longer a logical ordering of transaction is it was intended, see https://github.com/blockstack/stacks-blockchain/issues/2668#issuecomment-847010900 -- likely needs fixed in core then API can be updated to use it.
* [ ] Robust test suite for exercising various micro-forking scenarios.
  * Some tests implemented in https://github.com/blockstack/stacks-blockchain-api/pull/629, but not robust or exhaustive enough. Should probably create a new gh issue for this.
* [x] Update OpenAPI / json schemas with helpful descriptions/summaries for new microblock related properties, also update the json schema examples to include new microblock related properties.
  * Done: https://github.com/blockstack/stacks-blockchain-api/pull/612/commits/11d4a4fc8503e8d9007af70f7f4fc00a7f62c9f5


markmhx:
I've added @fbwoolf as an (optional) reviewer as well since I imagine she'll be interested in studying these changes.

docs/entities/blocks/block.schema.json
```diff
@@ -3,7 +3,7 @@
   "title": "Block",
   "description": "A block",
   "type": "object",
-  "required": ["canonical", "height", "hash", "parent_block_hash", "txs", "burn_block_time", "burn_block_time_iso", "burn_block_hash", "burn_block_height", "miner_txid"],
+  "required": ["canonical", "height", "hash", "parent_block_hash", "txs", "burn_block_time", "burn_block_time_iso", "burn_block_hash", "burn_block_height", "miner_txid", "parent_microblock_hash", "parent_microblock_sequence", "microblocks_accepted", "microblocks_streamed"],
```
kyranjamie, line 6: Could we run prettier on the schema json too? Think this will auto-format this on new lines.

kyranjamie (approved):
Great stuff Matt, especially appreciate the schema refactor  👏🏼 

Will this PR handle emitting microblocks via the socket? Couldn't see any code for this.

I don't have much to comment on implementation itself. I've opened an issue in stacks wallet to test out with these changes incase the missing properties cause issue.

docs/entities/blocks/block.schema.json
zone117x: Yeah we should. I also have a related couple TODOs in my notes:
* Print a warning in the doc type gen for schema objects that are missing a `title`.
* Print a warning in the doc type gen for schema objects that are missing `additionalProperties: false`

zone117x:
> Will this PR handle emitting microblocks via the socket? Couldn't see any code for this.

Not at the moment -- unconfirmed microblocks and microblock-txs don't get emitted (and so don't reach the socket.io server handler). Good catch, it's a relatively simple thing to add.

fbwoolf (approved):
Really helpful for me to review, lots of great work. 👍

zone117x:
There appears to be a regression with genesis block data missing from postgres on the mainnet microblocks deployment. For example, the bns contract deployment appears to be missing:
```
GET https://stacks-node-api-microblocks.stacks.co/extended/v1/contract/SP000000000000000000002Q6VF78.bns
{"error":"cannot find contract by ID SP000000000000000000002Q6VF78.bns"}
```
Causing the API to error when creating tx responses that are contract-calls to genesis contracts, for example:
```
GET https://stacks-node-api-microblocks.stacks.co/extended/v1/tx/0x20b8b99551174aa01e82d1e1ded3573d0a759f5b5cd019e654ca51a0c873b592
{"error":"Error: Failed to lookup smart contract by ID SP000000000000000000002Q6VF78.bns","stack":"Error: Failed to lookup smart contract by ID SP000000000000000000002Q6VF78.bns\n    at getTxFromDataStore (/app/src/api/controllers/db-controller.ts:722:13)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at Object.searchTx (/app/src/api/controllers/db-controller.ts:773:19)\n    at /app/src/api/routes/tx.ts:197:21\n    at wrappedMiddleware (/app/node_modules/@awaitjs/express/index.js:62:7)","errorTag":"7af6554a-57c3-4b9d-a30a-21ffc29a6661"}
```

I'm investigating.

kyranjamie:
@blockstack/ux-team will need a microblocks-ready types package to help prep for the launch, guess we can publish a beta version from this PR?

zone117x:
> need a microblocks-ready types package to help prep for the launch, guess we can publish a beta version from this PR?

Yep can do. We'll need another one for the `?unachored=true` changes here https://github.com/blockstack/stacks-blockchain-api/issues/625 since that adds a param to just about every endpoint, probably later this week. Then another beta with these changes https://github.com/blockstack/stacks-blockchain-api/issues/626 that add a new prop to every tx response object. I'll publish a beta off this current PR today 👍

src/api/routes/microblock.ts
```diff
@@ -0,0 +1,70 @@
...
+} from '../controllers/db-controller';
+import { has0xPrefix } from '../../helpers';
+import { parseLimitQuery, parsePagingQueryInput } from '../pagination';
+
+const MAX_MICROBLOCKS_PER_REQUEST = 200;
+
+const parseMicroblockQueryLimit = parseLimitQuery({
+  maxItems: MAX_MICROBLOCKS_PER_REQUEST,
+  errorMsg: '`limit` must be equal to or less than ' + MAX_MICROBLOCKS_PER_REQUEST,
+});
```
kyranjamie, line 22: Great to see my beloved `parseLimitQuery` still being used

src/event-stream/reader.ts
```diff
@@ -182,110 +186,135 @@ export function createTransactionFromCoreBtcTxEvent(
...
+      raw_tx: txBuffer,
+      parsed_tx: rawTx,
+      block_hash: blockData.block_hash,
+      index_block_hash: blockData.index_block_hash,
+      parent_index_block_hash: blockData.parent_index_block_hash,
+      parent_block_hash: blockData.parent_block_hash,
+      block_height: blockData.block_height,
+      burn_block_time: blockData.burn_block_time,
+      microblock_sequence: coreTx.microblock_sequence ?? I32_MAX,
+      microblock_hash: coreTx.microblock_hash ?? '',
```
kyranjamie, line 275: Might it be better to use `null` here?

zone117x: It would definitely be better, but that will probably have to be a new PR. Each small type or property change in a common object like this requires hundreds of lines of code changes, and a ton of time fixing unit/integration tests 🙁

zone117x:
@blockstack/ux-team I think this PR can be considered feature complete (for blockers at least) and so far all _known_ bugs have been fixed. The transaction response property `parent_burn_block_time`/`parent_burn_block_time_iso` were just implemented but not yet deployed (looking at mid week for next deployment update). 

I've listed the TODOs that I think are non-blockers, I think we could make issues for those and merge. 

Thoughts?

docs/entities/blocks/block.schema.json
zone117x: Issue created https://github.com/blockstack/stacks-blockchain-api/issues/657

docs/api/address/get-address-transactions-with-transfers.example.json
```diff
@@ -18,6 +18,10 @@
         "burn_block_time": 1613009951,
         "burn_block_time_iso": "2021-02-11T02:19:11.000Z",
         "canonical": true,
+        "is_unanchored": "false",
```
kyranjamie, line 21:
```suggestion
        "is_unanchored": false,
```

kyranjamie: These should be native bools right?

CharlieC3:
@zone117x Does this PR have any breaking changes in it? If so we should make sure that's specified in the merge commit's footer to `develop` following conventional commits to ensure the MAJOR version is incremented when it's eventually merged to `master`.
https://www.conventionalcommits.org/en/v1.0.0/#summary

zone117x:
> Does this PR have any breaking changes in it? 

It shouldn't

CharlieC3:
> > Does this PR have any breaking changes in it?
> 
> It shouldn't, for clients at least. For sql schema yes, but are we tracking those in conventional commits?

That might be something Splat as a whole should discuss and determine if it's right to increment the MAJOR version when the schema has a breaking change in it. At the very least the MAJOR version should increment for the API if there's a breaking change that affects compatibility between the stacks-node and the API.

PR #616, opened 2021-05-28
fix: bundling issues

kyranjamie:
This is a bit of an odd fix, but we need to either:

- Export the types as a regular typescript file, not a declaration
- Trick module bundlers into thinking it's a legit module

One of the later TS versions introduces the inability to import `.d.ts` files. Declaration files are generated to describe the contents of compiled output, and thus shouldn't be imported.

The contents of this package may only be typings, but it describes our API, not the contents of a compiled script. 

See https://github.com/octokit/types.ts for another example of a type-only package.

### DIFF
--- a/docs/index.js
+++ b/docs/index.js
@@ -1,0 +1,1 @@
+// empty file to satisfy module bundlers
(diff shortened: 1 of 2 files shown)


zone117x (approved):
Thanks @kyranjamie!

PR #620, opened 2021-06-01
Fix the latest flag issue for bns tables during re-org

### LINKED ISSUE #614: BNS names "disappear" during re-orgs
When a name is registered or updated, it has a boolean flag `latest` set to `true`. The purpose of this flag is to allow fast enumeration of lists of names given various filtering criteria. Without the `latest` flag, then endpoints like `/v1/namespaces/{tld}/names` would have to perform an inefficient recursive JOIN to identify the latest status for each possible name it could return. The `namespaces`, `names`, and `subdomains` tables all use this `latest` flag. 

However, there's a bug in the re-org handling where `latest` is never updated during forks. This leads to situations where the _actual_ latest canonical name is flagged as `latest=false`, and all queries filter it out. 

This just happened with the name `aulneau.stx`. See the SQL output below showing re-org contention at block 16415. The row with `id=1803` eventually "won", according to the stacks-node, as being part of the canonical block. While row `id=1802` has been orphaned. The out-of-date `latest` flag is also visible. So now requests to `/v1/names/aulneau.stx` and similar endpoints will return empty.

```json
[
  {
    "id": 1803,
    "name": "aulneau.stx",
    "address": "SP3N4FPBG1PXMKEW0Y4K2MCSWDXHZP42YV5YW3AY4",
    "registered_at": 16415,
    "zonefile_hash": "f1df7f6ea05a3e6de952bb393429a5048acc61ca",
    "zonefile": "",
    "namespace_id": "stx",
    "latest": true,
    "status": "name-register",
    "canonical": false,
    "atch_resolved": false
  },
  {
    "id": 1802,
    "name": "aulneau.stx",
    "address": "SP3N4FPBG1PXMKEW0Y4K2MCSWDXHZP42YV5YW3AY4",
    "registered_at": 16415,
    "zonefile_hash": "f1df7f6ea05a3e6de952bb393429a5048acc61ca",
    "zonefile": "$ORIGIN aulneau.stx\n$TTL 3600\n_http._tcp\tIN\tURI\t10\t1\t\"https://gaia.blockstack.org/hub/1QEin7SMRVDopWnXoL7dqob94cCNzeNm5M/profile.json",
    "namespace_id": "stx",
    "latest": false,
    "status": "name-register",
    "canonical": true,
    "atch_resolved": true
  }
]
```

The fix is straight forward:
1. When any of the BNS table rows are marked as `canonical=true` or `canonical=false` during re-org, a second pass must be made to re-identify the the latest names, and set the `latest` flag as well. That code should happen around here: https://github.com/blockstack/stacks-blockchain-api/blob/cba848585f2d372345aca9a35ae984b5dd7ab6d3/src/datastore/postgres-store.ts#L818-L858
For example:
```sql
UPDATE names 
SET canonical = $2
WHERE index_block_hash = $1 AND canonical != $2
RETURNING name
```
Then for each returned `name`, run a query that identifies the last `canonical=true` name with the highest block_height and set it to `latest=true`, and set others to `latest=false`. This should be possible with a single query per name using a CTE.

2. Looking at the above SQL results for `aulneau.stx`, it's clear that `zonefile` content should be stored as relational data in a separate table. The BNS related queries can do a simple JOIN against that table using the `zonefile_hash` to retrieve the content.

asimm241:
## Description
This PR fixes the issue of disappearing names during reorgs. During Reorgs, reiterate on BNS tables and set `latest` to true for the latest entry as per block height. 

fixes #614 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [ ] Tag @zone117x for review

### DIFF
--- a/src/api/routes/rosetta/network.ts
+++ b/src/api/routes/rosetta/network.ts
@@ -95,7 +95,9 @@ export function createRosettaNetworkRouter(db: DataStore, chainId: ChainID): Rou
 
     const refUrl = ReferenceNodes[getRosettaNetworkName(chainId)];
     if (refUrl !== undefined) {
-      const stacksReferenceCoreRpcClient = new StacksCoreRpcClient(refUrl);
+      //rosetta tests are failing due to this
+      // const stacksReferenceCoreRpcClient = new StacksCoreRpcClient(refUrl);
+      const stacksReferenceCoreRpcClient = new StacksCoreRpcClient();
       const referenceNodeInfo = await stacksReferenceCoreRpcClient.getInfo();
 
       const referenceNodeTipHeight = referenceNodeInfo.stacks_tip_height;
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -834,16 +834,17 @@ export class PgDataStore
           WHERE name = $1 AND canonical = true
           ORDER BY registered_at DESC LIMIT 1
         ), 
-        WITH non_canonical_latest AS (
+          non_canonical_latest AS (
           SELECT id
           FROM names
           WHERE name = $1 AND canonical = false AND latest = true
           ORDER BY registered_at DESC LIMIT 1
         )
         UPDATE names 
-        SET latest = CASE WHEN id = latest_name.id THEN true
-                          WHEN id =  non_canonical_latest.id THEN false
+        SET latest = CASE WHEN id in ( SELECT id FROM latest_name ) THEN true
+                          WHEN id =  ( SELECT  id FROM non_canonical_latest ) THEN false
                           ELSE latest
+                     END
         WHERE name = $1`,
         [name]
       );
@@ -873,16 +874,17 @@ export class PgDataStore
           WHERE namespace_id = $1 AND canonical = true
           ORDER BY ready_block DESC LIMIT 1
         ), 
-        WITH non_canonical_latest AS (
+          non_canonical_latest AS (
           SELECT id
           FROM namespaces
           WHERE namespace_id = $1 AND canonical = false AND latest = true
           ORDER BY ready_block DESC LIMIT 1
         )
         UPDATE namespaces 
-        SET latest = CASE WHEN id = latest_namespace.id THEN true
-                          WHEN id =  non_canonical_latest.id THEN false
+        SET latest = CASE WHEN id in ( SELECT id FROM latest_namespace ) THEN true
+                          WHEN id in ( SELECT id FROM non_canonical_latest ) THEN false
                           ELSE latest
+                     END
         WHERE namespace_id = $1`,
         [namespace]
       );
@@ -913,7 +915,7 @@ export class PgDataStore
           WHERE fully_qualified_subdomain = $1 AND canonical = true
           ORDER BY block_height DESC LIMIT 1
         ),
-        WITH non_canonical_latest AS (
+          non_canonical_latest AS (
           SELECT id
           FROM subdomains
           WHERE fully_qualified_subdomain = $1 AND canonical = false AND latest = true
@@ -921,9 +923,10 @@ export class PgDataStore
         )
 
         UPDATE subdomains 
-        SET latest = CASE WHEN id = latest_subdomain.id THEN true
-                          WHEN id =  non_canonical_latest.id THEN false
+        SET latest = CASE WHEN id in ( SELECT id FROM latest_subdomain )  THEN true
+                          WHEN id in ( SELECT id FROM  non_canonical_latest ) THEN false
                           ELSE latest
+                     END
         WHERE fully_qualified_subdomain = $1`,
         [subdomain]
       );
@@ -3679,7 +3682,7 @@ export class PgDataStore
       index_block_hash,
       atch_resolved,
     } = bnsName;
-    await client.query(`UPDATE names SET latest = $1 WHERE name= $2`, [false, name]);
+    if (canonical) await client.query(`UPDATE names SET latest = $1 WHERE name= $2`, [false, name]);
 
     await client.query(
       `
@@ -3699,7 +3702,7 @@ export class PgDataStore
         hexToBuffer(tx_id ? tx_id : '0x'),
         status,
         canonical,
-        index_block_hash,
+        hexToBuffer(index_block_hash ? index_block_hash : '0x'),
         atch_resolved,
       ]
     );
@@ -3724,10 +3727,11 @@ export class PgDataStore
       canonical,
       index_block_hash,
     } = bnsNamespace;
-    await client.query(`UPDATE namespaces SET latest = $1 WHERE namespace_id= $2`, [
-      false,
-      namespace_id,
-    ]);
+    if (canonical)
+      await client.query(`UPDATE namespaces SET latest = $1 WHERE namespace_id= $2`, [
+        false,
+        namespace_id,
+      ]);
 
     await client.query(
       `
@@ -3753,7 +3757,7 @@ export class PgDataStore
         latest,
         hexToBuffer(tx_id ? tx_id : '0x'),
         canonical,
-        index_block_hash,
+        hexToBuffer(index_block_hash ? index_block_hash : '0x'),
       ]
     );
   }
@@ -3840,6 +3844,7 @@ export class PgDataStore
         found: true,
         result: {
           ...queryResult.rows[0],
+          index_block_hash: bufferToHexPrefixString(queryResult.rows[0].index_block_hash),
           tx_id: bufferToHexPrefixString(queryResult.rows[0].tx_id),
         },
       };
@@ -3865,6 +3870,7 @@ export class PgDataStore
         found: true,
         result: {
           ...queryResult.rows[0],
+          index_block_hash: bufferToHexPrefixString(queryResult.rows[0].index_block_hash),
           tx_id: bufferToHexPrefixString(queryResult.rows[0].tx_id),
         },
       };
@@ -4019,6 +4025,7 @@ export class PgDataStore
         found: true,
         result: {
           ...queryResult.rows[0],
+          index_block_hash: bufferToHexPrefixString(queryResult.rows[0].index_block_hash),
           tx_id: bufferToHexPrefixString(queryResult.rows[0].tx_id),
         },
       };
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -824,7 +824,8 @@ export class PgDataStore
       `,
       [indexBlockHash, canonical]
     );
-    nameResult.rows.forEach(async name => {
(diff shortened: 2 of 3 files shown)


src/datastore/postgres-store.ts
```diff
@@ -820,9 +820,29 @@ export class PgDataStore
       UPDATE names
       SET canonical = $2
       WHERE index_block_hash = $1 AND canonical != $2
+      RETURNING name
       `,
       [indexBlockHash, canonical]
     );
+    nameResult.rows.forEach(async name => {
```
zone117x, line 827: `Array.forEach` is going to execute all these promises in parallel, then exit the function and leave these as dangling promises, with the `client` object instance being disposed off while in use. 
These need switched to a regular `for` loop with `await`s within the loop body.

src/datastore/postgres-store.ts
```diff
@@ -820,9 +820,29 @@ export class PgDataStore
...
+          WHERE name = $1 AND canonical = true
+          ORDER BY registered_at DESC LIMIT 1
+        )
+        UPDATE names 
+        SET latest = false
+        WHERE name = $1 AND canonical =true
+        
+        UPDATE names 
+        SET latest = true 
+        WHERE id = latest_name.id`,
```
zone117x, line 842: This looks like sql code accidentally left in?

src/datastore/postgres-store.ts
```diff
@@ -820,9 +820,29 @@ export class PgDataStore
...
+        `
+        WITH latest_name AS (
+          SELECT id
+          FROM names
+          WHERE name = $1 AND canonical = true
+          ORDER BY registered_at DESC LIMIT 1
+        )
+        UPDATE names 
+        SET latest = false
+        WHERE name = $1 AND canonical =true
```
zone117x, line 838: The `latest_name` CTE looks good for finding the latest name, but the UPDATE expression doesn't seem right. 
Also, it looks like the multiple UPDATE expressions here are intentional -- multiple sql statements in a single query should never be used.

It looks like what you might want to do instead is add another CTE like `non_canonical_latest` that identifies row (should only be one) that is now non-canoncal but still latest, like `WHERE canonical = false AND latest = true`, then change the `UPDATE names SET latest = <...>` expression to be `true` if the row matches the `latest_name` CTE, and `false` if the row matches the `non_canonical_latest` CTE.

src/datastore/postgres-store.ts
```diff
@@ -848,9 +890,30 @@ export class PgDataStore
...
+        UPDATE subdomains 
+        SET latest = false
+        WHERE fully_qualified_subdomain = $1 AND canonical =true
+
+        UPDATE subdomains 
+        SET latest = true 
+        WHERE id = latest_subdomain.id`,
+        [subdomain]
+      );
+    });
```
zone117x, line 916: Same comments as above for these `subdomains` and `namespaces` queries.

src/datastore/postgres-store.ts
asimm241: No! The previous query set all latest to false, this one set only the latest to be true.

asimm241:
@zone117x I want to have a test case for this behaviour, is this scenerio producible in test cases?

zone117x:
> I want to have a test case for this behaviour, is this scenerio producible in test cases?

Yep, you'd write a test simulating the re-org behavior like one of these:
https://github.com/blockstack/stacks-blockchain-api/blob/0565d4f7f3130ed418b4652be3ae3c2a548bffe2/src/tests/datastore-tests.ts#L2488

To simulate the forking behavior described in https://github.com/blockstack/stacks-blockchain-api/issues/614, you'd do something like 
1. Insert block `1A` (height 1).
2. Insert child block `2A` (height 2, parent block should be `1A`), with a `name` registration included in this block.
3. Test to ensure that the `name` is `latest=true` and `canonical=true`.
4. Insert block alternative child block `2B` (height 2, parent block should be `1A`), with the same name registration. The re-org handling will treat this block as `non-canonical` and no re-orgs will happen, block `2A` will still be considered canonical.
5. Insert block `3B` (height 3, parent block should be `2B`). The re-org handling will kick in and mark `2A` as non-canonical, and mark `2B` as canonical.
6. Test that the `name` registration from `2A` is now marked `canonical=false` and `latest=false`, and the `name` registration from `2B` is now marked `canonical=true` and `latest=true`.

The block forks look like:
```
1A -> 2A
  \_> 2B -> 3B
```

src/datastore/postgres-store.ts
```diff
@@ -3602,7 +3682,7 @@ export class PgDataStore
       index_block_hash,
       atch_resolved,
     } = bnsName;
-    await client.query(`UPDATE names SET latest = $1 WHERE name= $2`, [false, name]);
+    if (canonical) await client.query(`UPDATE names SET latest = $1 WHERE name= $2`, [false, name]);
```
asimm241, line 3685: @zone117x found this bug as well. For example, if we receive a non-canonical name, It was immediately being set out to latest false and canonical true.

src/datastore/postgres-store.ts
```diff
@@ -3647,10 +3727,11 @@ export class PgDataStore
       canonical,
       index_block_hash,
     } = bnsNamespace;
-    await client.query(`UPDATE namespaces SET latest = $1 WHERE namespace_id= $2`, [
-      false,
-      namespace_id,
-    ]);
+    if (canonical)
```
asimm241, line 3730: @zone117x same as above!

src/api/routes/rosetta/network.ts
```diff
@@ -95,7 +95,9 @@ export function createRosettaNetworkRouter(db: DataStore, chainId: ChainID): Rou
 
     const refUrl = ReferenceNodes[getRosettaNetworkName(chainId)];
```
asimm241, line 96: @raynaudoe  Rosetta tests were failing due to this, because in tests `refUrl` resolves to `testnet.stacks.co`.  
cc: @zone117x .

zone117x:
Thanks for all the fixes and new re-org unit tests @asimm241! 

So, while implementing https://github.com/blockstack/stacks-blockchain-api/issues/625, we ended up with another problem related to the state `latest: boolean` flag on BNS data. Unconfirmed microblock txs result in the `latest` flag updating to true, which then makes in impossible to fetch "confirmed-only" BNS data in a way that leverages the `latest` flag. 

In other words, to facilitate the default behavior of excluding unconfirmed BNS data, we have to solve the original problem of inefficient BNS sql queries, but _without_ the ability to use `latest` flag approach -- as explained in the referenced issue for this PR:
> When a name is registered or updated, it has a boolean flag `latest` set to `true`. The purpose of this flag is to allow fast enumeration of lists of names given various filtering criteria. Without the `latest` flag, then endpoints like `/v1/namespaces/{tld}/names` would have to perform an inefficient recursive JOIN to identify the latest status for each possible name it could return.

After some trial and error and lots of benchmarking we ended up with a stateless solution that doesn't use the `latest` flag, but still has results in fast sql queries. The gist is that we use a specialized compound + sorted index that looks like:
```js
  pgm.createIndex('subdomains', [
    { name: 'fully_qualified_subdomain' },
    { name: 'canonical', sort: 'DESC' },
    { name: 'microblock_canonical', sort: 'DESC' },
  ]);
```
...in conjunction with a postgres-specific SQL extension `DISTINCT ON(...)`, like:
```sql
  SELECT DISTINCT ON (fully_qualified_subdomain) fully_qualified_subdomain
  FROM subdomains
  WHERE block_height <= $2
  WHERE canonical = true AND microblock_canonical = true
  ORDER BY fully_qualified_subdomain, block_height DESC, tx_index DESC
  LIMIT 100
  OFFSET $1
```

With the current production dataset, the initial naive approach took ~10 seconds to look up a list of latest subdomains, with this new approach the query takes ~50ms.

My feeling is that we should take the re-org tests from this PR and fold it into the PR for https://github.com/blockstack/stacks-blockchain-api/issues/625. Alternatively, we can merge and deploy this PR as-is which would fix the "disappearing BNS data" bug a bit quicker. I'd prefer to avoid that latter option because it will introduce some painful merge conflicts, and I think the microblocks deployment is won't be too far behind anyway.

@diwakergupta @agraebe @markmhx thoughts?

asimm241:
> Thanks for all the fixes and new re-org unit tests @asimm241!
> 
> So, while implementing #625, we ended up with another problem related to the `latest: boolean` flag on BNS data. Unconfirmed microblock txs result in the `latest` flag updating to true, which then makes in impossible to fetch "confirmed-only" BNS data in a way that leverages the `latest` flag.
> 
> In other words, to facilitate the default behavior of excluding unconfirmed BNS data, we have to solve the original problem of inefficient BNS sql queries, but _without_ the ability to use `latest` flag approach -- as explained in the referenced issue for this PR:
> 
> > When a name is registered or updated, it has a boolean flag `latest` set to `true`. The purpose of this flag is to allow fast enumeration of lists of names given various filtering criteria. Without the `latest` flag, then endpoints like `/v1/namespaces/{tld}/names` would have to perform an inefficient recursive JOIN to identify the latest status for each possible name it could return.
> 
> After some trial and error and lots of benchmarking we ended up with a stateless solution that doesn't use the `latest` flag, but still has results in fast sql queries. The gist is that we use a specialized compound + sorted index that looks like:
> 
> ```js
>   pgm.createIndex('subdomains', [
>     { name: 'fully_qualified_subdomain' },
>     { name: 'canonical', sort: 'DESC' },
>     { name: 'microblock_canonical', sort: 'DESC' },
>     { name: 'block_height', sort: 'DESC' },
>     { name: 'tx_index', sort: 'DESC' },
>   ]);
> ```
> 
> ...in conjunction with a postgres-specific SQL extension `DISTINCT ON(...)`, like:
> 
> ```sql
>   SELECT DISTINCT ON (fully_qualified_subdomain) fully_qualified_subdomain
>   FROM subdomains
>   WHERE block_height <= $2
>   WHERE canonical = true AND microblock_canonical = true
>   ORDER BY fully_qualified_subdomain, block_height DESC, tx_index DESC
>   LIMIT 100
>   OFFSET $1
> ```
> 
> With the current production dataset, the initial naive approach took ~10 seconds to look up a list of latest subdomains, with this new approach the query takes ~50ms.
> 
> My feeling is that we should take the re-org tests from this PR and fold it into the PR for #625. Alternatively, we can merge and deploy this PR as-is which would fix the "disappearing BNS data" bug a bit quicker. I'd prefer to avoid that latter option because it will introduce some painful merge conflicts, and I think the microblocks deployment won't be too far behind anyway.
> 
> @diwakergupta @agraebe @markmhx thoughts?

It makes sense to wait for #625 and used tests from this PR

agraebe:
@asimm241 to close and create an issue to follow up

PR #627, opened 2021-06-09
Feat/running api docs

wileyj:
This PR adds 2 new documents with the goal of making it *easier* for someone to run an API. 
2 guides are added - the "easy" method using only docker, and a method on installing/running from source. 

Both guides deal with all the dependencies (stacks-blockchain, postgres) and address any incidental requirements (firewalls, links to API docs, etc). 

https://github.com/blockstack/docs/pull/1192/files

is dependent on this PR, since we'll be linking to these files from the docs site once the above is merged.


wileyj:
@pgray-hiro once this is merged, the docs updates should point to the `master` branch vs the current links to the feature branch.

zone117x (approved):
Awesome, thanks @wileyj!

PR #655, opened 2021-07-14
Add PG_SSL and PG_CONNECTION_URI

### LINKED ISSUE #653: REQUEST/QUESTION: Connect to api db via ssl
I would like to connect from the event server to the event db via ssl.  I do not see a way to configure this in any of the available examples/configs.

Maybe an env variable `SSL=true` that would append `?ssl=true` to the connection string 

Turning off db ssl is one option, but seems like a bad one.

**Additional context**
Although my connections are on a private network, I still want the communication via ssl.  This is not for a public endpoint.

whoabuddy:
## Description

This PR fixes #653 by adding an option to enable SSL connections to the PG DB via the `.env` file.

For the `PG_CONNECTION_URI` option - should that supersede the other settings? I'm thinking the best way to handle that would be to add something to the `../helpers.ts` file that checks if the value is included, then returns the config as either the URI string or the full `ClientConfig` object.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

It shouldn't - the default `.env` file includes the default value of `false` however anyone running a custom `.env` file will need to provide the `PG_SSL` value.

The docs point to `stacks-local-dev` for running an API server - so I don't think so?

## Testing information

Yes, @dantrevino may be able to help based on his use case?

## Checklist
- [ ] `npm run test` passes - I'm having some trouble getting this to run, it errors out with:
```
FetchError: request to http://127.0.0.1:20443/v2/info failed, reason: connect ECONNREFUSED 127.0.0.1:20443
    at ClientRequest.<anonymous> (/home/whoabuddy/Dev/stacks-blockchain-api/node_modules/node-fetch/lib/index.js:1461:11)
    at ClientRequest.emit (events.js:315:20)
    at Socket.socketErrorListener (_http_client.js:469:9)
    at Socket.emit (events.js:315:20)
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
```
Do I just need to run the docker image for stacks-blockchain to resolve that?
- [ ] Changelog is updated - not yet, what is the best way to determine next version number?
- [x] Tag @zone117x for review

Sorry this one is a bit messy - it's one thing to know which lines to change, but a whole different procedure to know all the bits needed to make the PR successful! Any feedback is appreciated.


whoabuddy:
Looks like my prettier config made way more changes than I anticipated in the markdown files, and not sure why test coverage dropped by such a large percentage? :shrug:

src/datastore/postgres-store.ts
```diff
@@ -72,6 +72,7 @@ export function getPgClientConfig(): ClientConfig {
     password: process.env['PG_PASSWORD'],
     host: process.env['PG_HOST'],
     port: parsePort(process.env['PG_PORT']),
+    ssl: process.env['PG_SSL'],
```
zone117x, line 75: I'm assuming the `pg` lib accepts a falsy value here, as in this is the equivalent of `ssl: process.env['PG_SSL'] ?? false`?

zone117x:
Awesome, thanks @whoabuddy for getting this started!

> For the `PG_CONNECTION_URI` option - should that supersede the other settings? I'm thinking the best way to handle that would be to add something to the `../helpers.ts` file that checks if the value is included, then returns the config as either the URI string or the full `ClientConfig` object.

In theory the datastore is supposed to be agnostic (a database other than PG could be implemented), so I think this code should stay within the `/src/datastore/postgres-store.ts` file.

I think the appropriate thing to do here is allow _only_ either `PG_CONNECTION_URI` _or_ the other `PG_*` env vars. If both are detected, immediately through a runtime error during init.

src/datastore/postgres-store.ts
zone117x: Actually, I think this needs to have bit more logic to handle the string value `"false"` and/or `"0"` because technically those are truthy, non-empty strings.

zone117x:
> not sure why test coverage dropped by such a large percentage? 🤷

Codecov reporting and its config is super unintuitive to me. Seems like every PR has an inaccurate coverage diff 🤷

whoabuddy:
Got it, I'll have to make some more time to really dive deep on this and it might not happen for the next two weeks with the focus on CityCoins. I was hoping it would be a quick change - appreciate the quick response and will update here if anything changes.

zone117x:
@whoabuddy I pushed a fix for the PG_SSL var parsing and support for PG_CONNECTION_URI. This needs rebased and could be merged.

whoabuddy:
@zone117x Looks like it's all fixed up for the changes we added, but the tests are still throwing errors.

There was a value `PG_SCHEMA` that's used below and is not recognized by the `pg` library.

https://github.com/whoabuddy/stacks-blockchain-api/blob/upd/db-connections/src/datastore/postgres-store.ts#L184-L193

I could comment out line 192 and change it, assuming it will work without it:

```ts
// [clientConfig.schema, clientConfig.database]
[clientConfig.database]
```

But without knowing where/why this is in there, I don't want to cause an unnecessary regression. Thoughts?

zone117x:
`PG_SCHEMA` is also needed by `node-pg-migrate`, see here https://github.com/whoabuddy/stacks-blockchain-api/blob/2ccf9fc2c6ef8011e95efffedf965c15b29b1017/src/datastore/postgres-store.ts#L152

It looks like `schema` can be specified in the connection uri with a query param, e.g. `?schema=public`, although I think it needs to be manually parsed out of the URL string. I can push a fix for that, but looks like this PR got mangled during a merge. The commits should just be rebased or cherrypicked on top of the latest develop branch.

whoabuddy:
Agreed, will submit a new PR with just the relevant changes and go from there!

whoabuddy:
Closing in favor of #666 :smiling_imp:

PR #665, opened 2021-07-16
Singular tx with stx transfers endpoint

MHassanTariq:
## Description

This feature includes fetching information regarding a single transaction based on their tx_id. It adds a new API that can respond with transaction info such that tx_id and principle are given to it. The test case for this functionality has also been merged with the previously added test case for transaction with transfers. 

```
const principle = 'ST000000000000000000002AMW42H';
const tx_id = '0xcba511741b230bd85cb5b3b10d26e0b92695d4a83f95c260cad82a40cd764235';

const api = `/extended/v1/address/${principle}/${tx_id}/transactions_with_transfers/`;
```
For more details, refer to the following issue
https://github.com/blockstack/stacks-blockchain-api/issues/622

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

This is a separate feature that doesn't have any impact on the previously developed features.

## Testing information
Upon calling the API with valid principle and tx_id, it should return the information for that particular transaction. If either of the two fields is incorrect, an error will be returned.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


src/datastore/postgres-store.ts
```diff
@@ -4345,6 +4345,114 @@ export class PgDataStore
...
+          count: number;
+          event_index?: number;
+          event_type?: number;
+          event_amount?: string;
+          event_sender?: string;
+          event_recipient?: string;
+        }
+      >(
+        `
+      SELECT
```
zone117x, line 4368: Various new microblock related properties need to be included in this query, see https://github.com/blockstack/stacks-blockchain-api/blob/ab45a151f4136a1e8907befc1d6047035f6fd6f8/src/datastore/postgres-store.ts#L4380-L4420

MHassanTariq: Changes have been made.
I have added microblock_canonical = true into the query, where required.

docs/openapi.yaml
```diff
@@ -1013,6 +1013,26 @@ paths:
...
+          required: true
+          schema: 
+            type: string
+        - name: tx_id
+          in: path
+          description: Transaction id
+          required: true
+          schema:
+            type: string
+
```
asimm241, line 1035: add a response for this api

asimm241 (changes requested):

src/datastore/postgres-store.ts
```diff
@@ -4345,6 +4345,114 @@ export class PgDataStore
...
+            recipient: r.event_recipient,
+          });
+          if (r.event_sender === stxAddress) {
+            txResult.stx_sent += eventAmount;
+          }
+          if (r.event_recipient === stxAddress) {
+            txResult.stx_received += eventAmount;
+          }
+        }
+      }
```
zone117x, line 4450: There's a lot of code here, should it be shared with the `getAddressTxsWithStxTransfers` code? It could be moved into a `parseTxsWithStxTransfers` function.

docs/api/address/get-address-single-transaction-with-transfers.schema.json
```diff
@@ -0,0 +1,190 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "description": "The root schema comprises the entire JSON document.",
+  "title": "The root schema",
```
zone117x, line 4: Boilerplate description and title should be filled in.

docs/api/address/get-address-single-transaction-with-transfers.schema.json
```diff
@@ -0,0 +1,190 @@
...
+        "hex": {
+          "type": "string"
+        },
+        "repr": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": true
+    },
+    "post_conditions": {
```
asimm241, line 98: @MHassanTariq you can use this as a `$ref ` https://github.com/blockstack/stacks-blockchain-api/blob/master/docs/entities/post-conditions/post-condition.schema.json

asimm241 (changes requested):

src/datastore/postgres-store.ts
MHassanTariq: Another function is created and updated into both functions.

docs/api/address/get-address-single-transaction-with-transfers.schema.json
MHassanTariq: I have used a predefined schema. Updated openapi.yaml with reference to that schema.

asimm241 (approved):

src/api/routes/address.ts
```diff
@@ -185,6 +186,38 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
     res.json(response);
   });
 
+  router.getAsync('/:stx_address/:tx_id/transactions_with_transfers', async (req, res) => {
```
zone117x, line 189: Since this is for a singular transaction, should this be renamed to `transaction_with_transfers`, or even just `transfers` because it's already in the context of a transaction?

MHassanTariq: I have changed the endpoint to `with_transfers` as suggested in the comment below the respective issue:
https://github.com/blockstack/stacks-blockchain-api/issues/622#issuecomment-857064925

zone117x (approved):
LGTM!

Issue #621, opened 2021-06-01
Create separate table for zone file content


asimm241:
The `zonefile` content should be stored as relational data in a separate table. The BNS related queries can do a simple JOIN against that table using the zonefile_hash to retrieve the content.

See here: #614

MHassanTariq:
The following functions contain queries that require change for implementing this flow:
- resolveBnsNames
- updateNames
- getName
- getHistoricalZoneFile
- getLatestZoneFile
- updateBatchSubdomains
- getSubdomain

PR #669, opened 2021-07-19
New table for zonefiles

MHassanTariq:
## Description

Due to the large size of zonefiles, it is appropriate for them to be allocated in another table, making the queries for names and subdomains faster. This PR makes another table for zonefiles including zonefiles themselves and their hashes for their search. It solves the problem for issue #621. 

The working of previous code remains the same, keeping in mind, only the queries and database structure is altered.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

It will make the lookup for subdomains and users a lot faster.

## Testing information
It can be tested using the already written test cases.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


src/datastore/postgres-store.ts
```diff
@@ -1514,20 +1514,42 @@ export class PgDataStore
...
+        WHERE tx_id = $2 AND canonical = true AND microblock_canonical = true
         `,
-        [zonefile, atch_resolved, hexToBuffer(tx_id)]
+        [atch_resolved, hexToBuffer(tx_id)]
       );
     });
     this.emit('nameUpdate', tx_id);
   }
 
+  private validateZonefileHash(zonefileHash: string) {
```
MHassanTariq, line 1546: Created this function to defined a standard for zonefile hash such that it starts with `0x`.

src/datastore/postgres-store.ts
```diff
@@ -1514,20 +1514,42 @@ export class PgDataStore
     });
   }
 
-  async resolveBnsNames(zonefile: string, atch_resolved: boolean, tx_id: string): Promise<void> {
+  async resolveBnsNames(
```
asimm241, line 1517: Why do we need `atch_resolved` now?

src/datastore/postgres-store.ts
asimm241: use `has0xPrefix` from `helpers.ts`

asimm241 (changes requested):

rafaelcr (approved):

rafaelcr (approved):
LGTM

src/datastore/postgres-store.ts
```diff
@@ -5135,44 +5174,51 @@ export class PgDataStore
...
-        atch_resolved,
-        hexToBuffer(blockData.index_block_hash),
-        hexToBuffer(blockData.parent_index_block_hash),
-        hexToBuffer(blockData.microblock_hash),
-        blockData.microblock_sequence,
-        blockData.microblock_canonical,
-      ]
-    );
+    // inserting remianing names information in names table
+    await this.queryTx(async client => {
```
zone117x, line 5186: This should use the existing `client` passed to the function rather than opening a new sql tx.

src/datastore/postgres-store.ts
```diff
@@ -3644,11 +3664,30 @@ export class PgDataStore
...
+      name: insertZonefileQueryName,
+      text: zonefileInsertQuery,
+      values: zonefileValues,
+    };
     try {
-      const res = await client.query(insertBnsSubdomainsEventQuery);
-      if (res.rowCount !== subdomains.length) {
-        throw new Error(`Expected ${subdomains.length} inserts, got ${res.rowCount}`);
-      }
+      const result = await this.queryTx(async client => {
```
zone117x, line 3676: This should use the existing `client` passed to the function rather than opening a new sql tx.

src/tests-bns/bns-integration-tests.ts
```diff
@@ -611,7 +609,7 @@ describe('BNS integration tests', () => {
       status: 'name-register',
       zonefile:
         '$ORIGIN zumrai.id\n$TTL 3600\n_http._tcp	IN	URI	10	1	"https://gaia.blockstack.org/hub/1EPno1VcdGx89ukN2we4iVpnFtkHzw8i5d/profile.json"\n\n',
-      zonefile_hash: '853cd126478237bc7392e65091f7ffa5a1556a33',
+      zonefile_hash: '0x853cd126478237bc7392e65091f7ffa5a1556a33',
```
zone117x, line 612: This should be reverted to use the original formatting (no `0x` prefix) as this is a `/v1` endpoint intended to be backwards compatible with the same endpoints from Stacks 1.0.

src/datastore/postgres-store.ts
```diff
@@ -5587,6 +5633,7 @@ export class PgDataStore
           UNION ALL (
             SELECT DISTINCT ON (fully_qualified_subdomain) fully_qualified_subdomain as name, zonefile
             FROM subdomains
+            LEFT JOIN zonefiles ON zonefiles.zonefile_hash = subdomains.zonefile_hash
```
zone117x, line 5636: Can you test how this change impacts the execution time of this query on the full BNS dataset? It's not clear to me if postgres will continue to use the sql index that makes this query fast enough. Related to the comment here: https://github.com/blockstack/stacks-blockchain-api/pull/730#discussion_r700394859

cc @asimm241 who might be able to help

src/tests-bns/bns-integration-tests.ts
MHassanTariq: The problem is that [handleBlockMessage](https://github.com/blockstack/stacks-blockchain-api/blob/master/src/event-stream/event-server.ts#L223) returns zonefile_hash without `0x`. 
handleBlockMessage

The [handleNewAttachmentMessage](https://github.com/blockstack/stacks-blockchain-api/blob/master/src/event-stream/event-server.ts#L557) returns zonefile_hash with `0x`. 
handleNewAttachmentMessage


So it stores the information without `0x` in [updateNames](https://github.com/blockstack/stacks-blockchain-api/blob/master/src/datastore/postgres-store.ts#L5296-L5300) and updates it for zonefile with `0x` in [resolveBnsNames](https://github.com/blockstack/stacks-blockchain-api/blob/master/src/datastore/postgres-store.ts#L1602-L1604).

Initially, we were searching with tx_id, as the same table was used to store `zonefiles` and `tx_ids`. Now, we have separate tables for both. 

For now, what we can do is drop `0x` from the `handleNewAttachmentMessage` to search for the same `zonefile_hash` as inserted. What do you think about that?

zone117x: > For now, what we can do is drop 0x from the handleNewAttachmentMessage to search for the same zonefile_hash as inserted. What do you think about that?

That sounds fine. Normally we decode hex strings into buffers (`bytea` in postgres) because it's about 50% smaller in size. But my primary concern is ensuring these endpoints stay backwards compatible, regardless of how they are stored in the db.

src/datastore/postgres-store.ts
MHassanTariq: With previous query (for same table for zonefiles and names):
Planning time: 0.860 ms
Execution time: 0.436 ms
Total time: 1.296 ms

With this query:
Planning Time: 10.892 ms
Execution Time: 1.321 ms
Total Time: 12.213 ms

@zone117x

zone117x: Thanks for the update @MHassanTariq. Can you confirm the query was done on the full BNS subdomain dataset, which should be something like 2 million rows? Also, were able to test the other modified query (inside `getSubdomain`)?

zone117x: Also, while the total time of both previous and new query appear low, the relative difference between the 2 suggests the JOIN is not particularly optimal. Ideally we be seeing a new total time of: `original query time, e.g. 1.3ms` + `query time to lookup a zonefile by hash (should be faster than first query)`.

Apologies if this seems nitpicky, but pg query performance is something we're becoming more stringent about.

MHassanTariq: @zone117x regarding the `getSubdomain` query, it was taking total time for execution of `18.539ms` and after dividing names and zonefiles into two tables, it is now taking a total time of `15.040ms`.

Your exact suggested solution was proposed by @asimm241 too. I analyzed with that and got a total time of `3.759 ms` for `getLatestZonefile`, I think I should change my approach to the new one. 
1. Getting the zonefile_hash from the first query gives the total time of `3.366ms`
2. Fetching zonefile from zonefile_hash takes `0.393ms`.

Yep, all times were calculated with the full BNS subdomain dataset having more than 2 million rows.

src/migrations/1627387313410_migrate-old-zonefiles.ts
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable @typescript-eslint/camelcase */
+import { MigrationBuilder, ColumnDefinitions } from 'node-pg-migrate';
+
+export const shorthands: ColumnDefinitions | undefined = undefined;
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+    await pgm.db.query(`
```
zone117x, line 7: I appreciate the proper sql migration here! However, I don't think we actually want a migration for this particular dataset given the number of zonefiles. Or, if we do merge with this migration, we'd probably want to "prune" it next time we making a sql-schema-breaking change. @rafaelcr thoughts?

zone117x (approved):
Looks good! Left one last comment around the sql migration of zonefiles.

src/migrations/1627387313410_migrate-old-zonefiles.ts
rafaelcr: This PR would require a re-sync of all events to the DB, correct? If so, couldn't we just remove this migration and also remove the `zonefile` columns from the original `names` table migration?

zone117x: Ahh yes, @rafaelcr is right. At first I didn't notice the additional changes to previous existing migration files. Those are breaking sql changes, and there is no need for this extra migration file.

rafaelcr (changes requested):
Just remove the extra migration step from `names` to the new `zonefiles` table along with the `zonefile` column on `names`.

rafaelcr (approved):
👍

PR #679, opened 2021-07-28
[Rosetta][Construction-API] fix constructing stacking tx using /preprocess

### LINKED ISSUE #674: [Rosetta][Construction-API][bug] constructing stacking tx using /preprocess not offline
when sending a stacking operation for constructing a transaction using /preprocess using the OFFLINE stacks API, it errors with
```
"error": "FetchError: request to http://127.0.0.1:20443/v2/contracts/interface/ST000000000000000000002AMW42H/pox failed, reason: connect ECONNREFUSED 127.0.0.1:20443" 
```
meaning that this is not offline as its trying to connect to the node.

/preprocess endpoint needs to be entirely offline, ie. cannot connect to the node.

I suspect it might be due to when creating a dummy stacking transaction: https://github.com/blockstack/stacks-blockchain-api/blob/master/src/api/routes/rosetta/construction.ts#L205-L214

asimm241:
## Description
This PR fixes the bug for rosetta `construction/preprocess`. The PR was making an api request to the node where it was supposed to work offline. 

fixes #674 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

_NO_
_NO_

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x]  @zone117x for review

### DIFF
--- a/src/api/routes/rosetta/construction.ts
+++ b/src/api/routes/rosetta/construction.ts
@@ -181,7 +181,6 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
           fee: new BN(0),
           // placeholder public key
           publicKey: '000000000000000000000000000000000000000000000000000000000000000000',
-          network: getStacksNetwork(),
           // We don't know the non yet but need a placeholder
           nonce: new BN(0),
         };
@@ -208,8 +207,7 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
           functionName: 'stack-stx',
           publicKey: '000000000000000000000000000000000000000000000000000000000000000000',
           functionArgs: [uintCV(options.amount), poxAddressCV, uintCV(0), uintCV(0)],
-          validateWithAbi: true,
-          network: getStacksNetwork(),
+          validateWithAbi: false,
         };
         transaction = await makeUnsignedContractCall(dummyStackingTx);
         break;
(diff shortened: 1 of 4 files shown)


src/api/routes/rosetta/construction.ts
```diff
@@ -231,8 +229,7 @@ export function createRosettaConstructionRouter(db: DataStore, chainId: ChainID)
             noneCV(),
             noneCV(),
           ],
-          validateWithAbi: true,
-          network: getStacksNetwork(),
+          validateWithAbi: false,
```
zone117x, line 232: Think we need to add these to the offline-rosetta tests? Like I don't know off the top of my head of if `makeUnsignedContractCall` works without network when the `fee`/`nonce` are unspecified

asimm241: You're right,  It won't work I double-checked and It doesn't work without fee and network. 
So, I think I should provide a hardcoded fee here as well. 

Thanks for pointing that out.

zone117x: Fix looks good. We can merge if you've manually tested and verified the fix, but we should try to get this in the offline rosetta tests at some point.

asimm241: I have manually tested it with offline mode.
We do have a test for `/preprocess`, I don't know how to test this behaviour though. one idea that comes to mind is that we can disallow all network calls with `nock` but I don't know if it will work or not.

asimm241: @zone117x It works!. checkout the new commit!

zone117x (approved):
Thanks @asimm241!

PR #689, opened 2021-08-03
[Rosetta] support stacking, delegate stacking and revoke stacking event

asimm241:
## Description

This PR is adding to rosetta support for stacking, delegate stacking and revoke stacking events.

This PR is created from https://github.com/blockstack/stacks-blockchain-api/pull/613

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

## Testing information
Tested using rosetta-cli

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x]@zone117x for review


src/tests-rosetta/api.ts
```diff
@@ -2206,7 +2206,7 @@ describe('Rosetta API', () => {
   // TODO: fails with:
   //  Response 500: Internal Server Error fetching http://127.0.0.1:20443/v2/pox - Failed to query peer info
   //  https://github.com/blockstack/stacks-blockchain/issues/2600
-  test.skip('construction/metadata - delegate-stacking', async () => {
+  test.skip('construction/metadata - delegate_stacking', async () => {
```
zone117x, line 2209: @asimm241 looks like this might have been fixed by https://github.com/blockstack/stacks-blockchain/pull/2634, can we try not skipping it for additional test coverage?

src/datastore/postgres-store.ts
```diff
@@ -5583,6 +5578,54 @@ export class PgDataStore
...
+      SELECT locked_amount, unlock_height, block_height, tx_id, locked_address
+      FROM stx_lock_events
+      WHERE canonical = true AND unlock_height <= $1 AND unlock_height > $2
+      `,
+      [current_burn_height, previous_burn_height]
+    );
+
+    const result: StxUnlockEvent[] = [];
+    lockQuery.rows.forEach(row => {
+      const unlockEvent: StxUnlockEvent = {
```
zone117x, line 5617: @asimm241 the code coverage report shows this path is never hit, along with pretty much all of the other Stacking `unlock` related code in `rosetta-helpers.ts` and `db-controller.ts`. Would you be able to write tests that cover these paths? I'm concerned about bugs around all these new functions.

src/rosetta-helpers.ts
```diff
@@ -446,6 +531,193 @@ export function getOptionsFromOperations(operations: RosettaOperation[]): Rosett
   return options;
 }
 
+function parseStackingContractCall(
```
zone117x, line 534: The code coverage is also showing that none of the conditions in this function are ever ran

src/datastore/postgres-store.ts
zone117x: These have test code coverage now, great job!

zone117x (approved):
Thanks @asimm241!
I've created an issue to track the remaining areas that need test coverage: https://github.com/blockstack/stacks-blockchain-api/issues/692

In the meantime, let's go ahead and merge this one.

zone117x:
@asimm241 I pushed an update to the rosetta stacking test so that it covers some of the tx operation parsing functions like `parseStackingContractCall`, `parseStackStxArgs`, `makeCallContractOperation`, `getOptionsFromOperations`, etc.

We still need tests for delegate-stacking and revoke-stacking https://github.com/blockstack/stacks-blockchain-api/issues/692

src/api/rosetta-constants.ts
```diff
@@ -40,7 +39,11 @@ export const RosettaOperationTypes = [
   'mint',
   'burn',
   'miner_reward',
-  'stx_lock',
+  'stx_lock', // stx event
+  'stx_unlock', // forged event
+  'stack_stx', // PoX contract function
+  'delegate_stx', // PoX contract function
+  'revoke_delegate_stx', // PoX contract function
```
zone117x, line 46: I don't see where some of these constants are used anywhere. For example I see `delegate-stx` used as the rosetta operation type here:
https://github.com/blockstack/stacks-blockchain-api/blob/d2af183033873e3957ed8277485465b00651af3f/src/rosetta-helpers.ts#L550
but not `delegate_stx` anywhere. 

Similar for `stack_stx  and `revoke_delegate_stx`.

Seems like these should all be defined as a enum or string literal type union, then used in all the areas where Rosetta operation types are checked or set.

zone117x: Looks like this was due to bad refactors which caused a mismatch between the operation types reported here and type actually being used. Should be fixed now in https://github.com/blockstack/stacks-blockchain-api/pull/689/commits/0d54de8b9daca11efbadb80b3df601db99029f5e

Issue #681, opened 2021-07-28
Fast graceful shutdown


zone117x:
When shutting down the handlers wait for all client connections to be closed. This causes issues with socket.io/websocket connections that can take hours to close on their own. Especially if the shutdown was caused by an unhandled exception. A critical error can happen, be logged, then not surface any problems for an extended period of time.

The shutdown handlers should give client connections a few seconds to close then force close them if necessary.

CharlieC3:
I think I have this mentioned in another ticket somewhere, but shutdowns should also result in the main process dying. Currently when the API shuts down, the ports are closed but the main process never dies. If the main process dies, K8s can restart the pod and the API would recover in a short amount of time.

zone117x:
Process should die from https://github.com/blockstack/stacks-blockchain-api/blob/d112c4c9e5d0ee3393a217b22429080c688fd7bb/src/shutdown-handler.ts#L26-L29
AFAIK there's no escaping [`process.exit`](https://nodejs.org/api/process.html#process_process_exit_code).

My guess is the code is never reached because there's always a registered handler (like websocket client waits) that have maybe never completed so far.

PR #707, opened 2021-08-12
graceful shutdown of stuck connections

rafaelcr:
## Description

Internal connections may take a long time to shut down, causing the API to hang and not report any errors until the process is complete. To avoid that, this PR sets a time limit for connection close handlers so we can run some logic when one of them gets stuck. Solves issue #681 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

No

## Testing information

Testing is done by sending a SIGINT to the nodejs process so shutdown sequences are triggered.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -25,7 +25,7 @@ import { createRosettaMempoolRouter } from './routes/rosetta/mempool';
 import { createRosettaBlockRouter } from './routes/rosetta/block';
 import { createRosettaAccountRouter } from './routes/rosetta/account';
 import { createRosettaConstructionRouter } from './routes/rosetta/construction';
-import { isProdEnv, logError, logger, LogLevel } from '../helpers';
+import { isProdEnv, logError, logger, LogLevel, waiter } from '../helpers';
 import { createWsRpcRouter } from './routes/ws-rpc';
 import { createSocketIORouter } from './routes/socket-io';
 import { createBurnchainRouter } from './routes/burnchain';
@@ -323,35 +323,14 @@ export async function startApiServer(opts: {
 
   const forceKill = async () => {
     logger.info('Force closing API server...');
-    await Promise.allSettled([
-      new Promise<void>((resolve, reject) => {
-        io.close(error => {
-          if (error) {
-            reject(error);
-          } else {
-            resolve();
-          }
-        });
-      }),
-      new Promise<void>((resolve, reject) => {
-        wss.close(error => {
-          if (error) {
-            reject(error);
-          } else {
-            resolve();
-          }
-        });
-      }),
-      new Promise<void>(resolve => {
-        for (const socket of serverSockets) {
-          socket.destroy();
-        }
-        resolve();
-      }),
-      new Promise<void>(resolve => {
-        server.close(() => resolve());
-      }),
-    ]);
+    const [ioClosePromise, wssClosePromise, serverClosePromise] = [waiter(), waiter(), waiter()];
+    io.close(() => ioClosePromise.finish());
+    wss.close(() => wssClosePromise.finish());
+    server.close(() => serverClosePromise.finish());
+    for (const socket of serverSockets) {
+      socket.destroy();
+    }
+    await Promise.allSettled([ioClosePromise, wssClosePromise, serverClosePromise]);
   };
 
   const addr = server.address();
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -49,6 +49,7 @@ export interface ApiServer {
   address: string;
   datastore: DataStore;
   terminate: () => Promise<void>;
+  forceKill: () => Promise<void>;
 }
 
 export async function startApiServer(opts: {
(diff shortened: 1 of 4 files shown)


src/shutdown-handler.ts
```diff
@@ -13,13 +18,23 @@ async function startShutdown() {
...
-      await Promise.resolve(handler());
+      logger.info(`Closing ${config.name}...`);
+      await resolveOrTimeout(Promise.resolve(config.handler()), 5000, timeoutError);
+      logger.info(`${config.name} closed`);
     } catch (error) {
       errorEncountered = true;
-      logError('Error running shutdown handler', error);
+      if (error === timeoutError) {
+        logError(`Closing of ${config.name} timed out`);
+        if (config.timeoutHandler) {
```
zone117x, line 32: Maybe the `else` for this condition (`if (config.timeoutHandler === null) { ... }`) should be something like:

```ts
logError(`Critical process has taken longer than ${timeoutConstant} to shutdown, possibly hanging indefinitely`);
```

src/shutdown-handler.ts
```diff
@@ -1,10 +1,15 @@
-import { logError, logger } from './helpers';
+import { logError, logger, resolveOrTimeout } from './helpers';
 
 const SHUTDOWN_SIGNALS = ['SIGINT', 'SIGTERM'] as const;
 
 export type ShutdownHandler = () => void | PromiseLike<void>;
+export type ShutdownConfig = {
+  name: string;
+  handler: ShutdownHandler;
+  timeoutHandler: ShutdownHandler | null;
```
zone117x, line 9: Would be helpful if this was more clear on what the difference between a `null` value and a function. Does a null value mean the handler shouldn't be time-constricted (because it's a critical routine), or vice versa?

Maybe either a tsdoc/jsdoc comment on the property, or an additional explicit property?

src/index.ts
```diff
@@ -116,7 +116,13 @@ async function init(): Promise<void> {
       datastore: db,
       chainId: configuredChainID,
     });
-    registerShutdownHandler(() => eventServer.closeAsync());
+    registerShutdownConfig({
+      name: 'Event Server',
```
zone117x, line 120: (for notes) this one is critical and can't be force closed

src/index.ts
```diff
@@ -135,14 +141,22 @@ async function init(): Promise<void> {
...
+      // TODO: Handle
+    },
   });
 
-  registerShutdownHandler(async () => {
-    logger.info('Closing DB...');
-    await db.close();
-    logger.info('DB closed.');
+  registerShutdownConfig({
+    name: 'DB',
```
zone117x, line 155: (for notes) this one is critical and can't be force closed

src/index.ts
```diff
@@ -135,14 +141,22 @@ async function init(): Promise<void> {
 
   const apiServer = await startApiServer({ datastore: db, chainId: getConfiguredChainID() });
   logger.info(`API server listening on: http://${apiServer.address}`);
-  registerShutdownHandler(async () => {
-    await apiServer.terminate();
+  registerShutdownConfig({
+    name: 'API Server',
```
zone117x, line 145: (for notes) this one and the other handlers that I didn't comment on are _not_ critical and can be force closed

zone117x (commented):
Left some comments that should be easy fixes, but this is looking great!
Btw, if you rebase on top of latest develop then that flaky `rosetta` test should stop failing in github CI.

kyranjamie (approved):

src/api/init.ts
```diff
@@ -320,6 +321,39 @@ export async function startApiServer(opts: {
     });
   };
 
+  const forceKill = async () => {
+    logger.info('Force closing API server...');
+    await Promise.allSettled([
```
zone117x, line 326: `Promise.allSettled` seems useful here. But, I think in order to prevent race conditions between the `server.close()` function the the `socket.destroy()` function, these need to be performed in a specific order, synchronously, and _then_ the callbacks for those could be used to "complete" a promise, and awaited inside `Promise.allSettled`.

For example, there's a function `waiter()` in `helper.ts` that returns a `Waiter<T>` object useful for these scenarios.

It could be used to do something like:
```ts
const [ioClosePromise, wssClosePromise, serverClosePromise] = [waiter(), waiter(), waiter()];
// Synchronously call `close` on the NodeJS http server which prevents _new_ connections.
// TODO: preserve the error logging in these close callbacks..
io.close(() => ioClosePromise.finish());
wss.close(() => wssClosePromise.finish());
server.close(() => serverClosePromise.finish());

// Destroy all the existing connections.
for (const socket of serverSockets) {
  socket.destroy();
}

await Promise.allSettled([ioClosePromise, wssClosePromise, serverClosePromise]);
```

zone117x (approved):
LGTM, thanks @rafaelcr!

PR #710, opened 2021-08-17
return tx_id on every asset in /extended/v1/address/[:addr]/assets

### LINKED ISSUE #678: Include txid in account asset transfers response
This endpoint: https://blockstack.github.io/stacks-blockchain-api/#operation/get_account_assets, https://stacks-node-api.regtest.stacks.co/extended/v1/address/ST1X6M947Z7E58CNE0H8YJVJTVKS9VW0PHEG3NHN3/assets can be used to construct an activity view of transfers of non-stx assets (fts, nfts). Currently however, the entries in the response do not include the original txid for the event. This makes it impossible to construct a rich view of the events in clients like the wallet or explorer.

rafaelcr:
## Description

Returns the transaction ID along with every asset queried on a particular address under the  `/extended/v1/address/[:addr]/assets` endpoint. This will be useful for wallets and the explorer so they can display additional information about the purchase or transfer of said assets.

Closes #678 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

Yes, the documentation for this particular endpoint should be updated.

## Testing information

Covered by current API and Datastore unit tests (ammended in this PR).

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/docs/entities/transaction-events/transaction-event-fungible-asset.schema.json
+++ b/docs/entities/transaction-events/transaction-event-fungible-asset.schema.json
@@ -14,6 +14,9 @@
           "type": "string",
           "enum": ["fungible_token_asset"]
         },
+        "tx_id": {
+          "type": "string"
+        },
         "asset": {
           "type": "object",
           "additionalProperties": false,
--- a/docs/entities/transaction-events/transaction-event-fungible-asset.schema.json
+++ b/docs/entities/transaction-events/transaction-event-fungible-asset.schema.json
@@ -8,7 +8,7 @@
     },
     {
       "type": "object",
-      "required": ["event_type", "asset"],
+      "required": ["event_type", "tx_id", "asset"],
       "properties": {
         "event_type": {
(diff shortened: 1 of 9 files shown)


zone117x (commented):
Can you add the new property to the OpenAPI spec? The routes are defined here: https://github.com/blockstack/stacks-blockchain-api/blob/develop/docs/openapi.yaml and those point to json schema files with the response structures.

docs/entities/transaction-events/transaction-event-fungible-asset.schema.json
```diff
@@ -14,6 +14,9 @@
           "type": "string",
           "enum": ["fungible_token_asset"]
         },
+        "tx_id": {
```
zone117x, line 17: Can you add these `tx_id` props to the `required` array above in all of these?

zone117x (approved):
LGTM!

PR #720, opened 2021-08-23
expose FT and NFT transfers in /extended/v1/address/[:principal]/transactions_with_transfers

### LINKED ISSUE #615: Expose associated transactions for FT/NFTs for a given address
We have an endpoint that does this but only for transactions that have a token transfer (stx) event -> https://blockstack.github.io/stacks-blockchain-api/#operation/get_account_inbound

We need an ability to see these kind of events for FT and NFT tokens too. 

Until this is exposed via the API, the wallets won't be able to display incoming transactions from others that result in some amount of a token being sent to the recipient. 

cc @andresgalante, maybe this is something user-x can take on?

rafaelcr:
## Description

Add FT and NFT transfers to the list of transactions listed in `/extended/v1/address/[:principal]/transactions_with_transfers` so wallets can display this information to users.

Closes #615 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No.

Yes. Endpoint descriptions should be updated.

## Testing information

Covered in existing unit tests.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/docs/api/address/get-address-transactions-with-transfers.example.json
+++ b/docs/api/address/get-address-transactions-with-transfers.example.json
@@ -70,7 +70,7 @@
       ],
       "ft_transfers": [
         {
-          "amount": "10.53",
+          "amount": "103",
           "asset_identifier": "SP466FNC0P7JWTNM2R9T199QRZN1MYEDTAR0KP27.miamicoin-token::miamicoin",
           "sender": "SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE",
           "recipient": "SP24Q9PK9DNTA2V89APY8WNBJ1QYKKSW9SWB04RJP"
--- a/docs/api/address/get-address-transactions-with-transfers.example.json
+++ b/docs/api/address/get-address-transactions-with-transfers.example.json
@@ -71,14 +71,14 @@
       "ft_transfers": [
         {
           "amount": "10.53",
-          "asset_identifier": "COOL",
+          "asset_identifier": "SP466FNC0P7JWTNM2R9T199QRZN1MYEDTAR0KP27.miamicoin-token::miamicoin",
           "sender": "SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE",
           "recipient": "SP24Q9PK9DNTA2V89APY8WNBJ1QYKKSW9SWB04RJP"
         }
       ],
       "nft_transfers": [
         {
-          "asset_identifier": "punk2",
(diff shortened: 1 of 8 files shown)


docs/api/address/get-address-transactions-with-transfers.example.json
```diff
@@ -67,6 +67,22 @@
           "sender": "SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE",
           "recipient": "SP26066SDPP4NXKGCVYZQDR5GX2QPE8KADZ0YK2J7"
         }
+      ],
+      "ft_transfers": [
+        {
+          "amount": "10.53",
+          "asset_identifier": "COOL",
```
zone117x, line 74: The actual responses are fully qualified right, should we use a more accurate example? E.g. `SP466FNC0P7JWTNM2R9T199QRZN1MYEDTAR0KP27.miamicoin-token::miamicoin`

docs/api/address/get-address-transactions-with-transfers.example.json
```diff
@@ -67,6 +67,22 @@
...
+          "sender": "SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE",
+          "recipient": "SP24Q9PK9DNTA2V89APY8WNBJ1QYKKSW9SWB04RJP"
+        }
+      ],
+      "nft_transfers": [
+        {
+          "asset_identifier": "punk2",
```
zone117x, line 81: Same here, something fully qualified like `SP497E7RX3233ATBS2AB9G4WTHB63X5PBSP5VGAQ.boom-nfts::boom`

zone117x (approved):
Thanks @rafaelcr, looks great!

Just waiting on CI flakiness right now, otherwise looks good to merge.

docs/api/address/get-address-transactions-with-transfers.example.json
```diff
@@ -67,6 +67,22 @@
           "sender": "SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE",
           "recipient": "SP26066SDPP4NXKGCVYZQDR5GX2QPE8KADZ0YK2J7"
         }
+      ],
+      "ft_transfers": [
+        {
+          "amount": "10.53",
```
zone117x, line 73: Ah one last thing -- this will always be an integer, since we don't evaluate possible `decimals` defined for a given FT. It would probably be useful to note that in the schema description, something like "Note this balance does not factor in possible SIP-010 `decimals`".

Issue #647, opened 2021-07-09
Fetch by array of ids for various endpoints


aulneau:
As we have been trying to get a handle on data fetching and many requests, I've often wished I could pass an array of txids to the tx endpoint to batch fetch a set of transactions. The first example of where this would be really useful is on the blocks page of the explorer. Currently the block response only includes an array of txids, and so to be able to display richer details for each transaction on that page, the application has to fetch each of them.

It'd be great if other endpoints outside of the `tx` endpoint supported this functionality too, eg query a set of addresses or blocks.

zone117x:
Great suggestion!

zone117x:
@aulneau let me know if you think this should be bumped in priority.

aulneau:
I think this is pretty important, specifically related to work on the explorer. we can make due now, but it will hammer the api with requests, especially as we need to fetch large lists of txids and microblock hashes

MHassanTariq:
I'm working on this issue.

zone117x:
Closed by https://github.com/blockstack/stacks-blockchain-api/issues/647

Note that a new endpoint was only added for fetching a set of txs. For other new batch endpoints we should open new issues.

PR #722, opened 2021-08-24
Transactions list details

MHassanTariq:
## Description

A new endpoint has been created to respond with a list of transactions, given the ids to them. Endpoint for a single transaction detail is already implemented. In case of multiple required transactions, the client would result in hammering the server for details regarding multiple transactions.

```
const api = `/extended/v1/tx/tx_list_details?tx_list=${tx_ids[0]}, ${tx_ids[1]}`;
```

For details refer to issue [#647](https://github.com/blockstack/stacks-blockchain-api/issues/647)

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


docs/openapi.yaml
```diff
@@ -213,6 +213,54 @@ paths:
...
 
+  /extended/v1/tx/tx_list_details:
+    parameters:
+      - name: tx_list
+        in: query
+        description: Array of transaction ids
+        required: true
+        schema:
+          type: array
+      - name: event_offset
```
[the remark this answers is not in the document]
MHassanTariq: As I am supposed to return a list of transactions and every transaction has events. So, just to make it clear, that the limit and offset are for the events within transactions, I used `event_` prefix. 
Plus, it was also used in the single [transaction detail endpoint](https://github.com/blockstack/stacks-blockchain-api/blob/master/docs/openapi.yaml#L216-L242).

src/api/controllers/db-controller.ts
```diff
@@ -786,43 +887,183 @@ export async function getTxFromDataStore(
...
+  // filtering out tx_ids that are not in canonical chain
+  const nonCanonical = minedTxs.filter(tx => !tx.canonical && !tx.microblock_canonical);
+
+  // filtering out tx_ids that were not mined / found
+  const notMinedTransactions: string[] = [];
+  args.txIds.forEach(txId => {
+    if (!minedTxs.find(minedTx => txId === minedTx.tx_id)) {
+      notMinedTransactions.push(txId);
+    }
+  });
```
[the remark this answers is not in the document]
MHassanTariq: I did that initially but that was allowing null values in the array, too. And additional checks were required to filter non-null values from array.

src/api/controllers/db-controller.ts
```diff
@@ -786,43 +887,183 @@ export async function getTxFromDataStore(
...
+    .map(tx => {
+      return { found: true, result: tx };
+    });
+  // all transactions that were not found anywhere
+  const foundOrNot_FoundTxs = [
+    ...foundOrNot_Mined,
+    ...foundOrNot_Mempool,
+    ...foundOrNot_NonCanonical,
+  ];
+  const foundOrNot_NotFoundTxs = args.txIds
```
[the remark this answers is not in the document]
MHassanTariq: Nice catch. 
I did that to make it more readable & understandable. `foundOrNot` arrays were created after calculating everything, i.e. adding 
```
found: true
``` 
However, keeping a track of all possibilities made too many `foundOrNot` variables. Mixing the purpose of that variable with `foundOrNot` prefix was making it difficult to understand for instance, `foundOrNotNonCanonical` and `foundOrNotCanonical`. So I used `_` as a separator between prefix and actual purpose for the array.

docs/openapi.yaml
```diff
@@ -213,6 +213,54 @@ paths:
...
+          default: false
+    get:
+      summary: Get list of details for transactions
+      tags:
+        - Transactions
+      operationId: get_tx_list_details
+      description: |
+        Get an array of transactions by IDs
+
+        `import type { Transaction } from '@stacks/stacks-blockchain-api-types';`
```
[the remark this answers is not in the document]
MHassanTariq: Corrected description

src/api/controllers/db-controller.ts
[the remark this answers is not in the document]
MHassanTariq: I have used camel-case.

src/api/controllers/db-controller.ts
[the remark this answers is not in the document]
MHassanTariq: I have changed the functionality according to your suggestion.

src/datastore/postgres-store.ts
```diff
@@ -3011,6 +3011,57 @@ export class PgDataStore
     return tx;
   }
 
+  async getMempoolTxs(args: {
+    txIds: string[];
+    includeUnanchored: boolean;
+    includePruned?: boolean;
+  }) {
+    return this.queryTx(async client => {
+      const conditionParam = this.generateParameterizedWhereOrClause('tx_id', args.txIds.length);
```
zone117x, line 3020: Looks like usages of `generateParameterizedWhereOrClause` can be replaced using `WHERE tx_id = ANY($1)` syntax, which would result in more readable and performant code

src/api/routes/tx.ts
```diff
@@ -89,6 +101,31 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
     res.json(response);
   });
 
+  router.getAsync('/tx_list_details', async (req, res, next) => {
+    const txList: string[] = getTxListFromQuery(req.query['tx_list'] as string | undefined);
```
zone117x, line 105: I think `getTxListFromQuery` should support the common serialization used for array query params. For example, Express supports automatic parsing of the array formats (also used by web browsers for forms like `<form method="get">`):
```
/?foo=1&foo=2
/?foo[]=1&foo[]=2
```


And I think by default OpenAPI uses the first one `/?foo=1&foo=2`: https://swagger.io/docs/specification/serialization/#query

The function should still support parsing comma-separated strings because that's also a common format.

docs/openapi.yaml
```diff
@@ -213,6 +213,56 @@ paths:
               example:
                 $ref: ./api/transaction/get-mempool-transactions.example.json
 
+  /extended/v1/tx/tx_list_details:
```
zone117x, line 216: @rafaelcr @agraebe thoughts on a possible more REST-compliant and readable endpoint name?

src/api/routes/tx.ts
rafaelcr: Since `tx_id`s are very long, is there any risk users get a `414 Request-URI Too Long` error when calling this endpoint?

docs/openapi.yaml
agraebe: this is essentially a filter on all txs by a property (ids in this case), right? If so, how about one of these:
-  /extended/v1/tx?id=x,y,z
-  /extended/v1/tx/filter?id=x,y,z

MHassanTariq: The purpose of this endpoint is not to filter transactions. It is designed to return the details for a list of transaction ids.

src/api/routes/tx.ts
MHassanTariq: @zone117x I have implemented your first approach that makes the endpoint as 
```
/extended/v1/tx/tx_list_details?tx_id=a&tx_id=b
```
where a and b are different `tx_id`s.

src/datastore/postgres-store.ts
MHassanTariq: Nice catch. Changed.

MHassanTariq:
@zone117x Vercel issue did not resolve for this PR. I rebased from develop and that fixed my Vercel issue on other PRs, however, for this, it remains constant. Any idea how to get this issue fixed?
You can look into the details [here](https://github.com/MHassanTariq/stacks-blockchain-api/actions/runs/1188630966):

docs/openapi.yaml
MHassanTariq: Yep, @zone117x `/multiple` sounds good. I'll change the endpoint to that.

docs/openapi.yaml
```diff
@@ -213,6 +213,56 @@ paths:
               example:
                 $ref: ./api/transaction/get-mempool-transactions.example.json
 
+  /extended/v1/tx/multiple:
+    parameters:
+      - name: tx_list
```
rafaelcr, line 218: I still have a question about this parameter. Do we know the max # of characters the API will accept in a GET request before throwing an error? So users have a known (even approximate) limit for how many `tx_id`s they can pass here.

src/api/routes/tx.ts
```diff
@@ -90,6 +91,31 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
     res.json(response);
   });
 
+  router.getAsync('/multiple', async (req, res, next) => {
+    const txList: string[] = req.query.tx_id as string[];
+    const eventLimit = parseTxQueryEventsLimit(req.query['event_limit'] ?? 96);
```
rafaelcr, line 96: Where does the 96 come from? Seems sensible to leave a short comment or something to know why this default exists.

docs/openapi.yaml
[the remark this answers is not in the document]
followmisha: ok @lgalabru

docs/openapi.yaml
MHassanTariq: Currently, I haven't set any limit to it. 
However, I do think there should be one. Do you have any recommendations for it? 
Or we can use the standard 50 or max of 100 as limit.

src/api/routes/tx.ts
MHassanTariq: Basically, I replicated the functionality of `/extended/v1/tx/{tx_id}`:
https://github.com/blockstack/stacks-blockchain-api/blob/master/docs/openapi.yaml#L216

96 was specified here for this functionality: 
https://github.com/blockstack/stacks-blockchain-api/blob/master/src/api/routes/tx.ts#L216

docs/openapi.yaml
rafaelcr: Maybe @CharlieC3 could chime in here. TBH I'm not sure what the limit could be, AFAIK that's more of a web server configuration and/or limitation.

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +929,176 @@ export async function getTxFromDataStore(
...
+    });
+  // all transactions that were not found anywhere
+  const foundOrNotFoundTxs = [...foundOrNotMined, ...foundOrNotMempool, ...foundOrNotNonCanonical];
+  const foundOrNotNotFoundTxs = args.txIds
+    .filter(txId => foundOrNotFoundTxs.findIndex(ftx => ftx.result.tx_id === txId) < 0)
+    .map(txId => {
+      return { found: false, result: txId };
+    });
+
+  return [...foundOrNotFoundTxs, ...foundOrNotNotFoundTxs];
```
zone117x, line 1101: Seems like this should be returning txs in the order that they were requested, as in this response array should be in the same order as the request arg array.

docs/openapi.yaml
```diff
@@ -213,6 +213,56 @@ paths:
...
+        If using TypeScript, import typings for this response from our types package:
+
+        `import type { Transaction } from '@stacks/stacks-blockchain-api-types';`
+      responses:
+        200:
+          description: Returns list of transactions with their details for corresponding requested tx_ids.
+          content:
+            application/json:
+              schema:
+                $ref: ./entities/transactions/transaction.schema.json
```
zone117x, line 260: The response schema should be an array of tx results right?

docs/openapi.yaml
CharlieC3: Responded in a DM.

docs/openapi.yaml
MHassanTariq: You're right. My bad. Created return type for this response and updated open.yaml.

src/api/controllers/db-controller.ts
MHassanTariq: Done.

docs/openapi.yaml
rafaelcr: @MHassanTariq per discussion with @CharlieC3 , my concern has to do more with the specific environment in which the API runs (which can vary greatly by installation), so we can mark it as resolved.

zone117x (changes requested):
Found the issue with vercel docs. Run `npm run build:docs` in the project directory, and you'll see the error:
image

Looks like an issue with the schema.

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +929,183 @@ export async function getTxFromDataStore(
...
+  return transactions;
+}
+
+async function getContractCallForTxsList(
+  db: DataStore,
+  dbTxs?: DbTx[],
+  dbMempoolTxs?: DbMempoolTx[]
+): Promise<Transaction[] | undefined> {
+  // get contract call ids
+  const transactions = dbTxs ?? dbMempoolTxs ?? undefined;
```
zone117x, line 974: This function is a bit confusing. Both dbTxs and dbMempoolTxs can be passed as separate args, but only one or none will be used. Seems like this function should instead take a single arg for txs using common interface like `txs: BaseTx[]`.

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +929,183 @@ export async function getTxFromDataStore(
...
 async function getContractCallMetadata(
   db: DataStore,
-  dbTx: DbTx | DbMempoolTx,
-  parsedTx: Transaction | MempoolTransaction
-): Promise<void> {
+  dbTx?: DbTx,
+  dbMempoolTx?: DbMempoolTx
+): Promise<ContractCallTransaction | MempoolContractCallTransaction | undefined> {
   // If tx type is contract-call then fetch additional contract ABI details for a richer response
+  const dbTransaction = db ?? dbMempoolTx ?? undefined;
```
zone117x, line 993: Same comment here -- unclear why there are 2 optional types of tx arrays for args, then only one of them is used.

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +929,183 @@ export async function getTxFromDataStore(
...
-    const functionAbi = contractAbi.functions.find(
-      fn => fn.name === parsedTx.contract_call.function_name
+    return parseContractCallMetadata(contract, dbTx);
+  }
+}
+
+function parseContractCallMetadata(
+  contract: FoundOrNot<DbSmartContract>,
+  dbTx?: DbTx,
+  dbMemTx?: DbMempoolTx
```
zone117x, line 1014: Same here

src/datastore/postgres-store.ts
```diff
@@ -3108,6 +3108,53 @@ export class PgDataStore
...
+        // The tx is marked as pruned because it's in an unanchored microblock
+        query.rows.forEach(tran => {
+          const transaction = result.rows.find(tx => tx.tx_id === tran.tx_id);
+          if (transaction) {
+            transaction.pruned = false;
+            transaction.status = DbTxStatus.Pending;
+          }
+        });
+      }
+      return result.rows.map(transaction => this.parseMempoolTxQueryResult(transaction));
```
zone117x, line 3154: There's non-trivial logic duplicated in this function (originally in `getMempoolTx`), is it possible to share the logic between the two?

src/datastore/postgres-store.ts
```diff
@@ -3754,6 +3998,26 @@ export class PgDataStore
 
   cachedParameterizedInsertStrings = new Map<string, string>();
 
+  generateParameterizedWhereAndOrClause(
```
zone117x, line 4001: The query output of this function is hard to reason about, and I'm not sure it's correct.
It looks like it could output statements like `WHERE ... AND ... OR ... AND ... ` which don't have parenthesis around them so unclear about order of operations, e.g. will it run `WHERE (... AND ...) OR (... AND ...)` or `WHERE ... AND ( ... OR ... AND ...)`, etc. It looks like only one function is calling this, can that function be simplified for a more readable query?

MHassanTariq: The requirement was to gather results based on two attributes (`tx_id` and `index_block_hash`) within events.
You're right. I'll change the query to `WHERE ( ... AND ... ) OR ( ... AND ... )` as we want results with similar `tx_id` and `index_block_hash`.
Regarding the readability, do you think loop is more preferred than recursion?

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +939,153 @@ export async function getTxFromDataStore(
...
+  const foundOrNotFoundTxs = [...foundOrNotMined, ...foundOrNotMempool, ...foundOrNotNonCanonical];
+  const foundOrNotNotFoundTxs = args.txIds
+    .filter(txId => foundOrNotFoundTxs.findIndex(ftx => ftx.result.tx_id === txId) < 0)
+    .map(txId => {
+      return { found: false, result: txId };
+    });
+
+  // sorting according to incoming tx_ids
+  const resp = [...foundOrNotFoundTxs, ...foundOrNotNotFoundTxs];
+  resp.sort((tx1, tx2) => {
```
rafaelcr, line 1083: Is this sorting really necessary? Seems like a very expensive operation that doesn't provide too much value to the response. If your goal is to allow users to get transaction results by index, why not switch to an associated array response? It would be much easier to use and much less expensive to produce.

For example, instead of returning this for each transaction:
```
{
  "found": true,
  "result": {
    "tx_id": "0x1234"
  }
}
```
You can do this:
```
{
  "0x1234": {
    "found": true,
    "result": {
      "tx_id": "0x1234"
    }
  }
  (...other txs here)
}
```
This way, users can parse the object and quickly search each tx by tx_id.

src/datastore/postgres-store.ts
```diff
@@ -3754,6 +3987,34 @@ export class PgDataStore
...
+      return '';
+    } else if (valuesCount === 1) {
+      return condition;
+    }
+
+    return (
+      condition +
+      ' OR ' +
+      this.generateParameterizedWhereAndOrClause(column1, column2, valuesCount - 1)
+    );
```
zone117x, line 4015: I'm still wary of the query string generation here. We've previously only done something like this for batch inserts of large datasets (where dynamically generating parameterized batch-insert queries is one of the fastest methods).

If generating a query string for this is absolutely necessary, we should increase the readability of this code. For example, it could use the syntax `WHERE (a,b) = ANY(VALUES (1,1),(1,2))` to avoid generating a string with a bunch of conditionals and redundant column names. Also, if this code is only called by one function, it would be more clear if it was defined within that function.

src/api/controllers/db-controller.ts
MHassanTariq: @rafaelcr that is a good suggestion. @zone117x suggested the returning txs be in the order they were requested.
@zone117x what do you think about this approach?

MHassanTariq: Done

src/datastore/postgres-store.ts
MHassanTariq: Done

src/api/controllers/db-controller.ts
```diff
@@ -778,17 +823,79 @@ export async function getMempoolTxFromDataStore(
...
   if (parsedMempoolTx.tx_type === 'contract_call') {
-    await getContractCallMetadata(db, mempoolTxQuery.result, parsedMempoolTx);
+    const transaction = await getContractCallMetadata(
+      db,
+      parseDbMempoolTx(mempoolTxQuery.result) as MempoolContractCallTransaction,
+      mempoolTxQuery.result
+    );
+    if (transaction) {
+      parsedMempoolTx = transaction as MempoolContractCallTransaction;
+    }
```
rafaelcr, line 836: Instead of repeating logic here, why don't you call the new `getMempoolTxsFromDataStore` and validate/return the first result?

src/api/controllers/db-controller.ts
```diff
@@ -804,11 +911,18 @@ export async function getTxFromDataStore(
...
   if (parsedTx.tx_type === 'contract_call') {
-    await getContractCallMetadata(db, dbTx, parsedTx);
+    const transaction = await getContractCallMetadata(
+      db,
+      parseDbTx(dbTx) as ContractCallTransaction,
+      dbTx
+    );
+    if (transaction) {
+      parsedTx = transaction as ContractCallTransaction;
+    }
```
rafaelcr, line 925: Same here, reuse the new function above.

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +942,154 @@ export async function getTxFromDataStore(
   };
 }
 
+function parseContractsWithDbTx(contracts: DbSmartContract[], dbTxs: DbTx[]) {
```
rafaelcr, line 945: Plural in `parseContractsWithDbTxs`

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +942,154 @@ export async function getTxFromDataStore(
...
+      );
+      if (transaction) {
+        transactions.push(transaction as Transaction);
+      }
+    }
+  });
+  return transactions;
+}
+
+function parseContractsWithMempoolTx(contracts: DbSmartContract[], dbMempoolTx: DbMempoolTx[]) {
```
rafaelcr, line 963: Plural `parseContractsWithMempoolTxs`

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +942,154 @@ export async function getTxFromDataStore(
...
+  // filtering out mined transactions in canonical chain
+  const minedAndInCanonicalChain = minedTxs.filter(tx => tx.canonical && tx.microblock_canonical);
+
+  // filtering out tx_ids that are not in canonical chain
+  const nonCanonical = minedTxs.filter(tx => !tx.canonical && !tx.microblock_canonical);
+
+  // filtering out tx_ids that were not mined / found
+  const notMinedTransactions: string[] = args.txIds.filter(
+    txId => !minedTxs.find(minedTx => txId === minedTx.tx_id)
+  );
```
rafaelcr, line 1055: Instead of traversing `minedTxs` twice and `args.txIds` once, you should do a `forEach` over `minedTxs` just once and populate all required temp arrays.

This function in particular (`searchTxs`) has an unusual number of filter/map/find calls that could be optimized into fewer traversals. I think some optimizations could be made.

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +942,154 @@ export async function getTxFromDataStore(
...
+  const mempoolTxs = [...nonCanonical.map(tx => tx.tx_id), ...notMinedTransactions];
+  const mempoolTxsQuery = await getMempoolTxsFromDataStore(db, {
+    txIds: mempoolTxs,
+    includeUnanchored: args.includeUnanchored,
+  });
+
+  const foundOrNotMined = minedAndInCanonicalChain.map(minedTx => {
+    return { found: true, result: minedTx };
+  });
+  const foundOrNotMempool = mempoolTxsQuery.map((mtx: any) => {
```
rafaelcr, line 1067: Could you use the correct type here instead of `any`?

rafaelcr (changes requested):

src/api/controllers/db-controller.ts
MHassanTariq: Small clarification required - are you recommending that I use the same function for an individual transaction and for a list of transactions?

src/api/controllers/db-controller.ts
MHassanTariq: Done

src/api/controllers/db-controller.ts
MHassanTariq: I've refactored for the other maps. However, this one is the only traversal I'm using for `args.txIds`. I wanted to get all the transaction ids that weren't mined.

src/datastore/postgres-store.ts
```diff
@@ -3401,6 +3437,147 @@ export class PgDataStore
...
+      let condition = '(tx_id, index_block_hash) = ANY(VALUES ';
+      let counter = 1;
+      const transactionValues = args.txs
+        .map(_ => {
+          const singleCondition = '($' + counter + '::bytea, $' + (counter + 1) + '::bytea)';
+          counter += 2;
+          return singleCondition;
+        })
+        .join(', ');
+      condition += transactionValues + ')';
```
zone117x, line 3462: Thanks @MHassanTariq for the readability improvements in this query!

At first glance this seems okay performance-wise, but we need to do some manually testing against the mainnet dataset to be sure. We may need to add a composite index for `(tx_id, index_block_hash)`.

zone117x: cc @asimm241 is the performance profiling something you could help with?

Also, if you can come up with an isolated SQL query with some sample mainnet data, I can test against our prod DBs.

src/api/controllers/db-controller.ts
rafaelcr: Yes. Here, in the individual tx function, you'd call `getMempoolTxsFromDataStore` but with an array of only one tx (the one given as arg) and then take the first result to return.

src/api/controllers/db-controller.ts
```diff
@@ -828,43 +944,173 @@ export async function getTxFromDataStore(
...
+  const foundOrNotNonCanonical: TransactionFound[] = nonCanonical
+    .filter(
+      tx =>
+        mempoolTxsQuery.findIndex(
+          (mtx: Transaction | MempoolTransaction) => mtx.tx_id === tx.tx_id
+        ) < 0
+    )
+    .map(tx => {
+      return { found: true, result: tx };
+    });
```
rafaelcr, line 1090: Perhaps I don't fully understand the purpose of these loops yet. Why do you need to separate mempool txs from non canonical, etc.? Is that information relevant to the response? It seems like in a few lines below you merge all these arrays together.

src/api/controllers/db-controller.ts
MHassanTariq: `getTxFromDataStore` function is called from various APIs outside `db-controller.ts`. I didn't replicate the entire implementation in `getTxsFromDataStore` and just refactored the code needed to work for this PR.

aulneau:
@zone117x @rafaelcr is this currently deployed?

zone117x:
> @zone117x @rafaelcr is this currently deployed?

Not yet, still in develop branch. We're looking to cut a new release this week though.

PR #730, opened 2021-08-31
Return only latest name by address

### LINKED ISSUE #714: Remove transferred names from address endpoint
Currently it appears that names are still returned by the address endpoint even once they've been transferred to another address.

See https://github.com/blockstack/stacks-wallet-web/issues/1551#issuecomment-897690761

> Hmm digging into this more, the api reports the original owner as the current owner: https://stacks-node-api.mainnet.stacks.co/v1/addresses/stacks/SPTCSBVXZ6KNV7JEMHAV7KJM9183VAP31053H0R4
> 
> However, there was a name transfer event: https://explorer.stacks.co/txid/0x79ad2a1a264d049f13842f0f055f0a40073488c74277fb4a899ec0355d13d8a6?chain=mainnet
> 
> The api shows this account: https://stacks-node-api.mainnet.stacks.co/extended/v1/address/SP3JKC09PH0RH8G1XVT031J3E3S0GJ15DEMG7EAZP/assets does infact own the name.
> 
> This is likely an API bug that should be fixed when we have a new deployment, but cc @zone117x to be sure.
> 
> Once microblocks is launched on the api side, we should come back and confirm this is correct.

_Originally posted by @aulneau in https://github.com/blockstack/stacks-wallet-web/issues/1551#issuecomment-897690761_

> Looking into this today, I seem to have the same issue in one of my accounts. This name transfer was successful, but it is still showing up in my wallet: https://explorer.stacks.co/txid/0x30a1a3d18709ddcd70ddb4fb904feeaf3cf5538aa724b7b5e78081fed6395ad0?chain=mainnet
> 
> Transferred from:
> https://stacks-node-api.mainnet.stacks.co/v1/addresses/stacks/SP2J1ZFZ5BYTD7H7F84CCZ2B4HEJSBV060XW79G1X
> 
> Transferred to:
> https://stacks-node-api.mainnet.stacks.co/v1/addresses/stacks/SP9CMJ7S8XR25H6ZKAJXT4M7KSPQ1B8PPVSYJRTC

_Originally posted by @fbwoolf in https://github.com/blockstack/stacks-wallet-web/issues/1551#issuecomment-898683711_

asimm241:
## Description
This PR fixes the bug described in #714. The names transferred to the new address should only be returned under the new address. 

closes #714

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -5647,7 +5647,7 @@ export class PgDataStore
 
       if (query.rowCount > 0) {
         const rows: { name: string; block_height: number }[] = [];
-        query.rows.forEach(async row => {
+        for (const row of query.rows) {
           const query_latest = await client.query<{ name: string }>(
             `
             SELECT name FROM (
@@ -5670,8 +5670,8 @@ export class PgDataStore
             [row.name, row.block_height]
           );
 
-          if (query_latest.rowCount === 0) rows.push(row); //if there is no names exit at greater height
-        });
+          if (query_latest.rowCount === 0) rows.push(row); //if there is no names exist at greater height
+        }
 
         query.rows = rows;
       }
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -5624,22 +5624,21 @@ export class PgDataStore
         `
         SELECT name FROM (
           (
-            SELECT DISTINCT ON (name) name, registered_at as block_height, tx_index
+            SELECT DISTINCT ON (name) name, registered_at as block_height, tx_index, address
             FROM names
-            WHERE address = $1
-            AND registered_at <= $2
+            WHERE registered_at <= $2
             AND canonical = true AND microblock_canonical = true
             ORDER BY name, registered_at DESC, tx_index DESC
           )
           UNION ALL (
-            SELECT DISTINCT ON (fully_qualified_subdomain) fully_qualified_subdomain as name, block_height, tx_index
+            SELECT DISTINCT ON (fully_qualified_subdomain) fully_qualified_subdomain as name, block_height, tx_index, owner as address
             FROM subdomains
-            WHERE owner = $1
-            AND block_height <= $2
+            WHERE block_height <= $2
             AND canonical = true AND microblock_canonical = true
             ORDER BY fully_qualified_subdomain, block_height DESC, tx_index DESC
           )
         ) results
+        WHERE address = $1
         ORDER BY block_height DESC, tx_index DESC
         `,
         [address, maxBlockHeight]
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -5646,7 +5646,8 @@ export class PgDataStore
       );
 
       if (query.rowCount > 0) {
-        query.rows = query.rows.filter(async row => {
+        const rows: { name: string; block_height: number }[] = [];
+        query.rows.forEach(async row => {
           const query_latest = await client.query<{ name: string }>(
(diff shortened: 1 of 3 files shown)


src/datastore/postgres-store.ts
```diff
@@ -5476,22 +5476,21 @@ export class PgDataStore
...
-            AND registered_at <= $2
+            WHERE registered_at <= $2
             AND canonical = true AND microblock_canonical = true
             ORDER BY name, registered_at DESC, tx_index DESC
           )
           UNION ALL (
-            SELECT DISTINCT ON (fully_qualified_subdomain) fully_qualified_subdomain as name, block_height, tx_index
+            SELECT DISTINCT ON (fully_qualified_subdomain) fully_qualified_subdomain as name, block_height, tx_index, owner as address
             FROM subdomains
-            WHERE owner = $1
```
zone117x, line 5489: The only reason this query was performant enough is because of this index:
```ts
  pgm.createIndex('subdomains', [
    { name: 'fully_qualified_subdomain' },
    { name: 'canonical', sort: 'DESC' },
    { name: 'microblock_canonical', sort: 'DESC' },
  ]);
```

Adding the extra `owner` column to the query makes the index unusable. 

I'm not sure we could just add `owner` to the index because it looks like this query would end up gathering _all_ the subdomains in memory before excluding entries in the outer WHERE clause. We'd need to run some manual benchmarks on this query against the fully imported subdomain list.

asimm241: @zone117x You were right about it, this approach was taking 129666ms(129 seconds). 
I have updated it, the new approach takes 137ms for the main query, and the second query will take 137* N (N=number of rows returned in the first query).

src/datastore/postgres-store.ts
```diff
@@ -5618,6 +5618,36 @@ export class PgDataStore
         `,
         [address, maxBlockHeight]
       );
+
+      if (query.rowCount > 0) {
+        query.rows = query.rows.filter(async row => {
```
zone117x, line 5623: I don't think this is working as intended. The `filter` function does not handle promises, so it will see the `truthy` promise and rows will never be filtered, and these promises will still be evaluating even after the function exists since nothing is awaiting them.

zone117x (approved):
We'll want to keep an eye on this query, since it will increase response time linearly with the number of names owned by an address. Seems alright for now though.

src/datastore/postgres-store.ts
```diff
@@ -5644,6 +5644,38 @@ export class PgDataStore
...
+                WHERE fully_qualified_subdomain = $1
+                AND block_height > $2
+                AND canonical = true AND microblock_canonical = true
+              )
+            ) results
+            `,
+            [row.name, row.block_height]
+          );
+
+          if (query_latest.rowCount === 0) rows.push(row); //if there is no names exist at greater height
```
rafaelcr, line 5673: Instead of moving results back and forth from postgres to node to filter, you could try doing all the filtering in a single query. The idea would be:

1. In a common table expression or subquery, get all names ever owned by $1
2. From step 1 name results, query to get the latest owner address (limit 1 ordered by block desc)
3. From step 2 results, select only the names where the owner is $1

rafaelcr: This way you avoid the O(n) problem in node and leave that to the DB to handle

zone117x (approved):
LGTM!

rafaelcr (approved):
Excellent 🏁

PR #736, opened 2021-09-02
Fee rate endpoint

### LINKED ISSUE #729: New competitive fee endpoint
Per the requirement to set higher fees, the API should expose an additional endpoint that contains fee rate information for clients.

This is preferable to hard coding in our client apps, as updates would require us to go through the slow release cycles.

The simplest implementation of this would be an endpoint with the schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "description": "FeeRate",
  "type": "object",
  "additionalProperties": false,
  "required": ["fee_rate"],
  "properties": {
    "fee_rate": {
      "type": "integer"
    }
  }
}
```

asimm241:
## Description
Added a new endpoint for fee_rate. See details here #729 

closes #729 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


rafaelcr (approved):
LGTM

docs/api/info/get-fee-rate.schema.example.json
```diff
@@ -0,0 +1,3 @@
+{
+  "fee_rate": 360
```
diwakergupta, line 2: Are the units obvious to everyone? I feel like somewhere in the OpenAPI docs or comments we should cover how exactly to use fee_rate to derive the fee. Thoughts @kyranjamie @zone117x ?

zone117x: It's not really clear to me. IIUC, the unit currently used by `/v2/fees/transfer` is simply byte length, which was a simple heuristic shipped with Stacks 2.0 but not very accurate? It seems like in the future (e.g. Stacks 2.1) there would be something like a fee _rate_ for multiple types of execution limits (read, write, runtime, etc), as opposed to a single value like this.

Given how fee estimation currently works, a simple fee multiplier makes more sense to me. This is what was just implemented in the send-many CLI.

docs/openapi.yaml
```diff
@@ -2659,3 +2659,20 @@ paths:
                 $ref: ./entities/tokens/fungible-token.schema.json
               example:
                 $ref: ./entities/tokens/fungible-token.schema.example.json
+
+  /extended/v1/fee_rate:
+    get:
```
zone117x, line 2664: This should be a POST that requires the serialized tx, so that in the future this endpoint can take advantage of more advanced fee estimation.

docs/api/info/get-fee-rate.schema.example.json
zone117x: Could use some guidance from @reedrosenbluth @kantai

kyranjamie (approved):
LGTM, we can quite integrate this into the wallets 👍🏼 

Before we merge, who owns the decision as to the exact number we'll set the fee rate too? Note a STX token transfer is 180 bytes (generated by stacks.js) cc/ @diwakergupta @markmhx

The wallet's currently have a fee rate of 400 hard coded.

markmhx:
This rate will become variable once we have the ability to estimate based on transaction and network characteristics, correct?

I presume we should release and integrate this API-side support in the meantime with the same value we've hardcoded into the wallet, to keep things at parity.

zone117x:
> This rate will become variable once we have the ability to estimate based on transaction and network characteristics, correct?

IIUC if we merge this endpoint with clients performing a simple multiplication of this `fee_rate` and the tx byte length, then this endpoint won't have much room to upgrade to the anticipated smart estimation involving block proportion. And this endpoint would become deprecated in favor of a new one.

fbwoolf (approved):

src/api/routes/fee-rate.ts
```diff
@@ -0,0 +1,20 @@
+import * as express from 'express';
+import { addAsync, RouterWithAsync } from '@awaitjs/express';
+import { DataStore } from '../../datastore/common';
+import { FeeRate } from '@stacks/stacks-blockchain-api-types';
+
+export const FEE_RATE = 346;
```
zone117x, line 6: @asimm241 can you change this to `400` to match the current hard coded number in the wallets?

PR #737, opened 2021-09-02
add execution cost data to transactions

### LINKED ISSUE #734: Store & expose tx execution cost data
Tx execution cost is reported by the stacks-node, the property `execution_cost` is defined here: https://github.com/blockstack/stacks-blockchain-api/blob/3100c5661e62fece6b33bfe2806940e3ea655425/src/event-stream/core-node-message.ts#L162-L169

and the interface here: https://github.com/blockstack/stacks-blockchain-api/blob/3100c5661e62fece6b33bfe2806940e3ea655425/src/event-stream/core-node-message.ts#L292-L298

Note: `execution_cost` shouldn't be defined as optional:
> Aaron B: It's not optional -- every tx message should have that value

This data is currently discarded. It should be stored in the db, and returned in API tx responses (the property should be added to https://github.com/blockstack/stacks-blockchain-api/blob/develop/docs/entities/transactions/abstract-transaction.schema.json)

Each of the execution cost units should probably have their own column in the `txs` table.

rafaelcr:
## Description

This change captures execution cost data from raw node transactions and saves it as columns in the PostgreSQL `txs` table. It also exposes this new data to transaction API endpoints so users can see them.

Closes #734 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

Only schema changes which will be automated into docs

## Testing information
Covered by unit tests

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


zone117x (approved):
Quick turnaround! LGTM!

PR #739, opened 2021-09-02
fix: sql optimizations to speed up various tx queries

zone117x:
* Add sql index to speed up tx list queries.
* Remove redundant sql queries for already found tx db data.
* Make the max pg connection pool limit configurable via env var.

### DIFF
--- a/src/api/routes/tx.ts
+++ b/src/api/routes/tx.ts
@@ -75,7 +75,7 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
 
     // TODO: use getBlockWithMetadata or similar to avoid transaction integrity issues from lazy resolving block tx data (primarily the contract-call ABI data)
     const results = await Bluebird.mapSeries(txResults, async tx => {
-      const txQuery = await getTxFromDataStore(db, { txId: tx.tx_id, includeUnanchored });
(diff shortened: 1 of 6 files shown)


src/api/routes/tx.ts
```diff
@@ -75,7 +75,7 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
 
     // TODO: use getBlockWithMetadata or similar to avoid transaction integrity issues from lazy resolving block tx data (primarily the contract-call ABI data)
     const results = await Bluebird.mapSeries(txResults, async tx => {
-      const txQuery = await getTxFromDataStore(db, { txId: tx.tx_id, includeUnanchored });
+      const txQuery = await getTxFromDataStore(db, { txId: tx.tx_id, dbTx: tx, includeUnanchored });
```
rafaelcr, line 78: Shouldn't this be `includeUnanchored: includeUnanchored` ?

zone117x: As opposed to the shorthand of `includeUnanchored`? It's only a stylistic difference. I don't feel strongly either way

rafaelcr: Oh, I think I just hadn't seen that shorthand before 👍 all good

rafaelcr (approved):
Looks great

wileyj (approved):
shipit

PR #749, opened 2021-09-09
optimize query that retrieves txs with asset transfers

rafaelcr:
## Description

This patch greatly optimizes the database query that retrieves transactions with asset transfers for an address or principal.

The main problem with the original query was that there were too many `LEFT JOIN` statements between event tables and the `txs` table which has hundreds of thousands of records. These operations were very expensive and created delays of a few seconds.

What was done was to first get the actual `tx_id`s that we're interested in from the event tables (using common table expressions for better performance), and then do an `INNER JOIN` between those results and the `txs` table.

These improvements reduced the latency of the original query from ~3,000 ms to ~50 ms in local tests.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Testing information
Tested using existing unit tests

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -4693,19 +4693,16 @@ export class PgDataStore
         }
       >(
         `
-        SELECT
-          tx_results.*,
-          events.event_index as event_index,
-          events.event_type_id as event_type,
-          events.amount as event_amount,
-          events.sender as event_sender,
-          events.recipient as event_recipient,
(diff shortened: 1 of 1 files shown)


zone117x (approved):
Awesome, thanks @rafaelcr!
> reduced the latency of the original query from ~3,000 ms to ~50 ms in local tests.
My local testing shows the same improvement, and result data looks correct.

PR #750, opened 2021-09-09
replicate query optimizations to other asset txs queries

rafaelcr:
## Description
This patch optimizes two other queries that retrieve transactions with asset transfers by using `INNER JOIN` instead of `LEFT JOIN` for the `txs` table, similar to #749 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Testing information
Covered by existing unit tests

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -4562,7 +4562,10 @@ export class PgDataStore
       }
       const resultQuery = await client.query<TxQueryResult & { count: number }>(
         `
-        WITH transactions AS (
+        WITH principal_txs AS (
+          WITH event_txs AS (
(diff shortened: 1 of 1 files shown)


zone117x (approved):
Thanks @rafaelcr!

On my local machine (postgres v12) I'm seeing a nearly 50% increase in query timing for `getAddressTxs`, however, on the managed instance (postgres v11) the speed is about the same or a bit slower. Looking at the query analysis, pg12 is transforming one of the CTEs into a subquery whereas pg11 is not, which seems like the source of the difference (not really sure).

Both the old and updated queries `getAddressTxs` queries take ~500-700 ms on prod right now, but I suspect would speed up once we upgrade to pg12.

PR #758, opened 2021-09-13
fix: disable http keep-alive for stacks-node /v2 proxied endpoints

zone117x:
(Hopefully) closes https://github.com/blockstack/stacks-blockchain-api/issues/756

Disables the http keep-alive header in the /v2 proxy to the Stacks core node. This should prevent individual API instances from sticking to specific stacks-node instances.


rafaelcr (approved):

CharlieC3 (approved):
:shipit:

PR #759, opened 2021-09-13
Execution cost for in block response

### LINKED ISSUE #735: Return block execution cost data
As a followup to https://github.com/blockstack/stacks-blockchain-api/issues/734, it would also be helpful if block data API responses included the total execution cost (should be a simple sum of the execution costs for each tx included in the block).

This can be done in a followup PR and is lower priority, since #734 would allow for clients to calculate this themselves. However, we'd want to implement this directly in the API if the Explorer were to present this data in order to prevent tx endpoints from being hammered.

Total block execution costs should probably be calculated and stored as new columns in the block table (as opposed to adding mildly expensive JOINs to the block sql queries).

asimm241:
## Description
Added the execution cost(sum of the execution costs for each tx included in the block) in `blocks` and block response. 

closes #735 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

NO

Schema changes should be enough

## Testing information
Unit tests coverage updated

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] Tag @zone117x for review

### DIFF
--- a/docs/entities/blocks/block.schema.json
+++ b/docs/entities/blocks/block.schema.json
@@ -18,7 +18,11 @@
     "parent_microblock_sequence",
     "microblocks_accepted",
     "microblocks_streamed",
-    "total_execution_cost"
+    "execution_cost_read_count",
+    "execution_cost_read_length",
+    "execution_cost_runtime",
+    "execution_cost_write_count",
+    "execution_cost_write_length"
   ],
   "properties": {
     "canonical": {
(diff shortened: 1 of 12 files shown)


src/tests/api-tests.ts
```diff
@@ -4928,6 +4948,7 @@ describe('api tests', () => {
       burn_block_height: 123,
       miner_txid: '0x4321',
       canonical: true,
+      total_execution_cost: 0,
```
rafaelcr, line 4951: Could you add (or modify) some tests to make the total execution cost non-zero? In order to exercise the reduce function.

docs/entities/blocks/block.schema.json
```diff
@@ -72,6 +88,10 @@
         "type": "string",
         "description": "Microblock hash"
       }
+    },
+    "total_execution_cost": {
```
zone117x, line 92: I think this should have each of the individual execution cost units (`read_count`, `read_length`, `runtime`, `write_count`, `write_length`).

zone117x (approved):
LGTM!

rafaelcr (approved):

PR #769, opened 2021-09-17
feat: add read-only mode

### LINKED ISSUE #642: Ability to run the API in a read-only mode
The API currently must be ran in a mode that handles both writing (consuming event-emitter data and performing postgres updates) and reading (serving the public API requests). 

It would be useful if the API could be ran in a read-only mode where it's only responsible for serving API requests. It would only perform postgres reads, and would not require running alongside an event-emitter stacks-node.

These read-only instances could be easily spun up and load balanced. A single write-mode instance would be populating the postgres db, and multiple read-mode instances could read from the db.

An initial point of complexity that comes to mind is supporting client-push (i.e. websockets / socket.io) when in read-only mode. The way it currently works is that the datastore class emits events on writes, which are picked up by the socket.io handler to send updates to clients. This EventEmitter mechanism could be replaced with a postgres LISTEN/NOTIFY system. This lib looks promising https://www.npmjs.com/package/pg-listen.

/cc @CharlieC3 @wileyj for any comments, suggestions

rafaelcr:
## Description

This change allows the API to run in read-only mode, i.e. without an event server that listens for node events to write to a local Postgres DB. This mode will allow us to spin up only one read/write instance that writes to a DB and multiple read-only instances that read from the same DB, enabling us to balance the request load much easier.

The change also incorporates the [pg-listen](https://www.npmjs.com/package/pg-listen) library to replace manual DB write notifications that were used for emitting Web Socket and socket-io events. These notifications, however, were implemented only at the `PgDataStore` level and are all translated into the usual EventEmitter so the rest of the API can receive them.

### Notes

In order to use Postgres NOTIFY/LISTEN messages, we have to be careful with the amount of information we send as these messages have [a hard 8000 byte cap](https://www.postgresql.org/docs/12/sql-notify.html) to their length. Thus, this PR also shortens the message payloads sent by the EventEmitter so we only send IDs instead of complete objects where possible.

Closes #642 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No, it is only toggled via an ENV var

Probably, to instruct users on how to enable read-only mode.

## Testing information
Running a read-only API locally against a remote testnet DB and node, for example, could be helpful to test. All the rest of the NOTIFY/LISTEN functionality is covered by existing unit tests.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


CharlieC3:
@rafaelcr Just a heads up, read-only and read/write APIs will likely be serving traffic on the same domain. Typically in a service cluster which has read/write and read-only service archetype, the read-only service will forward any _write_ requests to a known read/write service. So in this case, a read-only API which receives a write request, would forward/proxy the request to a known read/write API, where the known read/write API URL could be configured in an ENV var.

Is this functionality you're planning on adding?

rafaelcr:
@CharlieC3 actually, the only endpoints that write data belong to `/v2` so those would be covered by the usual forwarding. The only exception I'm looking at some Rosetta endpoints, but AFAIK they also issue the writes to the core node RPCs. Let me get back to you on that last part.

src/datastore/postgres-notifier.ts
```diff
@@ -0,0 +1,138 @@
...
+              dataType: 'Set',
+              value: Array.from(value.entries()),
+            };
+          }
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+          return value;
+        }),
+      parse: serialized =>
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+        JSON.parse(serialized, (_, value) => {
```
zone117x, line 79: Maybe safer and more performant to require all `PgNotificationPayloads` to be plain-old-JSON? Like this
```ts
export interface AddressTxUpdateInfo {
  address: string;
  txs: Map<string, Set<AddressTxUpdateEventInfo>>;
}
```
could be refactored to
```ts
export interface AddressTxUpdateInfo {
  address: string;
  txs: Record<string, AddressTxUpdateEventInfo[]>; // Record is just typescript type for object
}
```

And `bigint` properties can instead be string-quoted integers i.e. `"12345"` as opposed to `12345n`

rafaelcr: Yeah that's a good idea 👍

rafaelcr:
@zone117x PR is ready for review 🙌

src/datastore/common.ts
```diff
@@ -348,20 +343,42 @@ export interface AddressNftEventIdentifier {
...
+
 export type DataStoreEventEmitter = StrictEventEmitter<
   EventEmitter,
   {
-    txUpdate: (info: DbTx | DbMempoolTx) => void;
+    txUpdate: (txId: string) => void;
     blockUpdate: (
-      block: DbBlock,
+      blockHash: string,
       txIds: string[],
```
rafaelcr, line 374: @zone117x do you think I could also remove this array from this message and get the `tx_id`s later? Not sure if reorgs affect these results.

I'm afraid that if a block with hundreds of txs comes along, this array will be enough to break the 8000 byte limit for postgres NOTIFY. What do you think?

zone117x: Hmm yeah, seems like this could exceed the 8000 byte limit and should be removed.

> Not sure if reorgs affect these results

Shouldn't have an issue here. A given block hash should always have the same set of txs.

rafaelcr: Great. I'll make the change.

src/api/routes/socket-io.ts
```diff
@@ -125,93 +124,100 @@ export function createSocketIORouter(db: DataStore, server: http.Server) {
...
-    const addrTxTopic: AddressTransactionTopic = `address-transaction:${info.address}` as const;
+  db.on('addressUpdate', async (address, blockHeight) => {
+    const addrTxTopic: AddressTransactionTopic = `address-transaction:${address}` as const;
+    const addrStxBalanceTopic: AddressStxBalanceTopic = `address-stx-balance:${address}` as const;
+    if (!adapter.rooms.has(addrTxTopic) && !adapter.rooms.has(addrStxBalanceTopic)) {
+      return;
+    }
+    const dbTxsQuery = await db.getAddressTxsWithAssetTransfers({
+      stxAddress: address,
+      limit: 20,
```
rafaelcr, line 171: @zone117x what do you think is a sensible limit to place here? Ideally I'd like to get all the asset transfers in this tx

zone117x: When this query is limited to a single block, seems reasonable to fetch all txs

src/datastore/postgres-store.ts
```diff
@@ -1099,18 +1150,17 @@ export class PgDataStore
...
-    this.emit('blockUpdate', data.block, txIdList, microblocksAccepted, microblocksStreamed);
-    data.txs.forEach(entry => {
-      this.emit('txUpdate', entry.tx);
+    await this.notifier.sendBlock({
+      blockHash: data.block.block_hash,
+      microblocksAccepted: microblocksAccepted,
+      microblocksStreamed: microblocksStreamed,
+    });
+    data.txs.forEach(async entry => {
+      await this.notifier.sendTx({ txId: entry.tx.tx_id });
```
zone117x, line 1159: This needs to be a regular `for (...)` loop, otherwise `.forEach` will run all these concurrently, and not actually await. Unless that's the goal?

rafaelcr:
All integration tests for the web wallet passed when pointed at a read-only instance. I think this PR is ready for a final review.

zone117x (approved):
🚀

PR #771, opened 2021-09-20
feat: ability to configure multiple tx broadcast endpoints

zone117x:
Closes https://github.com/blockstack/stacks-blockchain-api/issues/765

Implements tx "multicast" ability where tx broadcasts to `POST /v2/transactions` can be forwarded to multiple configured stacks-nodes at once. 

A new environment variable `STACKS_API_EXTRA_TX_ENDPOINTS_FILE` is implemented which can be pointed at a file containing newline-delimitated URLs. When an http POST request is made to `/v2/transactions` it will be concurrently broadcasted to both the original proxied endpoint (e.g. `STACKS_CORE_PROXY_HOST`) and to each of the URLs defined in the file. The response from the original endpoint will proxied back to the client, and any response errors from the others will be logged. 

### Example config
Env var:
```
STACKS_API_EXTRA_TX_ENDPOINTS_FILE=/config/extra-tx-post-endpoints.txt
```
Contents of `extra-tx-post-endpoints.txt`:
```
http://another-node-a/v2/transactions
http://another-node-b/v2/transactions
http://another-node-c/v2/transactions
http://another-node-d/v2/transactions
```

Deployment note: the file is automatically refreshed and can be updated while the API is running, without any service restarts necessary.

### DIFF
--- a/.env
+++ b/.env
@@ -29,9 +29,9 @@ STACKS_CORE_RPC_PORT=20443
 # STACKS_CORE_PROXY_HOST=127.0.0.1
 # STACKS_CORE_PROXY_PORT=20443
 
-# Configure additional stacks-node `POST /v2/tranascation` URLs for the /v2 proxy to mutlicast.
-# Can be a single URL or a comma-separated list of URLs.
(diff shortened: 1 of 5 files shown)


CharlieC3 (approved):
Nice, LGTM

rafaelcr (approved):
Nice 👍

PR #780, opened 2021-09-27
feat: set api version in openapi schema automatically

### LINKED ISSUE #764: set api version in openapi.yaml automatically
The API version exposed by `openapi.yaml` is [currently hardcoded](https://github.com/blockstack/stacks-blockchain-api/blob/develop/docs/openapi.yaml#L13) to `1.0.0`. This is confusing to users [in the API docs](https://blockstack.github.io/stacks-blockchain-api/) because the real API version reported by `/extended/v1/status` looks very different (e.g. `v0.66.1`).

Now that we have the latest version string exposed through the `.git-info` file, we should modify our Github actions so they append that value into `openapi.yaml` before generating the Github pages documentation.

rafaelcr:
## Description

This change writes the current API version (taken from the latest git tag) to the `info.version` entry in `openapi.yaml` so our generated documentation can display the correct current version instead of a hardcoded "1.0.0". Version information is taken from the generated `.git-info` file.

Closes #764 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

These will happen automatically.

## Testing information
Tested locally by generating docs and specs using the built in rules.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


docs/openapi.yaml
```diff
@@ -10,7 +10,7 @@ servers:
     description: Local
 info:
   title: Stacks 2.0 Blockchain API
-  version: '1.0.0'
+  version: 'STACKS_API_VERSION'
```
zone117x, line 13: I suspect this change is what's breaking the vercel CI docs from building -- do you know if this stub is filled by the time the root package `build:docs` command is ran?

zone117x:
I'd defer my PR approval to @CharlieC3 who can more confidently say if these changes will work in CI as intended.

CharlieC3:
> I'd defer my PR approval to @CharlieC3 who can more confidently say if these changes will work in CI as intended.

The `gh-pages` workflow doesn't run as part of the CI process for PRs, but since `generate:docs` is failing in Vercel I'm guessing it'll fail once it runs against `master` as well.

docs/openapi.yaml
rafaelcr: Good point, it's not, I'll fix that 👍

rafaelcr:
Done @CharlieC3 .
What I did was to coalesce the value of `STACKS_API_VERSION` back to `1.0.0` only for Vercel since versioning with git won't be possible there.

Issue #537, opened 2021-04-02
Rosetta api validation tests with rosetta-cli


noumanahmads:
The tests should build and run the rosetta-cli docker-container. It should perform some test transactions (transfer, deploy contract, etc), and then validate the resulting output from rosetta-cli.

These tests are already implemented [here](https://github.com/fariedt/stacks-blockchain-api/tree/feat/tests-rosetta-cli) but failing. 
Changes would be required once the rosetta updates are done.

More details [here](https://github.com/blockstack/stacks-blockchain-api/pull/239)

agraebe:
@noumanahmad001 are you working on this already or not yet?

noumanahmads:
@agraebe not yet.

PR #781, opened 2021-09-28
Rosetta data api and construction validation with rosetta-cli

asimm241:
## Description

The test builds and runs the rosetta-cli docker-container, then it performs some test transactions (transfer, deploy the contract, etc), and then it validates the resulting output from rosetta-cli. _copied from #239_ 

This PR also fixes the following issue with the rosetta implementation

https://github.com/blockstack/stacks-blockchain-api/pull/781/files#r723152208

-> The status will not always be `Success`
-> in the case of `construction/parse`.  We are passing a transaction that is not even on the network yet, it should be an empty status. 

For details follow this issue #537 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Testing information

- npm run devenv:build
- npm run devenv:deploy
- npm run test:rosetta-cli:data
- npm run test:rosetta-cli:construction

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] @zone117x for review


src/tests-rosetta-cli-construction/validate-construction.ts
```diff
@@ -0,0 +1,182 @@
...
+    await transferStx(recipientAdd1, 1000000000, sender1.privateKey, api);
+    await sleep(15000);
+    await transferStx(recipientAdd1, 1000000000, sender1.privateKey, api);
+    await sleep(15000);
+    await transferStx(recipientAdd1, 1000000000, sender1.privateKey, api);
+    await sleep(15000);
+    await transferStx(recipientAdd1, 1000000000, sender1.privateKey, api);
+    await sleep(15000);
+    await transferStx(recipientAdd1, 1000000000, sender1.privateKey, api);
+    await sleep(15000);
```
rafaelcr, line 101: I think you could use the `addressUpdate` datastore event to wait for these transactions instead of sleeping, otherwise the test will take a minimum of 85 seconds to run.

src/tests-rosetta-cli-construction/validate-construction.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    // wait for rosetta-cli to exit
+    let check = true;
+    while (check) {
+      // todo: remove hardcoded container name with dynamic
+      check = await isContainerRunning('/stacks-blockchain-api_rosetta-cli_1');
+      await sleep(2000);
+    }
```
rafaelcr, line 109: Is there a specific reason why this check is here instead of in `afterAll`?

src/tests-rosetta-cli-construction/validate-construction.ts
asimm241: Sorry, this PR was supposed to be in the draft for now.  I'm still working on it, I have just copied the code from the original branch and fixed the bugs. I'm still working on it adding Github actions.

rafaelcr: Sorry I missed the draft status, all good 👍 let me know when it's ready for review

rafaelcr (approved):
Nice, LGTM

zone117x (commented):
@asimm241 do you think any of the files added in the root project directory could be moved into a sub-directory? For example the new `env.*` files?

rosetta-cli-config/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ $? -ne 0 ]; then
+  HOST_IP=$(ip route | awk 'NR==1 {print $3}')
+  echo -e "$HOST_IP\t$HOST_DOMAIN" >>/etc/hosts
+fi
+
+cat /etc/hosts
+echo "patched host.docker.internal"
+
+# wait for things to settle down
+sleep 20
```
zone117x, line 16: Is this needed? This adds a minimum of 20 extra seconds to run this test suite right? I'm not sure why this should be necessary.

zone117x (approved):
Left a question about a `sleep 20` in a docker init script, otherwise this looks good. Great job!

rosetta-cli-config/docker/docker-entrypoint.sh
asimm241: This is not necessary, It must have slipped in while debugging. I have updated it.

Issue #668, opened 2021-07-19
Empty events array in address transaction history endpoint 


yknl:
The events array for transactions from the `/extended/v1/address/${stxAddress}/transactions)` endpoint seems to be empty even though the events count is showing non-zero. 

For example:
Querying `https://stacks-node-api.stacks.co/extended/v1/address/SPR7RRXN8ZRXPNAFKPN2TB84J7T1RMZRGWQQV7P3/transactions`

The transaction with ID `0x32ac0d3cf8cb08b1e9f5bd2509c10b6ad46d955b1cac269a8f19d332ca090a93`
shows that there are 2 events, but the events array is empty.

Querying the transaction endpoint directly shows the correct events.
`https://stacks-node-api.stacks.co/extended/v1/tx/0x32ac0d3cf8cb08b1e9f5bd2509c10b6ad46d955b1cac269a8f19d332ca090a93`

zone117x:
Transaction events are only returned for the tx specific endpoint, and are paginated. Historically all events were returned for all txs in every endpoint, but that resulted in sql queries and JSON responses becoming way too large. In certain cases like the genesis transactions, the results would be several megabytes in size. 

If you happen to be looking for STX transfer related events, see the endpoint `https://stacks-node-api.stacks.co/extended/v1/address/SPR7RRXN8ZRXPNAFKPN2TB84J7T1RMZRGWQQV7P3/transactions_with_transfers`.

yknl:
Ok makes sense. But maybe the empty events array should be removed from the response to avoid confusion.

PR #792, opened 2021-10-07
Remove Empty Array From Address Endpoint

MHassanTariq:
## Description

The transaction's history endpoint returned transaction with event_count non zero. However, the transaction events contained an empty list. This PR removed the unnecessary events object from the response.

For details refer to issue #668 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/docs/generated.d.ts
+++ b/docs/generated.d.ts
@@ -765,13 +765,6 @@ export type TransactionStatus1 = "success" | "abort_by_response" | "abort_by_pos
  * String literal of all Stacks 2.0 transaction types
  */
 export type TransactionType = "token_transfer" | "smart_contract" | "contract_call" | "poison_microblock" | "coinbase";
-export type RpcSubscriptionType =
-  | "tx_update"
-  | "address_tx_update"
-  | "address_balance_update"
-  | "block"
-  | "microblock"
-  | "mempool";
 /**
  * Describes all transaction types on Stacks 2.0 blockchain
  */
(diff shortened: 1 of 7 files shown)


docs/generated.d.ts
```diff
@@ -475,7 +475,7 @@ export type AbstractTransaction = BaseTransaction & {
   /**
    * List of transaction events
    */
-  events: TransactionEvent[];
```
zone117x, line 478: This would make the tx response object a bit annoying to use for endpoints that are known to include the events property. A lot of client code would have to perform null/undefined checks on the property before use.

Would it make sense to have something like a "TransactionWithEvents" schema/type where this isn't optional, and specify that as the return type for endpoints like `/extendex/v1/tx/<txid>`?

src/api/controllers/db-controller.ts
```diff
@@ -170,9 +171,7 @@ export function getTxStatus(txStatus: DbTxStatus | string): string {
 
 type HasEventTransaction = SmartContractTransaction | ContractCallTransaction;
 
-export function getEventTypeString(
-  eventTypeId: DbEventTypeId
-): ElementType<Exclude<HasEventTransaction['events'], undefined>>['event_type'] {
+export function getEventTypeString(eventTypeId: DbEventTypeId) {
```
rafaelcr, line 174: Can you add the new return type?

rafaelcr (changes requested):

src/api/controllers/db-controller.ts
MHassanTariq: There isn't any new return type. However, there are no `events` in the `BaseTransaction`. So, I had to remove the return type from here. 
I have updated the return type, let me know if that looks fine.

rafaelcr (approved):

zone117x (approved):

PR #801, opened 2021-10-13
NodeJS cluster workers (multiple cpu core support)

zone117x:
While working on CPU profiling support, it became apparent that various approaches have an impact on how future nodejs [`cluster`](https://nodejs.org/api/cluster.html#cluster_cluster) support would work. I took a stab at implementing support and it turned out to be pretty straightforward (mostly due to Rafael's readonly mode work).

Clustering is disabled by default, and can be enabled with an env var `STACKS_ENABLE_CLUSTER=true`.

For context, see nodejs docs: https://nodejs.org/api/cluster.html#cluster_cluster
> A single instance of Node.js runs in a single thread. To take advantage of multi-core systems, the user will sometimes want to launch a cluster of Node.js processes to handle the load.
> The cluster module allows easy creation of child processes that all share server ports.

### DIFF
--- a/.env
+++ b/.env
@@ -1,9 +1,7 @@
 PG_HOST=127.0.0.1
-#PG_PORT=5432
 PG_PORT=5490
 PG_USER=postgres
 PG_PASSWORD=postgres
-#PG_DATABASE=postgres
 PG_DATABASE=stacks_blockchain_api
 PG_SCHEMA=public
 PG_SSL=false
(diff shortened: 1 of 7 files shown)


zone117x:
Note: this feature needs testing to ensure the prom metrics are merged and reported correctly between the cluster processes -- right now it uses the default behavior from the prom lib, which from their docs imply a "best effort" implementation.

This feature isn't a high priority right now and not yet worth the testing, so going to icebox the PR.

Issue #579, opened 2021-05-05
Store contract-call function ABI json in tx table


zone117x:
Contract-call tx metadata like the function ABI json should be stored in the tx table for `contract-call` tx types.

Currently the db controller has to perform an additional query for `contract-call` txs to transform the `DbTx` row data into a useful transaction API response object. This introduces transactional consistency bugs due to being two independent sql transaction queries. It's a bottleneck for several areas of the API that need to perform a series of db queries when the `DbTx` objects should be good enough to transform synchronously. 

Affected code:
https://github.com/blockstack/stacks-blockchain-api/blob/37bc66c6b5f9e8a687026810d9b32f2ae104787d/src/api/controllers/db-controller.ts#L651-L678

Additionally, mempool tx parsing just omits this contract-call data for performance and transactional consistency reasons: 
https://github.com/blockstack/stacks-blockchain-api/blob/37bc66c6b5f9e8a687026810d9b32f2ae104787d/src/api/controllers/db-controller.ts#L470-L484

zone117x:
Need to create an issue in the stacks-blockchain repo to request this in the event emitter payload

asimm241:
@zone117x what's the status on this?

zone117x:
I'm moving this to a P4 since this issue hadn't caused serious problems. Also, I don't think we'll be able to get the stacks-node event emitter changes that I last mentioned. 

We could look into adding the contract-call function ABI to the DbTx and DbMempoolTx interfaces, and see if it's reasonable for the related postgres queries to include that data. That would avoid needing to performing ABI lookup queries later on in the parsing code.

PR #807, opened 2021-10-15
Fetch abi json with tx

MHassanTariq:
## Description

Searching a specific transaction required 2 query calls to fetch data for a transaction. First query returned respective transaction details from `txs` table. Second query returned json abi for `contract_call`. This PR makes a single query call to fetch entire data for a transaction. 
The previous search time was `1.183` ms. Current has been reduced to `0.718` ms with 14000+ transactions.

For details refer to issue #579 
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Checklist
- [X] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/docs/entities/transactions/base-transaction.schema.json
+++ b/docs/entities/transactions/base-transaction.schema.json
@@ -49,6 +49,9 @@
     },
     "anchor_mode": {
       "$ref": "./anchor-mode.schema.json"
+    },
+    "abi": {
+      "type": "string"
     }
   }
 }
--- a/docs/entities/transactions/base-transaction.schema.json
+++ b/docs/entities/transactions/base-transaction.schema.json
@@ -49,9 +49,6 @@
     },
     "anchor_mode": {
       "$ref": "./anchor-mode.schema.json"
-    },
-    "abi": {
-      "type": "string"
     }
   }
 }
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -609,6 +609,7 @@ function parseDbBaseTx(dbTx: DbTx | DbMempoolTx): BaseTransaction {
     post_condition_mode: serializePostConditionMode(dbTx.post_conditions.readUInt8(0)),
     post_conditions: postConditions,
     anchor_mode: getTxAnchorModeString(dbTx.anchor_mode),
+    abi: dbTx.abi,
   };
   return tx;
 }
@@ -768,6 +769,7 @@ export function parseDbTx(dbTx: DbTx): Transaction {
   const result: Transaction = {
     ...abstractTx,
     ...txMetadata,
+    abi: dbTx.abi,
   };
   return result;
 }
@@ -779,6 +781,7 @@ export function parseDbMempoolTx(dbMempoolTx: DbMempoolTx): MempoolTransaction {
   const result: MempoolTransaction = {
     ...abstractTx,
     ...txMetadata,
+    abi: dbMempoolTx.abi,
   };
   return result;
 }
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -614,7 +614,7 @@ function parseDbBaseTx(dbTx: DbTx | DbMempoolTx): BaseTransaction {
   return tx;
 }
 
-function parseDbTxTypeMetadata(dbTx: DbTx | DbMempoolTx, abi?: string): TransactionMetadata {
+function parseDbTxTypeMetadata(dbTx: DbTx | DbMempoolTx): TransactionMetadata {
   switch (dbTx.type_id) {
     case DbTxTypeId.TokenTransfer: {
       const metadata: TokenTransferTransactionMetadata = {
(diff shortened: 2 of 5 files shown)


src/api/controllers/db-controller.ts
```diff
@@ -798,14 +797,13 @@ export async function getTxFromDataStore(
...
+    abi = txQuery.result.abi;
   }
 
   const parsedTx = parseDbTx(dbTx);
 
   // If tx type is contract-call then fetch additional contract ABI details for a richer response
-  if (parsedTx.tx_type === 'contract_call') {
-    await getContractCallMetadata(db, dbTx, parsedTx);
-  }
+  parseContractCallMetadata(dbTx, parsedTx, abi);
```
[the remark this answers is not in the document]
MHassanTariq: I tried moving the logic from `parseContractCallMetadata` to `parseDbBaseTx`, however, it has a few problems.
- That approach requires an additional check for `contract_call` in `parseDbBaseTx` which is already implemented in `parseDbTxTypeMetadata`. 
- Moreover, the `function_signatures` and `function_args` are assigned in `parseDbTxTypeMetadata` too.

So, the best suitable place I figured was `parseDbTxTypeMetadata`. Let me know if this approach doesn't look good to you.

zone117x:
Note that we should wait until PR https://github.com/hirosystems/stacks-blockchain-api/pull/858 is merged before this one.

src/datastore/common.ts
```diff
@@ -602,7 +602,7 @@ export interface DataStore extends DataStoreEventEmitter {
     txId: string;
     includeUnanchored: boolean;
     includePruned?: boolean;
-  }): Promise<FoundOrNot<DbMempoolTx>>;
+  }): Promise<FoundOrNot<DbMempoolTx & { abi?: string }>>;
```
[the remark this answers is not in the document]
zone117x: I don't think we need to send an additional `abi` field -- it's the various function signatures where `& { abi?: string }` is appended to an existing type. Seems like the some of those types should include the `abi?: string` property, unless that ends up causing a bunch of other issues.

docs/entities/transactions/base-transaction.schema.json
```diff
@@ -49,6 +49,9 @@
     },
     "anchor_mode": {
       "$ref": "./anchor-mode.schema.json"
+    },
+    "abi": {
```
zone117x, line 53: Related to previous comment, I think this should be reverted

src/api/controllers/db-controller.ts
```diff
@@ -893,19 +897,19 @@ export async function getTxFromDataStore(
...
+  // if (parsedTx.tx_type === 'contract_call') {
+  //   const transaction = await getContractCallMetadata(
+  //     db,
+  //     parseDbTx(dbTx) as ContractCallTransaction,
+  //     dbTx
+  //   );
+  //   if (transaction) {
+  //     parsedTx = transaction as ContractCallTransaction;
+  //   }
+  // }
```
zone117x, line 912: We can just delete these lines right?

zone117x (approved):
Left a few comments but otherwise looks good.

src/api/controllers/db-controller.ts
```diff
@@ -659,25 +660,40 @@ function parseDbTxTypeMetadata(dbTx: DbTx | DbMempoolTx): TransactionMetadata {
         dbTx.contract_call_function_name,
         () => 'Unexpected nullish contract_call_function_name'
       );
+      // paste code here
```
zone117x, line 663: Left over debug code?

src/api/controllers/db-controller.ts
```diff
@@ -659,25 +660,39 @@ function parseDbTxTypeMetadata(dbTx: DbTx | DbMempoolTx): TransactionMetadata {
...
+                  : { name: '', type: undefined };
                 return {
                   hex: bufferToHexPrefixString(serializeCV(c)),
                   repr: cvToString(c),
-                  name: '',
+                  name: functionArgAbi.name,
                   // TODO: This stacks.js function throws when given an empty `list` clarity value.
                   //    This is only used to provide function signature type information if the contract
                   //    ABI is unavailable, which should only happen during rare re-org situations.
                   //    Typically this will be filled in with more accurate type data in a later step before
```
zone117x, line 692: Now that the contract-call parsing logic is here now, shouldn't the PR include removal of the previous functions that used to do this? So the comment 
> Typically this will be filled in with more accurate type data in a later step before being sent to client.

Would no longer be true and should be updated right?

For example, are the following functions needed anymore `parseContractsWithDbTxs(...)`, `parseContractsWithMempoolTxs(...)`, `getContractCallMetadata(...)`?

MHassanTariq: I removed `getContractCallMetadata` as it was not being used. However, maybe due to a merge error it was added again. I'll remove that. 
As far as the `parseContractsWithDbTxs` and `parseContractsWithMempoolTxs` are concerned, this PR was made when a single transaction was retrieved from DB. Feature for multiple transactions was added later. I will update the queries for the multiple transactions feature too. Once done, we won't need the 2 functions mentioned above.

PR #821, opened 2021-10-22
Production-capable CPU profiling

zone117x:
For project overview see: [Stacks API Profiling](https://www.notion.so/hiropbc/Stacks-API-Profiling-c9f4ac602f7547e19f1be96f2b10b79a)

This PR allows an API instance to be configured with private http endpoints that provide CPU profiling capabilities. These allow CPU and memory profiling samples to be easily generated and downloaded (e.g. curl commands) on-demand.

The new profiler http server can be enabled by specifying the `STACKS_PROFILER_PORT` env var.

#### `GET /profile/cpu?duration={seconds}`

Initializes an in-process `inspector` session, enables CPU sampling, then after the given duration, closes down the session and returns the resulting `.cpuprofile` report file in the http response. 

The most effective way to analyze this report is to simply open the file in VSCode which has [built in support](https://github.com/microsoft/vscode-js-profile-visualizer) with tree views and flame charts. 

The report can also be viewed in Chrome by going to the url `chrome://inspect`, selecting "DevTools for Node" > "Profiler" > "Load".

Usage example (curl):
```
curl -OJ localhost:9119/profile/cpu?duration=5
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 40886    0 40886    0     0   8023      0 --:--:--  0:00:05 --:--:-- 10513
curl: Saved to filename 'cpu_1634929163_5-seconds.cpuprofile'
```

#### `GET /profile/heap_snapshot`

*Note: this was an easy additional feature to add, but memory profiling is not within the scope of this project.* Initializes an in-process `inspector` session, triggers a single-pass heap snapshot, closes down the session and returns the resulting `.heapsnapshot` file in the http response. 

The report can be viewed in Chrome by going to the url `chrome://inspect`, selecting "DevTools for Node" > "Memory" > "Load". 

#### `GET /profile/cancel`

Generally just a fail safe to cancel any on-going (likely accidental) long profiling sessions. Note: this could probably be removed, and instead have the report-generation http request handlers observer themselves for an aborted req/res and cancel the report generation themselves.

## Example analysis

Example of evaluating a report using VSCode's default view:

image

And using the flame graph view:

image

Using Chrome's DevTools for Node.js -- tree view:

image


zone117x:
We ran the first test against canary-routed prod traffic (around 15% of API traffic), and the resulting report has already provided invaluable data. A ton of CPU is spent in a areas that would have otherwise been impossible to figure out -- a hashing helper function for c32 address parsing is performing a ton of redundant Node.js `require(...)` calls which are relatively expensive:

image

Pod CPU and memory usage was also unaffected while the profiler was running (blue line here):
image
(the memory usage graph looks the same as above)

rafaelcr (approved):
Very nice

CharlieC3 (approved):
:shipit:

PR #830, opened 2021-11-02
chore: standardize updates between websockets and socket.io

### LINKED ISSUE #815: subscribe to anchor blocks or microblocks via websocket
I have an app that creates swaps that expire on a certain blockheight for this I need to periodically check if there's a new block. 

I'd like to have a way to subscribe to anchor block updates with the websocket and preferably also microblock updates. So Everytime there's a new anchor block or microblock a message would be triggered. I could update my app db and go fetch the block contents if I'm interested in them.

I'm currently querying get_block_list with limit=1 every minute.

rafaelcr:
## Description

This PR standardizes the event updates we send through websockets and socket.io so users of any can receive the same events. Specifically, these changes were made:

**Socket.io**
- Add individual tx updates
- Send microblock metadata along with canonical block updates

**Websockets**
- Add block updates
- Add microblock updates
- Add mempool updates

These two files were also moved to a unified `src/routes/ws/` directory.

Closes #815 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Other

No

Docs are not yet available for these features.

## Testing information
Local testing

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


docs/generated.d.ts
```diff
@@ -3335,6 +3350,57 @@ export interface RpcAddressTxSubscriptionRequest {
   params: RpcAddressTxSubscriptionParams;
   [k: string]: unknown | undefined;
 }
+export interface RpcBlockNotificationResponse {
+  jsonrpc: "2.0";
+  method: "block";
+  params: Block;
+  [k: string]: unknown | undefined;
```
kyranjamie, line 3357: additionalProperties: false Gets rid of the any key type, if it's not needed

rafaelcr: Great idea, done, thanks

rafaelcr:
@kyranjamie @zone117x the current websocket updates for transactions and balances include very little information about the actual event. For example: An address update in websockets only gives users the `tx_id`, `tx_status` and `tx_type`, while socket.io returns a full `AddressTransactionWithTransfers` object with all the details (pretty much saving one extra API call from the user if they want more info).

Do you think I should change these websocket updates to include the full object to be on par with socket.io? I ask because it would be a breaking change for current websocket users.

zone117x (approved):
This looks great, thanks @rafaelcr!

> Do you think I should change these websocket updates to include the full object to be on par with socket.io?

That change makes sense to me, although doesn't have to be done in this PR. Do we have insight into usage for this specific ws event to determine how bad a breaking change would be?

rafaelcr:
Thanks @zone117x , that's a good question and I just found out that we don't have metrics for websocket usage. I'll file a bug to add those and then we can figure out how to upgrade the events.

PR #834, opened 2021-11-05
feat: chaintip-based cache-control, caching with zero stale data

zone117x:
This is an implementation of http-level caching support and validation as described here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#cache_validation

The API always set the cache-control header to "revalidate" -- every HTTP request still hits the API in order to determine if the client (or CDN) cached response is still fresh. In general, this is implemented by specifying http response headers with an `ETag` value and a `Cache-Control` value that tells the client/CDN to ask the server if a cached response is still usable. 

In our case, the API specifies the chain tip hash as the [http response `ETag` value](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag). And sets the http response `Cache-Control` header to [`no-cache`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#requiring_revalidation) (the equivalent of `max-age=0, must-revalidate`).

When a client request is being processed, the API checks for the existence of the [`If-None-Match` http request header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match), which, if the client already has a cached response, should be the chain tip hash that was provided in the previous request `ETag` header. If there hasn't been a new block/microblock then the client's cache is usable, the API will return a `304 Not Modified` response, and the client/CDN will use the cached response. If the API's chain tip has updated, a new response will be sent, along with the updated `ETag` value with the new chain tip. 

Adds chaintip-based cache-control support to the following routes (none of which should be mempool-sensitive):
* `/extended/v1/block`
* `/extended/v1/block/:hash`
* `/extended/v1/block/by_height/:height`
* `/extended/v1/block/by_burn_block_height/:burnBlockHeight`
* `/extended/v1/block/by_burn_block_hash/:burnBlockHash`
* `/extended/v1/tx`

A few new prometheus metrics have also been added:
* `chain_tip_cache_hits`: Total count of requests with an up-to-date chain tip cache header
* `chain_tip_cache_misses`: Total count of requests with a stale chain tip cache header
* `chain_tip_cache_no_header`: Total count of requests that did not provide a chain tip header (over time, will indicate what percentage of clients are not using cache-control-enabled http fetching, or doing something to bypass it at the cloudflare level like appending bogus query params)

TODO:
* [x] unit tests
* [ ] test that CDN (cloudflare) caches correctly with this implementation
* [x] add prometheus metrics for cache hits


CharlieC3:
Sounds/Looks good to me so far! A couple of comments/notes:

1. For requests which a cache-check is performed, invalidated, and fresh data is re-fetched, do you have a rough estimate how much extra time this loop might add to these requests? With extra logic for the API to validate/invalidate the cache, I would assume this would make requests with a validated cache much faster, and requests with an invalidated cache a tiny bit slower on average due to the extra SQL query that has to be ran.
2. Metrics on cache hits are already being collected in Cloudflare, although we have no way of alerting on this. Adding cache prom metrics will most likely be valuable from an alerting angle, but may not be exposing any new metrics (which is ok esp if adding that is an easy win).
3. Will the new prometheus metrics also account for the `/v2` cache hits/misses?
4. What kind of changes should devops expect to make in our infrastructure to support this besides configuring Cloudflare?

rafaelcr:
Looks really good to me as well, I also have a couple of suggestions though:
1. I believe the `blocks` and `microblocks` tables could greatly benefit from a DESC sorted index on `block_height` and `microblock_sequence` respectively so the chain tip lookup is way faster
2. (As discussed on Discord) I believe the `/v1/tx/mempool` endpoint should be excluded from this caching (and maybe others that would be affected by mempool contents)
3. I know you're probably working on this but I think the regexes need a whole bunch of examples to be run on unit tests to make sure we're as bulletpoof as possible

zone117x:
@CharlieC3 
> I would assume this would make requests with a validated cache much faster, and requests with an invalidated cache a tiny bit slower on average due to the extra SQL query that has to be ran.

Correct. Although, given the issues I'm running into with needing more fine-grained control over conditionally caching (as opposed to the current PR's middleware approach), I think this PR may change to re-use the already queried chain-tip performed by the majority of datastore queries behind these requests. Which would mean pretty much zero slow down / overhead from adding this.

> Will the new prometheus metrics also account for the /v2 cache hits/misses?

Not currently, all the caching implemented here doesn't touch the /v2 proxy. Also, the most expensive /v2 requests are POSTs AFAIK, which wouldn't really work with this cache-control behavior anyway.

> What kind of changes should devops expect to make in our infrastructure to support this besides configuring Cloudflare?

Assuming sane behavior from cloudflare, none!


@rafaelcr up agreed on the desc block_height index, it's already the most common query (and this feature will just increase that). 

Express's middleware architecture does not lend itself well to the kind of generalized approach currently implemented in this PR (set some caching headers before the handlers for various routes). We need to only cache on 200 responses. Also, in many route handlers, detect if a payload includes mempool tx data and avoid caching -- code that is specific per each of these kinds of routes.

Middleware implemented before the route handlers doesn't have knowledge of the above conditions in order to perform the correct cache logic. Middleware implemented _after_ the route handlers is too late, because response headers will have already been sent. Express does not have a hook system like "onBeforeSendHeaders". Looking around at other caching approaches, it looks like the "express way" of doing this involves adding a bit of code to each route. Unfortunate because this adds some additional maintenance burden for creating/modifying routes to ensure it will have correct caching behavior, and just more lines of code. Still seems reasonable though.

zone117x:
Update: many of the endpoints can check the mempool, which means a "state hash" of the mempool needs to be used in conjunction with the chain-tip. This can be done with a query like
```sql
SELECT digest(
(
    SELECT array_agg(tx_id)
    FROM (
        SELECT tx_id
        FROM mempool_txs
        WHERE pruned = false
        ORDER BY receipt_time DESC
    ) m
)::text, 'sha256')
```
Could also be a materialized view that is refreshed on mempool table updates. 

In the mean time, it probably makes sense to do an initial pass to support endpoints that only need chaintip.

src/api/routes/block.ts
```diff
@@ -18,92 +20,119 @@ const parseBlockQueryLimit = parseLimitQuery({
 
 export function createBlockRouter(db: DataStore): RouterWithAsync {
   const router = addAsync(express.Router());
+  const cacheHandler = getChainTipCacheHandler(db);
+  router.get(
+    '/',
+    cacheHandler,
+    asyncHandler(async (req, res) => {
```
zone117x, line 27: Note for reviewers: we've been using `router.getAsync`, an Express extension provided by the [`@awaitjs/express` lib](https://www.npmjs.com/package/@awaitjs/express). However, the `getAsync` function has incorrect/bugged behavior when given multiple router handler functions. It executes both synchronously at the same time. This breaks the ability for the handlers registered first to control the route flow -- in the case of this PR, prevents caching handler from being able to skip the non-caching handler.

In addition, the `@awaitjs/express` lib is incorrectly typed and is forcing this repo to disable linting for `no-misused-promises`. The author of that lib has rejected PRs attempting to fix the typing.

zone117x:
PR ready for review (also see my last comment above regarding mempool-sensitive routes).

Chaintip-based cache-control has been implemented on all the `/extended/v1/block/*` endpoints for now -- see list of updated routes in PR description. 

Several more routes can also be supported, but will wait for some PR feedback before adding to them. Same with extensive unit tests -- will wait on PR feedback first.

src/api/controllers/cache-controller.ts
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * Sets the response `Cache-Control` and `ETag` headers using the chain tip previously added
+ * to the response locals.
+ * Uses the latest unanchored microblock hash if available, otherwise uses the anchor
+ * block index hash.
+ */
+export function setChainTipCacheHeaders(res: Response) {
+  const chainTip: FoundOrNot<DbChainTip> | undefined = res.locals[CHAIN_TIP_LOCAL];
+  if (!chainTip) {
+    logger.error(
```
rafaelcr, line 53: Can this be a warning instead of an error? I guess we'll hit this path very often when the cache tip is still being generated, right?

src/api/controllers/cache-controller.ts
```diff
@@ -0,0 +1,193 @@
...
+      `Cannot set cache control headers, no chain tip was set on \`Response.locals[CHAIN_TIP_LOCAL]\`.`
+    );
+    return;
+  }
+  if (!chainTip.found) {
```
rafaelcr, line 58: We can merge this conditional above (if the above can be just a warning)

src/api/controllers/cache-controller.ts
zone117x: This condition would only be hit from buggy cache-control code. There should be an object at `res.locals[CHAIN_TIP_LOCAL]` even if it's a `found: false` chain tip.

src/api/controllers/cache-controller.ts
zone117x: This condition can be reached normally if the node hasn't synced a block yet. The condition above is checking for an incorrect `res.locals` state.

rafaelcr (approved):
This looks very good to me. It makes sense to leave the mempool-sensitive endpoints to a subsequent PR as you mentioned so we can start trying this out right away 👍

zone117x:
I've added a few more unit tests, and created issues for some of the TODOs/comments I've listed in here. Merging the currently implement cache-control support to develop for initial testing before expending support to more routes.

Issue #828, opened 2021-11-01
Implement Postgres materialized views to pre-calculate heavy queries and improve performance


rafaelcr:
Given that the Stacks blockchain and our Hiro tools have seen a huge spike in use recently, we've had a growing number of performance issues in our mainnet API pods. Specifically, some of the heaviest DB queries have used a lot of CPU capacity, causing performance bottlenecks that affect API response times dramatically. Even though we've optimized some of these queries in the past few weeks and gotten great performance gains, the CPU trend indicates that these improvements won't last for much longer given the increasing volume in user load. There are not many table indexes we could add and even doing so would give us improvements which are marginal at best.

To give an example, some of the most popular uses of our APIs right now is to query NFT ownership by some specific smart contracts (e.g. marketplaces). These contracts have tens of thousands of entries in the `nft_events` table (so far), making these queries extremely heavy to run. The nature of these queries also requires us to do complex JOINs to figure out the last owner of an asset which in itself is another heavy sub-query, etc.

As a solution, we could start creating and using [Postgres materialized views](https://www.postgresql.org/docs/12/rules-materializedviews.html) to pre-calculate heavy queries when we receive events from the Stacks node. They could then be indexed to make queries against them trivial to run with very fast response times.

Some examples of views we could use are:
- Minted NFTs with their current owner address
- Address balances (STX, FT, NFT balances)
- Asset events by address

cc @zone117x @CharlieC3

rafaelcr:
Related to #816

PR #835, opened 2021-11-05
Add nft_custody pg materialized view to speed up nft event lookup

rafaelcr:
## Description

This PR adds a Postgres materialized view `nft_custody` that stores the current owner address of an NFT (a combination of `asset_identifier` and `value`) at the **current canonical chain tip**. It also adds indexes so lookups by NFT or owner address is very quick.

After using this view, the current NFT lookup query sees a huge increase in performance. Local tests showed an improvement from ~400ms before to ~35ms after.

Related to #828 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Other

No

No

## Testing information
Local performance tests

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -582,11 +582,17 @@ export class PgDataStore
   implements DataStore {
   readonly pool: Pool;
   readonly notifier?: PgNotifier;
-  private constructor(pool: Pool, notifier: PgNotifier | undefined = undefined) {
+  readonly eventReplay: boolean;
+  private constructor(
+    pool: Pool,
+    notifier: PgNotifier | undefined = undefined,
+    eventReplay: boolean = false
+  ) {
     // eslint-disable-next-line constructor-super
     super();
     this.pool = pool;
     this.notifier = notifier;
+    this.eventReplay = eventReplay;
   }
 
   /**
@@ -1214,7 +1220,7 @@ export class PgDataStore
             await this.updateNamespaces(client, entry.tx, namespace);
           }
         }
-        if (newNftEvents) {
+        if (newNftEvents && !this.eventReplay) {
           await client.query(`REFRESH MATERIALIZED VIEW nft_custody`);
         }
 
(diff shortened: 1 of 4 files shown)


src/datastore/postgres-store.ts
```diff
@@ -1217,6 +1219,9 @@ export class PgDataStore
             await this.updateNamespaces(client, entry.tx, namespace);
           }
         }
+        if (newNftEvents) {
+          await client.query(`REFRESH MATERIALIZED VIEW nft_custody`);
```
asimm241, line 1223: @rafaelcr shouldn't this refresh for macroblocks updates also? https://github.com/hirosystems/stacks-blockchain-api/blob/b9a8c4a60044e2713c882df5326acca42dfd5bc2/src/datastore/postgres-store.ts#L1409

rafaelcr: Right now the materialized view only keeps custody status at the canonical chain tip. However, this is a good question that I think @zone117x can answer: if a user requests NFT events with `includeUnanchored` set to true, I would need to ignore the materialized view and do the whole query with the latest non-canonical chain tip, correct? I think this is something I'd need to fix in the current code.

src/datastore/postgres-store.ts
```diff
@@ -1217,6 +1219,9 @@ export class PgDataStore
             await this.updateNamespaces(client, entry.tx, namespace);
           }
         }
+        if (newNftEvents) {
```
zone117x, line 1222: If possible we should skip this step during event-import (otherwise it would be slowed down dramatically), and only perform the refresh at the end of the replay.

src/migrations/1636130197558_nft_custody.ts
```diff
@@ -0,0 +1,25 @@
...
+      nft_events
+    WHERE
+      canonical = true AND microblock_canonical = true
+    ORDER BY
+      asset_identifier DESC,
+      value DESC,
+      block_height DESC,
+      microblock_sequence DESC,
+      tx_index DESC,
+      event_index DESC
```
zone117x, line 20: Note: this should work in returning only anchor block data _if_ this view is only ever refreshed directly after handling an anchor block (there will be no unanchored data right after that).

src/datastore/postgres-store.ts
rafaelcr: Done @zone117x LMK what you think about this approach

asimm241 (approved):

zone117x (approved):
Awesome, thanks @rafaelcr!

PR #839, opened 2021-11-08
fix: transaction broadcast log

### LINKED ISSUE #838: /v2/transactions tx broadcast log is skipped when tx multicast is enabled
The `/v2/transactions` proxy is supposed to log whenever a transaction is broadcasted to the node during forwarding, however, when node multicast is enabled by the use of `STACKS_API_EXTRA_TX_ENDPOINTS_FILE`, the codepath that logs the tx is skipped (as well as some `cache-control` header management).

The current theory is that the multicast middleware function needs to call `next()` when finished so the rest of the middleware chain can be executed.

rafaelcr:
## Description

This patch fixes the transaction broadcast logs when an API pod is set to multicast.

Closes #838 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Testing information
Local tests

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/routes/core-node-rpc-proxy.ts
+++ b/src/api/routes/core-node-rpc-proxy.ts
@@ -220,6 +220,12 @@ export function createCoreNodeRpcProxyRouter(db: DataStore): express.Router {
       proxyResp.headers.forEach((value, name) => {
         res.setHeader(name, value);
       });
+      if (proxyResp.status === 200) {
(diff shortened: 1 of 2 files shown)


zone117x (approved):
LGTM!

CharlieC3 (approved):
:shipit:

PR #841, opened 2021-11-09
feat: stx addr encoding LRU cache

zone117x:
Latest CPU profiling results show significant amount of time is spent in the native sha256 digest, and js c32 encoding code for transforming raw Clarity values into Stacks address strings. 

This PR implements an in-memory LRU cache to keep a map of bytes to STX address strings. 

There are several code paths that end up performing this encoding, so it's not as simple as switching some code in the API to call a new function. For example, decoding a Clarity value into a repr string, decoding a tx, etc can all end up calling into the `c32check` lib. So this PR takes an approach of overriding the function on that module with a function that uses an LRU cache.

Current implementation specifies a max cache size of 50k entries, ~~which napkin math shows something like 5-10 MB of additional memory usage -- needs tested~~.

EDIT: added a benchmark script, and runs consistently show around ~4.5 megabytes of memory used per 10,000 unique STX addresses cached.

### Usage
The default max cache size is 50,000 (should be around 25 megabytes of additional memory usage). This can be specified with the an environment variable `STACKS_ADDRESS_CACHE_SIZE`. 

### TODO:
* [x] unit tests
* [x] memory usage tests
* [x] cpu perf tests with prod canary traffic routing

image


rafaelcr:
This looks very good so far.

A concern I have, however, is how this will perform in a scaled-up environment where we have several pods responding to requests. Once we have more and more pods I assume the gains would be minimal given that subsequent requests from a client could hit different pods (unless we enable sticky sessions?). What are your thoughts on this @zone117x ?

zone117x:
It's hard to tell with current metrics, but my assumption is that address distribution follows something like the 1% rule or the Pareto principle .e.g `90% of the participants of a community only consume content, 9% of the participants change or update content, and 1% of the participants add content`.

Assuming we aren't wiping the API pod memory super often (like more often than every hour), I think every pod will pretty quickly generate a cache of the majority of address encodings used through most http requests. It's possible we might need something like a weight-aware LRU cache (as opposed to the simple LRU cache in this PR) to account for usage that doesn't fit a normal distribution (perhaps caused by many unique wallet users, idk). But, I think this approach is good enough for now, and we can add some cache usage/eviction metrics to tell us otherwise.

zone117x:
This is ready for review and canary prod testing. Local benchmarks of memory usage look pretty good to me (~4.5 megabytes of memory used per 10,000 unique STX addresses cached) -- although I'm curious if @CharlieC3 also finds that acceptable. 

Also, this kind of in-memory caching could possible be done better with redis or postgres, but that would be a larger undertaking due to the async vs sync differences. I'd estimate at least 100 LoC changes throughout stacks.js and the API to enable asynchronous resolving throughout all the areas that encode a STX address.

CharlieC3:
> ~3.6 Roentgen~ ~4.5 megabytes of memory used per 10,000 unique STX addresses cached

Not great, not terrible. But given the severity shown in the flame graph, and the low amount of memory the API consumes today, I see this as a compromise in the user's favor. So if we kept the default of 50,000 STX addresses cached, that would bring the total memory usage of a single API to around 110MB, assuming it's in a load-balanced pool of other APIs.

Once cached, do the unique addresses have a cache expiration?

zone117x:
> Once cached, do the unique addresses have a cache expiration?

No, the least used addresses are only evicted if the cache is full. The lib supports datetime cache expiration, although it probably increases the memory usage.

rafaelcr (approved):
Very cool 🚀

zone117x:
@CharlieC3 and I ran CPU profiling tests with this branch against current release, and PR appears to be working as expected. The CPU time spent on STX / c32 encoding has been drastically reduced.

image

CharlieC3:
That's awesome, I am a bit surprised to see however that the memory consumption for the API nodes running this branch don't consume more memory than API nodes running `0.71.0`. Not sure if we miscalculated something in our extra memory usage estimates, or if this is "too good to be true" and we should be skeptical.

@zone117x Any concerns about this?

zone117x:
@CharlieC3 my guess is unique number of stx addresses being accessed is pretty low, probably around 1k if I had to guess. Which means it will probably take quite a while to see the cache and memory usage increase.

CharlieC3 (approved):
:shipit:

PR #842, opened 2021-11-10
fix: build/publish Dockerfile python installation issues

rafaelcr:
## Description

This patch fixes the Build-publish Github action which started to fail during the `python` installation.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [x] Other

No

No

## Testing information
N/A

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM node:14-alpine
 WORKDIR /app
 COPY . .
 
-RUN apk add --no-cache --virtual .build-deps alpine-sdk python git openjdk8-jre
+RUN apk add --no-cache --virtual .build-deps alpine-sdk python2 git openjdk8-jre
 RUN echo "GIT_TAG=$(git tag --points-at HEAD)" >> .env
 RUN npm config set unsafe-perm true && npm install && npm run build && npm prune --production
 RUN apk del .build-deps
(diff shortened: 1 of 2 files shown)


CharlieC3 (approved):
LGTM! Any idea what's actually using Python here?

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:14-alpine
 WORKDIR /app
 COPY . .
 
-RUN apk add --no-cache --virtual .build-deps alpine-sdk python git openjdk8-jre
+RUN apk add --no-cache --virtual .build-deps alpine-sdk python2 git openjdk8-jre
```
zone117x, line 6: Curious what the cause of this break was? An breaking change in the latest alpine image? If so, seems like a good opportunity to pin the docker image major versions to prevent breaking changes like these. E.g. `FROM node:14-alpine` to `FROM node:14-alpine3.13`

CharlieC3: I'm guessing the `python` package alias was removed from `https://pkgs.alpinelinux.org/`.
Also @rafaelcr, I think the `python` alias pointed to `python3` rather than `python2`. Not sure how much it matter in this project though. It may have only been needed to install `git`, for example.

rafaelcr: Whoops... thanks @CharlieC3 I'll keep that in mind. When I do the build publish v2 PR I'll revisit these Dockerfiles to see if we can version them better and adjust the python version (or remove it if not needed).

PR #843, opened 2021-11-10
docs: expanded docs for faucet API endpoints

pgray-hiro:

## Description

This PR adds detailed descriptions and example request bodies for the endpoints with the Faucet tag. It also adds a description of the Faucet tag in the OpenAPI spec.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

No

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


docs/openapi.yaml
```diff
@@ -14,30 +14,59 @@ info:
...
+      obtaining tokens to test transactions, exercise smart contracts, or perform network operations such as stacking or
+      mining on the testnet.
 
 paths:
   /extended/v1/faucets/stx:
     parameters:
       - name: address
         in: query
-        description: STX address
+        description: A valid testnet STX address
```
agraebe, line 31: we could point out that these addresses have to start with `ST`

docs/api/faucet/run-faucet.schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "description": "POST request that runs the faucet",
+  "description": "POST request that adds tokens to a specified testnet address",
```
agraebe, line 3: nit: could say that this will _initiate a transfer_ of tokens to the account - to convey that there's a waiting period of a few minutes for the transaction to be mined

docs/openapi.yaml
```diff
@@ -48,21 +77,38 @@ paths:
               example:
                 $ref: ./api/faucet/run-faucet.example.json
         500:
-          description: Failed to run faucet
+          description: Faucet call failed
```
agraebe, line 80: nit: could say that one potential reason might be that the faucet ran out of tokens?

agraebe (approved):
left some minor comments; approving so you can identify if/what could be tweaked and move on to merge it when ready

docs/openapi.yaml
```diff
@@ -14,30 +14,59 @@ info:
 
     It is comprised of two parts; the Stacks Blockchain API and the Stacks Core API.
 
-    <a href="collection.json" download="collection.json">Download Postman collection</a>
+    <a href="https://hirosystems.github.io/stacks-blockchain-api/collection.json" download="collection.json">Download Postman collection</a>
```
rafaelcr, line 17: I assume you're setting the absolute URL because it's now hosted on the new docs site, right?
If so, you should also set the URL in the `download` attribute.

docs/openapi.yaml
pgray-hiro: This field is meant to hold a very short string, it isn't meant for a long description. I can add that information in the description of the endpoint, however.

docs/openapi.yaml
pgray-hiro: Again, this field is meant for a short string. I can add this information in the endpoint description.

docs/openapi.yaml
pgray-hiro: The value on the download attribute is actually optional and all it does is specify what the filename will be when it's downloaded. Since we're not changing the filename, I'll just remove it and save the space.

rafaelcr: I see. I don't think you should remove it, though. IIRC when I was adding that link I tested on Safari and Chrome and without the `download` attribute, the JSON file was opened as a text file in the current tab rather than a file download.

pgray-hiro: Right - I'm only removing the value on the attribute. The final line of HTML looks like this `<a href="https://hirosystems.github.io/stacks-blockchain-api/collection.json" download>Download Postman collection</a>`

rafaelcr: Ah, perfect 👍

zone117x (approved):

Issue #651, opened 2021-07-09
Include entity metadata in search endpoint responses


zone117x:
Right now the search endpoint returns an array of matches that only include what the search result entity type is (e.g. a block, tx, address, etc). Consumers like the explorer show additional metadata in the result display, which currently requires several more network requests. 

It seems reasonable to include the metadata in the initial search response. Should be implemented as an opt-in (probably via query params).

PR #848, opened 2021-11-15
Include Metadata

MHassanTariq:
## Description

The purpose of this PR is to exclude multiple API calls for search from stacks explorer.

For details refer to issue #651 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -57,6 +57,7 @@ import {
   DbTxStatus,
   DbTxTypeId,
   DbSmartContract,
+  DbSearchResultWithMetadata,
 } from '../../datastore/common';
 import {
   unwrapOptional,
@@ -1121,4 +1122,63 @@ export async function searchTx(
       return { found: false };
     }
   }
+}
+
+export async function searchHashWithMetadata(
+  hash: string,
+  db: DataStore
+): Promise<FoundOrNot<DbSearchResultWithMetadata>> {
+  // checking for tx
+  const txQuery = await db.getTxListDetails({ txIds: [hash], includeUnanchored: true });
+  if (txQuery.length > 0) {
+    // tx found
+    const tx = txQuery[0];
+    return {
+      found: true,
+      result: {
+        entity_type: 'tx_id',
+        entity_id: tx.tx_id,
+        entity_data: tx,
+      },
+    };
+  }
+  // checking for mempool tx
+  const mempoolTxQuery = await db.getMempoolTxs({
+    txIds: [hash],
+    includeUnanchored: true,
+    includePruned: true,
+  });
+  if (mempoolTxQuery.length > 0) {
+    // mempool tx found
+    const mempoolTx = mempoolTxQuery[0];
+    return {
+      found: true,
+      result: {
+        entity_type: 'mempool_tx_id',
+        entity_id: mempoolTx.tx_id,
+        entity_data: mempoolTx,
+      },
+    };
+  }
+  // checking for block
+  const blockQuery = await db.getBlockWithMetadata({ hash }, { txs: true, microblocks: true });
+  if (blockQuery.found) {
+    // block found
+    const result = parseDbBlock(
+      blockQuery.result.block,
+      blockQuery.result.txs.map(tx => tx.tx_id),
+      blockQuery.result.microblocks.accepted.map(mb => mb.microblock_hash),
+      blockQuery.result.microblocks.streamed.map(mb => mb.microblock_hash)
+    );
+    return {
+      found: true,
+      result: {
+        entity_type: 'block_hash',
+        entity_id: result.hash,
+        entity_data: result,
+      },
+    };
+  }
+  // found nothing
+  return { found: false };
 }
--- a/src/api/routes/search.ts
+++ b/src/api/routes/search.ts
@@ -19,6 +19,7 @@ import {
   ContractSearchResult,
   AddressSearchResult,
   SearchErrorResult,
+  AddressStxBalanceResponse,
 } from '@stacks/stacks-blockchain-api-types';
 import {
   getTxTypeString,
@@ -183,6 +184,9 @@ export function createSearchRouter(db: DataStore): RouterWithAsync {
                 },
               },
             };
+            if (includeMetadata) {
+              contractResult.result.metadata = parseDbTx(txData);
+            }
             return contractResult;
           } else {
             // Associated tx is a mempool tx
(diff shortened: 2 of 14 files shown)


src/datastore/postgres-store.ts
```diff
@@ -5503,13 +5504,16 @@ export class PgDataStore
         `SELECT ${TX_COLUMNS} FROM txs WHERE tx_id = $1 LIMIT 1`,
         [hexToBuffer(hash)]
       );
+      // const tx = await this.getTxListDetails({ txIds: [hash], includeUnanchored: false });
       if (txQuery.rowCount > 0) {
         const txResult = this.parseTxQueryResult(txQuery.rows[0]);
+        // const txResult = tx[0];
```
asimm241, line 5510: can you remove extra comments?

asimm241 (changes requested):

asimm241:
@MHassanTariq The PR needs a rebase because the tests are failing.

src/datastore/postgres-store.ts
```diff
@@ -5511,6 +5513,66 @@ export class PgDataStore
     });
   }
 
+  async searchHashWithMetadata({
```
asimm241, line 5516: don't you think this method should be in db controller?

src/tests/api-tests.ts
```diff
@@ -1745,8 +1745,30 @@ describe('api tests', () => {
     };
     await db.updateMempoolTxs({ mempoolTxs: [mempoolTx] });
 
+    const blockMetadata = {
```
asimm241, line 1748: can you please write another test to include metadata instead of modifying the existing tests?

asimm241 (changes requested):

src/tests/api-tests.ts
```diff
@@ -1877,6 +1877,235 @@ describe('api tests', () => {
...
+      burn_block_height: 123,
+      miner_txid: '0x4321',
+      canonical: true,
+      execution_cost_read_count: 0,
+      execution_cost_read_length: 0,
+      execution_cost_runtime: 0,
+      execution_cost_write_count: 0,
+      execution_cost_write_length: 0,
+    };
+    await db.updateBlock(client, block);
```
rafaelcr, line 1900: Can you change these direct updates (updateBlock, updateTx, etc.) for regular `update()` calls?

src/api/controllers/db-controller.ts
```diff
@@ -1111,3 +1112,62 @@ export async function searchTx(
...
+        entity_id: mempoolTx.tx_id,
+        entity_data: mempoolTx,
+      },
+    };
+  }
+  // checking for block
+  const blockQuery = await db.getBlockWithMetadata({ hash }, { txs: true, microblocks: true });
+  if (blockQuery.found) {
+    // block found
+    const result = parseDbBlock(
```
rafaelcr, line 1156: Why not just return the complete `DbBlock` here and let the caller do the parsing? Similar to the other types of results. Seems strange that you're only returning a few things about the block to the user instead of everything we have.

rafaelcr (changes requested):

src/api/controllers/db-controller.ts
MHassanTariq: Basically, I'm replicating the exact working of the search endpoint. For the case of a block, the endpoint `/extended/v1/search/{id}` responds with `entity_type` as `block_hash`. The client then fetches block data of type `Block` from another endpoint `  /extended/v1/block/{hash}`.
Here, I'm returning the `Block` type block data to the client directly in case it is requested with `include_metadata`.

src/api/routes/search.ts
```diff
@@ -192,11 +271,11 @@ export function createSearchRouter(db: DataStore): RouterWithAsync {
     };
   };
 
-  router.getAsync('/:term', async (req, res) => {
+  router.getAsync('/:term', async (req, res, next) => {
     const { term: rawTerm } = req.params;
+    const includeMetadata = booleanValueForParam(req, res, next, 'include_metadata');
```
zone117x, line 276: Should `include_metadata` be added to the openapi schema for this route?

src/api/routes/search.ts
```diff
@@ -41,9 +56,35 @@ export function createSearchRouter(db: DataStore): RouterWithAsync {
...
+                  canonical: blockData.canonical,
+                  hash: blockData.hash,
+                  parent_block_hash: blockData.parent_block_hash,
+                  burn_block_time: blockData.burn_block_time,
+                  height: blockData.height,
+                },
+                metadata: blockData,
+              },
+            };
+            return blockResult;
```
zone117x, line 86: Any reason why the `block` result for `includeMetadata` is different than the others (this one creates a different object rather than assigning `metadata` for the existing one)?

src/tests/api-tests.ts
```diff
@@ -1912,6 +1912,255 @@ describe('api tests', () => {
...
+          sender_address: 'sender-addr',
+          sponsored: false,
+          tx_id: '0x8912000000000000000000000000000000000000000000000000000000000000',
+          tx_status: 'success',
+          tx_type: 'coinbase',
+        },
+      },
+    };
+    expect(JSON.parse(searchResult4.text)).toEqual(expectedResp4);
+
```
zone117x, line 2133: Can you add a test for a regular transaction? Looks like that's the only search result type missing a test.

zone117x (approved):
Left a couple comments to address, otherwise looks good!

src/api/routes/search.ts
MHassanTariq: wrt my [this comment](https://github.com/hirosystems/stacks-blockchain-api/pull/848#discussion_r760140530) the endpoint `extended/v1/block/{hash}` [responds](https://github.com/hirosystems/stacks-blockchain-api/blob/develop/docs/openapi.yaml#L472-L500) with a `Block` type object with metadata. This is fetched from the [`getBlockWithMetadata`](https://github.com/hirosystems/stacks-blockchain-api/blob/develop/src/datastore/postgres-store.ts#L2554-L2616). However, the search endpoint returns a `DbBlock` object which does not provide information sufficient enough to create enriched block object.

src/tests/api-tests.ts
MHassanTariq: Sure, will do that.

rafaelcr (approved):
👍

Issue #816, opened 2021-10-21
Address transactions endpoint causes database CPU spike in API pods


rafaelcr:
We're seeing latency spikes affecting several mainnet API pods caused by recurrent calls to the `/extended/v1/address/:stx_address/transactions` endpoint.

We should look into optimizing the complex query used by that endpoint and/or look for a table index we could add to speed it up.

PR #849, opened 2021-11-15
feat: add latest smart contract txs materialized view

rafaelcr:
## Description

This PR creates a new postgres materialized view `latest_contract_txs` which stores the latest 50 transactions for every deployed smart contract. Only the last 50 are stored because based on usage logs for the past couple of weeks, the worst performing /v1 endpoint requests we get *by far* (2-5 seconds to complete on avg) are to get the latest ~30-50 transactions for the most popular smart contracts (marketplaces, citycoins, etc.).

Once this view is applied, we should see massive performance upgrades for the `/extended/v1/address/:addr/transactions` endpoint.

From local tests against the `SP000000000000000000002Q6VF78.bns` contract, the original query took ~380ms to return the last 50 transactions. After using the view, the same results now take ~5 ms to be ready.

#816 
#828 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

No

## Testing information
Local performance tests and unit tests.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


rafaelcr:
@zone117x picking up our Discord convo, it turns out the missing pg notification problem was that we weren't waiting for those notifications to fire before finishing the test.

I'm not positive about this, but I think there could be a scenario in which we run a heavy query and then issue a notification, and if somehow we don't finish the runloop correctly that notification might be lost to our clients. I'm adding `await` to all of our notification emissions, but please let me know your thoughts on this. AFAIK this won't add any significant delay to our endpoints.

src/datastore/postgres-store.ts
```diff
@@ -5099,33 +5128,48 @@ export class PgDataStore
...
-        FROM principal_txs
-        ${atSingleBlock ? 'WHERE block_height = $4' : 'WHERE block_height <= $4'}
-        ORDER BY block_height DESC, microblock_sequence DESC, tx_index DESC
-        LIMIT $2
-        OFFSET $3
-        `,
+        // Smart contracts with a very high tx volume get frequent requests for the last N tx, where N
+        // is commonly <= 50. We'll query a materialized view if this is the case.
+        principal.type == 'contractAddress' && !atSingleBlock && args.limit + args.offset <= 50
+          ? `
```
zone117x, line 5139: Nit: my preference here would be to break this out of a ternary expression and have an explicit `useMaterializedView: boolean`

zone117x (approved):
Small code style nit, otherwise LGTM. Awaiting the pg-notifies seems okay to me.

Issue #852, opened 2021-11-16
Error in tx broadcast logging


zone117x:
The tx broadcast logging implemented in the RPC proxy for `POST /v2/transactions` has critical bug when used with tx multicasting:
```
SyntaxError: Unexpected token d in JSON at position 1
    at JSON.parse (<anonymous>)
    at logTxBroadcast (/app/src/api/routes/core-node-rpc-proxy.ts:156:23)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at /app/src/api/routes/core-node-rpc-proxy.ts:227:9
    at wrappedMiddleware (/app/node_modules/@awaitjs/express/index.js:62:7)
```

The following line uses `<fetch result>.json()` which strips the double quotes from the response `"<txid>"`
https://github.com/hirosystems/stacks-blockchain-api/blob/0a485a3db35f783652ccfebe9139503a80c8c377/src/api/routes/core-node-rpc-proxy.ts#L226-L227

then calls `logTxBroadcast(txid)`, which attempts to strip the quotes again with `JSON.parse(response)`, and throws an error because it's already been stripped, and the raw string is invalid json:
https://github.com/hirosystems/stacks-blockchain-api/blob/0a485a3db35f783652ccfebe9139503a80c8c377/src/api/routes/core-node-rpc-proxy.ts#L150-L156

PR #853, opened 2021-11-16
fix: tx broadcast logging error

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/852

Implemented the fix at a few layers for robustness:
A. Use `<fetch result>.text()` rather than `<fetch resuilt>.json()` to get the unaltered response (similar to the non-tx-multicasting behavior).
B. Use a regex to strip the double quotes, rather than `JSON.parse()`, which allows for them to be optional `"123"` -> `123`, or `123` -> `123` (unaltered, no error).
C. Wrap the logging function in a try/catch/log-error so that any potential future errors in this function (e.g. resolving the current block height) won't result in an error response to the client.

Updated the unit test so that this code path is covered. This logging path is skipped when there are no blocks in the db, which was the case for the tx multicast unit tests.


rafaelcr (approved):
My bad, thanks for the quick fix 🙏

Issue #850, opened 2021-11-16
Error reading contract ABI


zone117x:
Seeing a critical error in parsing tx responses:
`/extended/v1/tx/0x8407751d1a8d11ee986aca32a6459d9cd798283a12e048ebafcd4cc7dadb29af`

```
TypeError: Cannot read property 'functions' of null
    at parseContractCallMetadata (/app/src/api/controllers/db-controller.ts:996:35)
    at /app/src/api/controllers/db-controller.ts:925:27
    at Array.forEach (<anonymous>)
    at parseContractsWithDbTxs (/app/src/api/controllers/db-controller.ts:922:13)
```

The contract ABI json object is null here: https://github.com/hirosystems/stacks-blockchain-api/blob/285632a983f2ae83d4f6250c339898dda75bb14e/src/api/controllers/db-controller.ts#L995

Seems like this bug was introduced in https://github.com/hirosystems/stacks-blockchain-api/pull/722

@MHassanTariq can you take a look?

MHassanTariq:
Sure, I'll look into it.

MHassanTariq:
@zone117x the `abi` is [required](https://github.com/hirosystems/stacks-blockchain-api/blob/develop/src/datastore/common.ts#L232) in `DbSmartContract`. Therefore, I ignored this edge case. 
I will put a check for `abi` here. Thanks for pointing it out!

asimm241:
@zone117x The `abi` is not supposed to be null here and this endpoint does not always fail doesn't this means the issue is not on the API side?

zone117x:
Here's some db data for the smart contract referenced by the above tx `0x8407751d1a8d11ee986aca32a6459d9cd798283a12e048ebafcd4cc7dadb29af`:

```sql
SELECT canonical, microblock_canonical, encode(tx_id, 'hex') tx_id, (abi = 'null') has_abi
FROM smart_contracts
WHERE contract_id = 'SP3YK7KWMYRCDMV5M4792T0T7DERQXHJJGGEPV1N8.pg-mdomains-v1'
```
```
 canonical | microblock_canonical |                              tx_id                               | has_abi 
-----------+----------------------+------------------------------------------------------------------+---------
 t         | t                    | 668142abbcabb846e3f83183325325071a8b4882dcf5476a38148cb5b738fc83 | f
 t         | t                    | ab599004a2c4b5a11010b90ee44dfbdac4ec6c8c673f8feed3fb78026625e10e | t
[...]

PR #858, opened 2021-11-17
Error reading contract abi

MHassanTariq:
## Description

This PR fixes the bug produced by fetching a `contract_call` transaction. It was generally caused due to the query that returned `smart_contracts` with null `abi`.

For details refer to issue #850 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


asimm241 (changes requested):

src/tests/api-tests.ts
```diff
@@ -5393,6 +5394,171 @@ describe('api tests', () => {
     expect(blockQuery.body.execution_cost_write_length).toBe(3);
   });
 
+  test.only('empty abi', async () => {
```
asimm241, line 5397: `only` test typo?

asimm241 (changes requested):

src/tests/api-tests.ts
```diff
@@ -5393,6 +5394,171 @@ describe('api tests', () => {
...
+        function_signature:
+          '(define-public (bns-name-preorder (hashedSaltedFqn (buff 20)) (stxToBurn uint) (paymentSIP010Trait trait_reference) (reciprocityTokenTrait trait_reference) (referencerWallet principal)))',
+      },
+    };
+
+    await db.updateBlock(client, dbBlock);
+    await db.updateTx(client, tx1);
+    await db.updateTx(client, tx2);
+    await db.updateSmartContract(client, tx1, contractCall);
+    await db.updateSmartContract(client, tx2, contractCall2);
```
rafaelcr, line 5553: Instead of making these updates directly, can you do this through the standard `update()` call? We're starting to use materialized views and other logic that needs to happen on update and I think it would be best if the code gets exercised as much as possible.

src/api/controllers/db-controller.ts
```diff
@@ -992,6 +992,10 @@ function parseContractCallMetadata(
   if (!contract.found) {
     throw new Error(`Failed to lookup smart contract by ID ${parsedTx.contract_call.contract_id}`);
   }
+  if (!contract.result.abi)
```
rafaelcr, line 995: Small style nit to avoid possible future confusion: can you keep the curly braces here?

rafaelcr (changes requested):

src/datastore/postgres-store.ts
```diff
@@ -4543,8 +4543,7 @@ export class PgDataStore
         `
       SELECT tx_id, canonical, contract_id, block_height, source_code, abi
       FROM smart_contracts
-      WHERE contract_id = ANY($1)
-      ORDER BY abi != 'null' DESC, canonical DESC, microblock_canonical DESC, block_height DESC
+      WHERE contract_id = ANY($1) AND abi != '' AND abi IS NOT NULL AND canonical = true AND microblock_canonical = true
```
zone117x, line 4546: This change has the side effect of no longer returning smart contract data that is orphaned or micro-orphaned. The previous query used `ORDER BY` to _prefer_ returning the canonical result, if available. However, the recent change to query for many smart contracts was missing an `ORDER BY contract_id` and a `DISTINCT ON(contract_id)`.

src/datastore/postgres-store.ts
```diff
@@ -4545,18 +4545,18 @@ export class PgDataStore
...
       }>(
         `
-      SELECT tx_id, canonical, contract_id, block_height, source_code, abi
-      FROM smart_contracts
-      WHERE contract_id = ANY($1)
-      ORDER BY abi != 'null' DESC, canonical DESC, microblock_canonical DESC, block_height DESC
+        SELECT DISTINCT ON (contract_id) contract_id, canonical, tx_id, block_height, source_code, abi
+        FROM smart_contracts
+        WHERE contract_id = ANY($1)
+        ORDER BY contract_id DESC, abi != 'null' DESC, canonical DESC, microblock_canonical DESC, block_height DESC
```
rafaelcr, line 4559: Can you add a database index with this sorting?

rafaelcr (changes requested):

src/datastore/postgres-store.ts
MHassanTariq: Done

src/api/controllers/db-controller.ts
```diff
@@ -938,6 +940,8 @@ function parseContractsWithDbTxs(contracts: DbSmartContract[], dbTxs: DbTx[]): T
       if (transaction) {
         transactions.push(transaction as Transaction);
       }
+    } else if (dbTx.type_id !== DbTxTypeId.ContractCall) {
```
zone117x, line 943: I'm confused on the condition here. What happens if there is no `contract` and the typeID not a contract call? There's no `else` to handle that, so the tx is just dropped?

src/api/controllers/db-controller.ts
```diff
@@ -948,17 +952,21 @@ function parseContractsWithMempoolTxs(
...
         { found: true, result: contract },
-        parseDbMempoolTx(dbMempool) as MempoolContractCallTransaction,
-        dbMempool
+        parseDbMempoolTx(dbMempoolTx) as MempoolContractCallTransaction,
+        dbMempoolTx
       );
       if (transaction) {
         transactions.push(transaction as MempoolTransaction);
       }
+    } else if (dbMempoolTx.type_id !== DbTxTypeId.ContractCall) {
```
zone117x, line 968: Same here, what happens in the condition where neither of the above are true?

src/api/controllers/db-controller.ts
```diff
@@ -1000,6 +1008,11 @@ function parseContractCallMetadata(
   if (!contract.found) {
     throw new Error(`Failed to lookup smart contract by ID ${parsedTx.contract_call.contract_id}`);
   }
+  if (!contract.result.abi) {
+    throw new Error(
```
zone117x, line 1012: What happens if the contract is found but missing an ABI? Won't this cause the same error that the PR is intending to fix?

src/migrations/1588266891631_smart_contracts.ts
```diff
@@ -62,4 +62,10 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
   pgm.createIndex('smart_contracts', 'canonical');
   pgm.createIndex('smart_contracts', 'contract_id');
 
+  pgm.createIndex('smart_contracts', [
+    { name: 'contract_id', sort: 'DESC' },
+    { name: 'canonical', sort: 'DESC' },
+    { name: 'microblock_canonical', sort: 'DESC' },
+    { name: 'block_height', sort: 'DESC' },
```
rafaelcr, line 69: You're missing the `abi` column

src/api/controllers/db-controller.ts
MHassanTariq: Another misunderstanding. I am considering contracts without `ABI` to be faulty. Thus, throwing error on this point. 
I will return `parsedTx` at this point to return other information (i.e. without `ABI`)

src/api/controllers/db-controller.ts
MHassanTariq: I was confused about whether it would be mandatory for every transaction of type `ContractCall` to have the associated contract in the DB or not. So, if such a situation occurs, I drop the transaction.
Do you suggest parsing that transaction like all others and passing it on, without the contract information?

zone117x (approved):
Need to get tests passing then LGTM!

rafaelcr (approved):
👍

PR #860, opened 2021-11-18
Feat: Return all the contracts implement a given trait

### LINKED ISSUE #649: Endpoint for contracts that implement a given trait
I was speaking with @zone117x and I think it'd be great if we had an endpoint that would expose all contracts that implement a specific trait. I personally think there are many instances where someone might want to get a list of contracts that all implement a trait, for example all contracts that implement the stacking pool trait, or all sip 10 contracts.

> aulneau
> is there a way to get a list of all contracts that implement a certain trait?
>
>mattl
> hmm. no endpoints at the moment, but I think it would be possible to implement, at least for contracts that explicitly implement a trait, like with this syntax (impl-trait 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6.nft-trait.nft-trait)
but I think contracts can implicitly implement a trait, as in they don't have to specify that impl-trait but they will still be treated as a contract with that trait if the correct functions are implemented
in that case, I think maybe the client could POST the trait JSON ABI, and I could write some interesting postgres json queries to find those contracts

asimm241:
## Description
This PR implements a contract endpoint `extended/v1/contract/trait` which returns all those contracts implement a given trait. The trait will be provided in the post request in contract abi format. 

closes #649 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Testing information
- Added unit test cases

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated


src/api/routes/contract.ts
```diff
@@ -35,5 +37,24 @@ export function createContractRouter(db: DataStore): RouterWithAsync {
     res.json({ limit, offset, results: parsedEvents });
   });
 
+  router.postAsync('/trait', async (req, res) => {
+    const trait = req.body.trait_abi;
+    if (!trait || !('functions' in trait)) {
+      res.status(400).json({ error: 'invalid trait abi' });
```
zone117x, line 43: Should `return` after this line

src/datastore/postgres-store.ts
```diff
@@ -4649,6 +4649,60 @@ export class PgDataStore
...
+        canonical: boolean;
+        contract_id: string;
+        block_height: number;
+        source_code: string;
+        abi: string;
+      }>(
+        `
+        SELECT tx_id, canonical, contract_id, block_height, source_code, abi
+        FROM smart_contracts
+        WHERE abi->'functions' @> $1::jsonb AND canonical = true AND microblock_canonical = true
```
zone117x, line 4682: Looks like a GIN index supports jsonb containment operator, e.g.
```sql
CREATE INDEX my_index ON smart_contracts USING GIN (abi)
```

src/migrations/1588266891631_smart_contracts.ts
```diff
@@ -61,4 +61,9 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
   pgm.createIndex('smart_contracts', 'microblock_canonical');
   pgm.createIndex('smart_contracts', 'canonical');
   pgm.createIndex('smart_contracts', 'contract_id');
+  pgm.sql('CREATE INDEX index_contract_abi ON smart_contracts USING GIN (abi)');
```
zone117x, line 64: Does `pgm.createIndex` have options to configure the index type to be `gin`?

asimm241: No, I spent some time reading the documentation but had no luck.

rafaelcr (approved):
I agree with Matt's observations and I think they should be addressed, otherwise LGTM 👍

src/migrations/1588266891631_smart_contracts.ts
zone117x: Looking at the ts defs, it seems like this should work:
```ts
pgm.createIndex('smart_contracts', 'abi', { method: 'gin' });
```

docs/openapi.yaml
```diff
@@ -794,12 +803,43 @@ paths:
           required: false
           schema:
             type: integer
+      x-codegen-request-body-name: key
+      requestBody:
```
zone117x, line 807: A POST body generally seems like the right shape for this endpoint, however, it won't work with the up-coming HTTP caching changes, because the browser/CDN caching is keyed off the URL. 

I wonder if it would be reasonable to specify the ABI as a query param. The sample NFT trait comes out to around 1KB in length after doing `encodeURIComponent(JSON.stringify(traitAbiJson)).length)`. The default max http header size of of nginx, apache, nodejs all appear to be around 8KB to 16KB. Perhaps that's a reasonable limit anyway? We need to impose some size limit on the JSON body anyway so we aren't sending megabytes of sql query to postgres.

asimm241: done!

zone117x (approved):
LGTM!

docs/openapi.yaml
```diff
@@ -774,17 +774,32 @@ paths:
           type: boolean
           default: false
 
-  /extended/v1/contract/{contract_id}/events:
+  /extended/v1/contract/trait/contracts:
```
rafaelcr, line 777: To be consistent with other paths on the API (get block by height, etc.), I think this path should be: `/extended/v1/contract/by_trait`

rafaelcr (approved):
Small path change suggestion, otherwise LGTM

docs/openapi.yaml
asimm241: This endpoint will be conflicting with `/extended/v1/contract/by_trait/:contract_id`

rafaelcr: As long as you declare it above the variable route you'll be OK: https://stackoverflow.com/questions/31493190/express-parameterized-route-conflict

asimm241: Done!

aulneau:
Omg this is so exciting

PR #861, opened 2021-11-19
feat: add heartbeat to websockets

### LINKED ISSUE #820: A heartbeat for web sockets
Please give web sockets a heart beat.  Ideally every 5 seconds, but could be as long as 15 seconds.  

This permits keep alive, health check, and can spawn recovery on the client side.  

@agraebe I could blog on a breakdown on how current clients handle web sockets pub/sub today in the browser to help create a better use case to drive requirements if that helps.

rafaelcr:
## Description

This patch adds a heartbeat to the Websockets server by sending a `ping` control message to every subscribed client on each of our supported topics every 5 seconds.

There was no need to change anything in our client library. From the [`ws` library docs](https://www.npmjs.com/package/ws):
> Pong messages are automatically sent in response to ping messages as required by the spec.

Closes #820 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No

No

## Testing information
Local tests.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/routes/ws/ws-rpc.ts
+++ b/src/api/routes/ws/ws-rpc.ts
@@ -53,11 +53,10 @@ class SubscriptionManager {
   heartbeatInterval?: NodeJS.Timeout;
   readonly heartbeatIntervalMs = 5_000;
 
-  constructor() {
-    this.setUpHeartbeat();
-  }
-
   addSubscription(client: WebSocket, topicId: string) {
+    if (this.subscriptions.size === 0) {
+      this.startHeartbeat();
+    }
     let clients = this.subscriptions.get(topicId);
     if (!clients) {
       clients = new Set();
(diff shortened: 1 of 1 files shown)


src/api/routes/ws/ws-rpc.ts
```diff
@@ -68,6 +80,24 @@ class SubscriptionManager {
         this.subscriptions.delete(topicId);
       }
     }
+    this.liveSockets.delete(client);
+  }
+
+  setUpHeartbeat() {
+    setInterval(() => {
```
zone117x, line 87: `setInterval` returns a timer object https://nodejs.org/api/timers.html#class-timeout -- does this need to be cleared during server shutdown?

zone117x (approved):
LGTM!

Issue #851, opened 2021-11-16
Return 400 on invalid requests to `/extended/v1/tx/{id}`


zone117x:
The API currently throws on requests like `/extended/v1/tx/0x[object%20Object]`, with `Error: Hex string is an odd number of digits: 0x[object Object]`, and spams our log output with `error` level messages. Should return 400 and not log an error.

Issue #854, opened 2021-11-16
Return 400 on invalid req pagination queries


zone117x:
API currently throws, returns a 500, and logs a critical error when invalid pagination query params are provided. For example:
`/extended/v1/address/SPJW1XE278YMCEYMXB8ZFGJMH8ZVAAEDP2S2PJYG.stacks-art-market-v2/transactions?limit=50&offset=-50`

```
Error: `limit` and `offset` must be integers
    at parsePagingQueryInput (/app/src/api/pagination.ts:9:11)
    at /app/src/api/routes/address.ts:170:41
    at wrappedMiddleware (/app/node_modules/@awaitjs/express/index.js:62:16)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:137:13)
```

This should return a 400, and not spam `error` level log messages.

PR #864, opened 2021-11-26
Return 400 for invalid requests on `/extended/v1/tx/{id}`

MHassanTariq:
## Description

The endpoint `/extended/v1/tx/{id}` logs error in console. This PR returns the error to client with status code 400. 
For details refer to issue #851 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

In case of any error due to the parsing of a transaction, the server will respond with the error to the client with status code 400.

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


src/api/routes/tx.ts
```diff
@@ -238,19 +238,28 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
...
+      /*
     const schemaPath = require.resolve(
       '@stacks/stacks-blockchain-api-types/entities/transactions/transaction.schema.json'
     );
     await validate(schemaPath, txQuery.result);
     */
-    res.json(txQuery.result);
+      return res.json(txQuery.result);
+    } catch (error: any) {
+      res.status(400).json(error.message);
```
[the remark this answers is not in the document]
zone117x: The main problem I see here is blindly catching all errors and returning a 400. The `searchTx` function should return a `found: false` value for 4xx errors. If it throws then it's likely a bug which should be a 5xx internal server error. Or it's because of bad user input which should be explicitly checked beforehand in order to return a 400. 

For example, the query param parsing above this `try` statement is where the Error is actually thrown, which results in the problem described in the issue https://github.com/hirosystems/stacks-blockchain-api/issues/851.

Wrapping all these routes in a `try ... catch { return 400 }` is the wrong approach. Seems like there's two reasonable approaches: 

A) Refactor the functions that parse the query params so that they return something indicating bad input rather than throwing, and modify the route handlers to check for the bad input results and return a 400.

B) Create a custom Error instance that can indicate to the Express error handler that the error already triggered a 400 response, and avoid logging an `level: error` message.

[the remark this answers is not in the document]
MHassanTariq: @zone117x I have worked with your second suggested approach. Can you just go through the code once and let me know if this looks good? If you're fine with it, I will use this `RespError` class for throwing errors in my other PR, too. 

I have also added status code and `RespErrorType` into the class obj, to use different status codes for any such issue in the future that requires a different status code or treatment.

src/api/routes/tx.ts
```diff
@@ -251,19 +252,31 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
...
+      // TODO: this validation needs fixed now that the mempool-tx and mined-tx types no longer overlap
+      /*
     const schemaPath = require.resolve(
       '@stacks/stacks-blockchain-api-types/entities/transactions/transaction.schema.json'
     );
     await validate(schemaPath, txQuery.result);
     */
-    res.json(txQuery.result);
+      return res.json(txQuery.result);
+    } catch (error: any) {
```
zone117x, line 274: Instead of wrapping try/catches around all the route handlers, the registered Express error handler should check for this error type, and return a response status.

src/api/routes/tx.ts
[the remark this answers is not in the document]
zone117x: Left a comment, generally looks like you're headed in the right direction.

src/api/routes/tx.ts
MHassanTariq: Thanks for pointing this approach out. Helped a lot.

src/api/init.ts
```diff
@@ -224,6 +224,9 @@ export async function startApiServer(opts: {
 
   // Setup error handler (must be added at the end of the middleware stack)
   app.use(((error, req, res, next) => {
+    if (error instanceof RespError) {
```
zone117x, line 227: This should be moved further down, next to the existing `if (error && !res.headersSent)` condition.

If headers have already been sent, then this will fail.

src/api/init.ts
```diff
@@ -242,6 +242,9 @@ export async function startApiServer(opts: {
       setResponseNonCacheable(res);
     }
     if (error && !res.headersSent) {
+      if (error instanceof RespError) {
+        return res.status(error.status).send({ error: error.message });
+      }
       res.status(500);
       const errorTag = uuid();
       Object.assign(error, { errorTag: errorTag });
```
zone117x, line 250: The `expressWinston.errorLogger` also needs configured so that it doesn't emit `level: error` logs for these. Something like...
```ts
  app.use(
    expressWinston.errorLogger({
      winstonInstance: logger as winston.Logger,
      metaField: (null as unknown) as string,
      blacklistedMetaFields: ['trace', 'os', 'process'],
      skip: (_req, _res, error) => {
        // Do not log errors for client 4xx responses
        return error instanceof RespError;
      },
    })
  );
```

MHassanTariq: Well, so far it's not emitting any error like that because I return after sending the response. Logs are after the error handler middleware. But I'll write the skip script instead. That makes more sense.

src/helpers.ts
```diff
@@ -36,6 +36,20 @@ export const EMPTY_HASH_256 = '0x00000000000000000000000000000000000000000000000
 
 export const pipelineAsync = util.promisify(stream.pipeline);
 
+// This class and enum are to throw errors that are supposed to be sent to the client
+enum RespErrorType {
+  invalid_tx_id = 'Invalid tx id',
+}
+export class RespError extends Error {
```
zone117x, line 43: This would make more sense if it was named something like `InvalidRequestError`

MHassanTariq: I want to use the same class for all the errors that are to be sent to the client and not logged at all. So, would be using the same class for #854. That was the reason behind naming this class as `RespError`. But I'll convert it to `InvalidRequestError`, if that is the category of errors we are concerned about.

src/helpers.ts
```diff
@@ -516,10 +530,16 @@ export function hexToBuffer(hex: string): Buffer {
     return Buffer.alloc(0);
   }
   if (!hex.startsWith('0x')) {
-    throw new Error(`Hex string is missing the "0x" prefix: "${hex}"`);
+    throw new InvalidRequestError(
```
zone117x, line 533: There's a bit of a separation of concern issue here. We're adding in public API Express route handling code into this `hexToBuffer` function, which is used in a lot of areas unrelated to the express routes.

The alternatives are more LOC, but would be a lot more explicit and we'd probably better off. One alternative could be going through the express routes, and validating any request params. Something we should probably be doing anyway. An example would be something like:
```ts
router.getAsync('/:tx_id', async (req, res, next) => {
  const { tx_id } = req.params;
  if (!has0xPrefix(tx_id)) {
    return res.redirect('/extended/v1/tx/0x' + tx_id);
  }

  try {
    hexToBuffer(tx_id);
  } catch (error: any) {
    throw new InvalidRequestError(error.message, InvalidRequestErrorType.invalid_tx_id);
  }

  ...

});
```

It should be pretty straightforward to do this for all the routes that take a hex-string input.

src/api/routes/tx.ts
```diff
@@ -248,6 +251,12 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
       return res.redirect('/extended/v1/tx/0x' + tx_id);
     }
 
+    try {
+      hexToBuffer(tx_id);
+    } catch (error: any) {
+      throw new InvalidRequestError(error.message, InvalidRequestErrorType.invalid_tx_id);
+    }
+
```
zone117x, line 259: This looks like it should work well. Can you apply similar logic for the pagination parsing errors? E.g.
```ts
let limit: number;
let offset: number;
try {
  const limit = parseTxQueryLimit(req.query.limit ?? 96);
  const offset = parsePagingQueryInput(req.query.offset ?? 0);
} catch (error) {
  throw new InvalidRequestError(error.message, InvalidRequestErrorType.invalid_tx_id);
}
```

And can you scan through the rest of the endpoints fof hex input values to apply the same validation?

MHassanTariq: Sure! Thanks a lot for the thorough feedback.

zone117x: Also, in the case of the `parseTxQueryLimit` and `parsePagingQueryInput` functions, since those are explicitly used for route handling, they can directly throw the `InvalidRequestError` error

src/api/routes/tx.ts
```diff
@@ -248,6 +251,12 @@ export function createTxRouter(db: DataStore): RouterWithAsync {
       return res.redirect('/extended/v1/tx/0x' + tx_id);
     }
 
+    try {
+      hexToBuffer(tx_id);
+    } catch (error: any) {
+      throw new InvalidRequestError(error.message, InvalidRequestErrorType.invalid_tx_id);
+    }
```
zone117x, line 258: Also note if you find quite of a bit places that need this code, you should probably create a helper function in `src/api/query-helpers.ts` like `validateRequestHexInput`

zone117x (commented):
Also need to scan for additional areas that handle 400 status and throw an error, like the `handleBadRequest` function, looks like that can be modified to throw an `InvalidRequestError` error

zone117x (commented):
@MHassanTariq can you add tests for this behavior?

* Test that invalid hex input on all modified routes now results in a 400 response rather than a 500

* Add a hook into the `logger` instance to ensure that these types of invalid requests no longer have `level: error`

src/helpers.ts
```diff
@@ -36,6 +36,23 @@ export const EMPTY_HASH_256 = '0x00000000000000000000000000000000000000000000000
...
+}
+export class InvalidRequestError extends Error {
+  type: InvalidRequestErrorType;
+  status: number;
+  constructor(msg: string, type: InvalidRequestErrorType, status: number = 400) {
+    super(msg);
+    this.type = type;
+    this.status = status;
+  }
+}
```
rafaelcr, line 54: We should move these (and probably the new validation functions) to the `errors.ts` file. We will need to add more useful types for debugging in the future, IMO, so we should keep them organized.

rafaelcr (changes requested):

rafaelcr (approved):
LGTM

src/api/query-helpers.ts
```diff
@@ -201,3 +202,20 @@ export function parseTraitAbi(req: Request, res: Response, next: NextFunction):
   }
   handleBadRequest(res, next, `Invalid 'trait_abi'`);
 }
+
+export function validateRequestHexInput(hash: string) {
+  try {
+    hexToBuffer(hash);
```
zone117x, line 208: The `hexToBuffer` function only does basic checking for input validation (the 0x prefix, even number of chars), but it doesn't account for situations like `Buffer.from('boop', 'hex').toString('hex') === ''` (invalid hex chars are generally ignored without throwing). 

I think we should perform that check within this `validateRequestHexInput` function since it will add a bit of overhead, and lots of other code calls `hexToBuffer` with known good data.

There's a few ways to validate if the `hash` string is valid hex chars. I'd run a quick benchmark (or find a resource if someone else has) to pick a faster one. For example: regex, `Buffer.from(hash, 'hex').toString('hex') === hash).toLowerCase()`, etc.

zone117x (changes requested):

src/api/query-helpers.ts
```diff
@@ -201,3 +202,23 @@ export function parseTraitAbi(req: Request, res: Response, next: NextFunction):
   }
   handleBadRequest(res, next, `Invalid 'trait_abi'`);
 }
+
+export function validateRequestHexInput(hash: string) {
+  try {
+    const buffer = hexToBuffer(hash);
+    if (buffer.toString() === '') {
+      throw new Error('Invalid hash characters');
```
zone117x, line 210: How does this validate the hex string? I left a couple examples of how to validate the string in my previous comment.

src/tests/api-tests.ts
```diff
@@ -9208,6 +9208,51 @@ describe('api tests', () => {
     expect(blockQuery.body.execution_cost_write_length).toBe(3);
   });
 
+  test('400 response errors', async () => {
```
zone117x, line 9211: Can you add tests to ensure http status code === 400 for these requests?

zone117x (approved):
LGTM, thanks @MHassanTariq!

PR #866, opened 2021-11-29
Return 400 on invalid req pagination queries

MHassanTariq:
## Description

This PR is an extension of [PR 846](https://github.com/hirosystems/stacks-blockchain-api/pull/864). This PR returns error msg to client rather than printing them on console.

For details refer to issue #854 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/routes/address.ts
+++ b/src/api/routes/address.ts
@@ -1,7 +1,7 @@
 import * as express from 'express';
 import { addAsync, RouterWithAsync } from '@awaitjs/express';
 import * as Bluebird from 'bluebird';
-import { DataStore } from '../../datastore/common';
+import { DataStore, DbTx } from '../../datastore/common';
 import { parseLimitQuery, parsePagingQueryInput } from '../pagination';
 import { isUnanchoredRequest, getBlockParams } from '../query-helpers';
 import {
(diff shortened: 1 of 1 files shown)


src/api/routes/address.ts
```diff
@@ -164,16 +164,25 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
...
+      offset = parsePagingQueryInput(req.query.offset ?? 0);
+      const addressTxsDb = await db.getAddressTxs({
+        stxAddress: stxAddress,
+        limit,
+        offset,
+        ...blockParams,
+      });
+      txResults = addressTxsDb.results;
+      total = addressTxsDb.total;
+    } catch (err: any) {
```
zone117x, line 183: Catching everything like this isn't good practice. Either throw and handle a specific error type, or refactor the functions to return something like an optional result rather than throw

kyranjamie: I sometimes use the safe await pattern to avoid of try/catch blocks and more deliberately catch errors https://github.com/DavidWells/safe-await

src/api/routes/address.ts
```diff
@@ -164,16 +164,25 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
...
+      res.status(400).json({ error: err.message });
```
rafaelcr, line 184: Also here, should be return a generic error string to avoid revealing any internal info?

zone117x (changes requested):
This approach of `try ... catch { return 400 }` has the same problems described in https://github.com/hirosystems/stacks-blockchain-api/pull/864#discussion_r763178034

zone117x:
Closing this PR for now in favor of https://github.com/hirosystems/stacks-blockchain-api/pull/864 which seems like it will be fixing the same issue. Feel free to re-open otherwise.

PR #872, opened 2021-12-03
Fixed issue with spnsored transaction

### LINKED ISSUE #857: Issues related to sponsored transactions
I've been playing with sponsored transactions, and I've noticed a few bugs related to how the API handles these transactions:

The biggest issue is probably related to fees, and how balances are effected. I'm not sure what the "root" payload is for sponsored transactions, but it appears that I can set two things:
- sender spending condition fee
- sponsor spending condition fee

What happens is the "sender fee" is picked up by the API as the "fee", even though this fee has no effect. The "sponsor fee", a different value, is not picked up by the API. Downstream, an effect is that the sender's balance is updated (using the void fee), and the sponsor's balance is not effected.

I first noticed this because I ended up with an account with a negative balance (according to the API). Core node APIs reflect balances as expected.

The secondary issue is related to nonces. The sponsor's nonce is not picked up, so the usual nonce APIs are incorrect for that principal. This isn't the end of the world, as the sponsor is running some backend service and can keep track of it. It would be really great to have _some_ index of this in the future, though.

Here is a [STX address with a negative balance](https://explorer.stacks.co/address/SPY3T5EAGZQYCG3294CYSQGZB6CADZX9V17C0XNB?chain=mainnet). The first and only transaction from that account is sponsored, but these bugs are unrelated to whether the account had any balance before. 

The [sponsor](https://explorer.stacks.co/address/SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE?chain=mainnet) of that transaction has an incorrect balance from the API. The 'sponsor fee' was 0.1 STX, and you can see there is that discrepancy between the explorer / API and the [core node /v2/accounts](https://stacks-node-api.stacks.co/v2/accounts/SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE) response.

To repro, just follow the example in the [stacks.js README](https://github.com/blockstack/stacks.js/tree/master/packages/transactions#sponsoring-transactions).

asimm241:
## Description
I have found **four** issues with sponsored transactions.

1. Incorrect guide in the [readme](https://github.com/blockstack/stacks.js/tree/master/packages/transactions#sponsoring-transactions) for sponsored transactions.  Here is the PR for that issue:  to be added
2. Invalid signature for sponsored transactions. Here is the PR for that issue: https://github.com/blockstack/stacks.js/pull/1103
3. Incorrect fee rate for the sponsored transactions. This PR fixes that.
4. Incorrect nonce calculation for the sponsored transactions. The PR fixes that. 

Closes #857 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

_NO_

## Testing information

- Updated unit tests for sponsored transactions. 
- Need to add more tests once this gets reviewed

## Checklist
- [x] `npm run test` passes


zone117x (commented):
Do the queries for generating address stx balances also need to be updated (like https://stacks-node-api.mainnet.stacks.co/extended/v1/address/SPY3T5EAGZQYCG3294CYSQGZB6CADZX9V17C0XNB/stx)?

asimm241:
> Do the queries for generating address stx balances also need to be updated (like https://stacks-node-api.mainnet.stacks.co/extended/v1/address/SPY3T5EAGZQYCG3294CYSQGZB6CADZX9V17C0XNB/stx)?

@zone117x I think so, but I don't see information about sponsored transactions in `stx_events`. STX events only have sender and receiver. What do you think.?

zone117x:
> @zone117x I think so, but I don't see information about sponsored transactions in `stx_events`. STX events only have sender and receiver. What do you think.?

The STX events table queries should be okay, it would be the tx fee queries that I assume creates the balance bug.

rafaelcr (approved):
LGTM

src/datastore/postgres-store.ts
```diff
@@ -4806,7 +4806,7 @@ export class PgDataStore
       `
       SELECT sum(fee_rate) as fee_sum
       FROM txs
-      WHERE canonical = true AND microblock_canonical = true AND sender_address = $1 AND block_height <= $2
+      WHERE canonical = true AND microblock_canonical = true AND (sender_address = $1 OR sponsor_address = $1) AND block_height <= $2
```
zone117x, line 4809: Is there a unit test that replicates the negative balance bug in https://github.com/hirosystems/stacks-blockchain-api/issues/857? Won't this query include fees for a sponsored tx for both the sender and the sponsor?

asimm241: > Won't this query include fees for a sponsored tx for both the sender and the sponsor?

hmm, I think you're right. 

> Is there a unit test that replicates the negative balance bug in #857?
Not yet, I'm working on it.

zone117x: Can you include an explicit unit test for the negative balance behavior created by the situation in the referenced issue?

asimm241: Done!

src/tests/api-tests.ts
```diff
@@ -5297,6 +5325,132 @@ describe('api tests', () => {
...
+      lock_tx_id: '',
+      locked: '0',
+      lock_height: 0,
+      burnchain_lock_height: 0,
+      burnchain_unlock_height: 0,
+    };
+    const fetchStxBalance = await supertest(api.server).get(`/extended/v1/address/${address}/stx`);
+    expect(fetchStxBalance.status).toBe(200);
+    expect(fetchStxBalance.type).toBe('application/json');
+    expect(JSON.parse(fetchStxBalance.text)).toEqual(expectedResp);
```
zone117x, line 5451: Can you also confirm expected STX balance of the sponsor address?

zone117x: Still don't see a any testing for the sponsor address `SP2ZRX0K27GW0SP3GJCEMHD95TQGJMKB7GB36ZAR0`

asimm241: oh, I thought you meant that create a scenario like that.

asimm241: > Still don't see a any testing for the sponsor address `SP2ZRX0K27GW0SP3GJCEMHD95TQGJMKB7GB36ZAR0`

Sorry, I don't understand, please elaborate, how can I test live db locally with updated queries?

zone117x: The test does not demonstrate that the STX balance of both the origin account and the sponsor account are correct. 

The above only does a balance check for the variable `address` (the origin sender), but does not test the balance for the sponsor address, which `SP2ZRX0K27GW0SP3GJCEMHD95TQGJMKB7GB36ZAR0` in the above code.

src/datastore/postgres-store.ts
```diff
@@ -1712,7 +1712,7 @@ export class PgDataStore
         `
         SELECT MAX(nonce) nonce
         FROM txs
-        WHERE sender_address = $1
+        WHERE ((sender_address = $1 AND sponsored = false) OR (sponsor_address = $1 AND sponsored= true))
```
zone117x, line 1715: These sql query changes can have considerable performance impact. Previously these were simple equality conditions against an index column `sender_address`. Now, it's checking multiple columns, and I don't think the `sponsor_address` or `sponsored` columns are indexed, and there would also have to be composite index now for both pairs of columns.

src/tests/api-tests.ts
zone117x: Balance tests look good now, thanks!

src/datastore/postgres-store.ts
zone117x: @asimm241 do you have a local pg db with these changes and the new indexes? I'd like to see the performance impact, and the `EXPLAIN` result to show that the indexes are used for this query.

zone117x: I ran this query on a local pg instance with full prod data. Performance difference seems alright, although hard to compare that to how it will perform on google cloud sql.

@rafaelcr can you also give this a review?

zone117x (approved):
Approving, but we need to hold off on merging this into `develop` because we're about to cut a release, and there's a few breaking changes in this PR that will prevent us from easily testing on our current test deployment.

PR #874, opened 2021-12-03
Move build-publish Github actions to docker/build-push-action@v2

### LINKED ISSUE #832: Upgrade to docker/build-push-action v2 and move images to hirosystems/
Since we're migrating this repository from blockstack/ to the hirosystems/ Github org, we should also move our Docker images to be under hirosystems/. However, we'd like to keep supporting blockstack/ images for a while so current users don't lose access to newest versions right away.

In order to achieve this, we should upgrade our docker/build-push-action Github action from v1 to v2 so we can upload the built images to multiple destinations.

cc @CharlieC3

rafaelcr:
## Description

This patch upgrades our `build-publish` Github actions step to use `docker/build-push-action@v2`. It also makes a few improvements:
- Push images to `blockstack/` and `hirosystems/` on Docker hub
- Limit `branch` event hook to only run actions on `master` and `develop` to avoid duplicating work with every PR push
- Adds `cmake` dependency to Dockerfiles so CPU profiling tools work correctly
- Checks out the repo with full depth so git tags work correctly

Closes #832 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

No

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/.github/workflows/stacks-blockchain-api.yml
+++ b/.github/workflows/stacks-blockchain-api.yml
@@ -440,6 +440,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           token: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
 
       - name: Semantic Release
         uses: cycjimmy/semantic-release-action@v2.5.0
@@ -458,7 +459,7 @@ jobs:
             semantic-release-slack-bot
             @semantic-release/exec
 
-      - name: Docker meta
+      - name: Docker Meta
         id: meta
         uses: docker/metadata-action@v3
         with:
@@ -466,10 +467,17 @@ jobs:
             blockstack/${{ github.workflow }}
             hirosystems/${{ github.workflow }}
           tags: |
-            type=ref,event=branch
-            type=ref,event=pr
             type=semver,pattern=${{ steps.semantic.outputs.new_release_version }}
-            type=sha
+
+      - name: Docker Standalone Meta
+        id: meta_standalone
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            blockstack/${{ github.workflow }}-standalone
+            hirosystems/${{ github.workflow }}-standalone
+          tags: |
+            type=semver,pattern=${{ steps.semantic.outputs.new_release_version }}
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/stacks-blockchain-api.yml
```diff
@@ -10,6 +10,8 @@ on:
       - '**/CHANGELOG.md'
       - '**/package.json'
   pull_request:
+    types:
+      - opened
```
CharlieC3, line 14: Instead of only having CI run when a PR is opened, ideally it would run on every new commit to the PR as well to ensure the merged code is being used in a build. If it's only running on branch commits and not PRs, then the images will potentially be missing new commits added downstream in the `develop` or `master` branch.

[See how this is configured here](https://github.com/hirosystems/gh-actions-example/blob/main/.github/workflows/ci.yml#L41-L43), where `pull_request` isn't filtered, but the _branches_ are filtered to only run on long-lived branches, like main/master and develop. If you're looking to avoid duplicate builds on a single commit, this would be a good setup for that.

.github/workflows/stacks-blockchain-api.yml
```diff
@@ -452,28 +459,55 @@ jobs:
...
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            blockstack/${{ github.workflow }}
+            hirosystems/${{ github.workflow }}
+          tags: |
+            type=ref,event=branch
+            type=semver,pattern=${{ steps.semantic.outputs.new_release_version }}
+            type=sha
```
CharlieC3, line 472: Not sure this will work for either image, I think the `pattern` is invalid, and it won't create a docker tag for PRs. [We typically configure it like this](https://github.com/hirosystems/gh-actions-example/blob/main/.github/workflows/ci.yml#L80-L83) to account for pr's, branches, semantic versions and a shorter version. You can tack on the `type=sha` too if you want to keep that.

.github/workflows/stacks-blockchain-api.yml
```diff
@@ -452,28 +459,55 @@ jobs:
...
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
-          tags: ${{ steps.semantic.outputs.new_release_version }}
-          tag_with_ref: true
-          add_git_labels: true
+
+      - name: Build/Tag/Push Image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
```
CharlieC3, line 498: This can be deleted for both images, I believe it's the default value.

.github/workflows/stacks-blockchain-api.yml
```diff
@@ -452,28 +459,55 @@ jobs:
...
+        with:
+          images: |
+            blockstack/${{ github.workflow }}-standalone
+            hirosystems/${{ github.workflow }}-standalone
+          tags: |
+            type=ref,event=branch
+            type=semver,pattern=${{ steps.semantic.outputs.new_release_version }}
+            type=sha
+
+      - name: Set up Docker Buildx
```
CharlieC3, line 486: Buildx is only needed with certain advanced docker usage features, like multi-platform builds or exporting a binary from a built container. You should be able to remove this step for both images.

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:14-alpine
 WORKDIR /app
 COPY . .
 
-RUN apk add --no-cache --virtual .build-deps alpine-sdk python2 git openjdk8-jre
+RUN apk add --no-cache --virtual .build-deps alpine-sdk python2 git openjdk8-jre cmake
```
CharlieC3, line 6: Strange, I though we changed this to `python3`?

CharlieC3 (changes requested):
Awesome job thus far! Suggesting a few minor changes that should be made for both images.

.github/workflows/stacks-blockchain-api.yml
rafaelcr: My original idea was to leave this 'PR opened' event plus the 'all branches' setting above which builds on every push for every branch (regardless of PR status). Do you think this could work as well?

I'd like to avoid building only on `develop`/`master` because sometimes it's very useful to push a branch to Github (without a PR) and do some very quick tests against the full test suite.

Dockerfile
rafaelcr: I think we patched directly to `master`. I'll make the change here.

.github/workflows/stacks-blockchain-api.yml
CharlieC3: I can see how that's useful, but I think the important part missing is that if a PR is opened, there always needs to be a build that reflects the most recent _merge_ commit. Building on a PR vs a feature branch is quite different in CI; they're technically two different commit SHAs. If CI is only configured to run when the PR is first opened, and not on subsequent merge commits, then it opens us up to potential bugs that would only be caught post-merge.

If you want it to build on every commit to a branch, that should be fine to keep that in there as well. The only downside is that commits will trigger two workflows and consume more GH runners (this is a shared resource throughout the org). Configuring it like this also means Github will consume resources on commits to branches that aren't ready for testing yet, and opening a draft PR I find is a good way to signal to Github that it's at a state where it can be tested and you'd find the results useful in some way.

.github/workflows/stacks-blockchain-api.yml
CharlieC3: > Thanks, I'll try that. If I make this change I still need to keep the "Semantic Release" step because it affects other things, right?

Yes, the Semantic Release step is the step actually performing the release. This meta step is only forming a string in the correct semver syntax than can be used when pushing up a docker tag.

.github/workflows/stacks-blockchain-api.yml
rafaelcr: Gotcha, thanks for the clarification 👍

CharlieC3 (approved):
LGTM! Thanks for making the changes! :shipit:

Issue #878, opened 2021-12-07
Refactor all usages of `router.getAsync(...)` to `asyncHandler`


zone117x:
https://github.com/hirosystems/stacks-blockchain-api/blob/8d6738817ee88f924ca964319fa125f0b725a4a6/src/api/async-handler.ts

Creating this issue from comment at https://github.com/hirosystems/stacks-blockchain-api/pull/834#discussion_r762938701

We've been using `router.getAsync` for all of the API routes. This function is an Express extension provided by the [`@awaitjs/express` lib](https://www.npmjs.com/package/@awaitjs/express). However, the `router.getAsync(<route>, handler1, handler2, handler3...)` function has incorrect/bugged behavior when given multiple router handler functions. It executes both synchronously at the same time, breaking the ability for the handlers to control the route flow based on the order in which they are specified.

Additionally, the `@awaitjs/express` lib is incorrectly typed. The `router.*Async(...)` functions incorrectly expect a `void` (synchronous) handler rather than a promise, which has forced this repo to disable the `@typescript-eslint/no-misused-promises` eslint rule, which could have prevent numerous bugs and PR review change-requests.

A better alternative has been implemented in [`src/api/async-handler.ts`](https://github.com/hirosystems/stacks-blockchain-api/blob/8d6738817ee88f924ca964319fa125f0b725a4a6/src/api/async-handler.ts) which can be used like:
* Before: `router.getAsync('/', async (req, res) => { ... })`
* After: `router.get('/', asyncHandler(async (req, res) => { ... })`

A cursory look shows we have ~60 usages. The refactor is pretty simple, but it tends to result in many LOC changes due to indentation changes, so it ends up being a bit disruptive to other PRs currently in-flight.

followmisha:
**@diwakergupta @lgalabru**

PR #880, opened 2021-12-08
refactor: remove `@await/express` usage

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/878

Removed the buggy `@awaitjs/express` lib and replaced all instances of `router.getAsync('/', async (req, res) => { ... })` with `router.get('/', asyncHandler(async (req, res) => { ... })`.

This PR _should_ have only been around 60 LOC changes, but the linting enforced whitespace changes make the changeset look huge. No routes should have any actual code changes.

With this refactor out of the way, the follow-up PR(s) adding cache-control support will be much more readable, since those will actually change the route handling code.

### DIFF
--- a/src/event-stream/event-server.ts
+++ b/src/event-stream/event-server.ts
@@ -3,7 +3,7 @@ import * as net from 'net';
 import { Server, createServer } from 'http';
 import * as express from 'express';
 import * as bodyParser from 'body-parser';
-import { addAsync } from '@awaitjs/express';
+import { asyncHandler } from '../api/async-handler';
 import PQueue from 'p-queue';
 import * as expressWinston from 'express-winston';
 import * as winston from 'winston';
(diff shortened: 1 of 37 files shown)


src/event-stream/event-server.ts
```diff
@@ -785,90 +785,112 @@ export async function startEventServer(opts: {
...
+        // Skip logging massive event payloads, this _should_ only exclude the genesis block payload which is ~80 MB.
+        if (payload.length > 10_000_000) {
+          payload = 'payload body too large for logging';
+        }
+        logger.debug(`[stacks-node event] ${eventPath} ${payload}`);
       }
-      logger.debug(`[stacks-node event] ${eventPath} ${payload}`);
-    }
-  });
+      next();
```
zone117x, line 807: This route did have a minor code change to add this `next()` call. The `@awaitjs/express` lib had a bug that caused the routes handlers below to be incorrectly called, synchronously, at the same time as this one.

rafaelcr (approved):
Awesome

Issue #762, opened 2021-09-15
OpenAPI spec: Group and/or mark proxied endpoints


agraebe:
It is unclear which endpoints are proxied to the Node API. To make the project relationship clear, we should categorize or mark proxied endpoints appropriately.

smcclellan:
Thank you for filing! I'll close this due to age & reasoning in the linked PR.

Issue #745, opened 2021-09-07
OpenAPI spec: Add linting rules


agraebe:
It seems like it is possible to [lint YAML files](https://stoplight.io/open-source/spectral/). We could use this to ensure that the OpenAPI spec includes all mandatory fields and follows best practices guidelines. 

These guidelines would become more clear once we did a manul review of the current file, likely in the next quarter.

cc @pgray-hiro

zone117x:
@agraebe this was closed by https://github.com/hirosystems/stacks-blockchain-api/pull/882 right?

agraebe:
correct!

followmisha:
**BE COOL**

Issue #744, opened 2021-09-07
OpenAPI spec: Naming conventions


agraebe:
As part of the OpenAPI spec review, we realized that there is a lack of naming conventions. It would be useful to extend out contributions guidelines with some initial conventions based on the [Google's Cloud API conventions](https://cloud.google.com/apis/design/naming_convention). Some points we should cover:
- case styles
- abbreviations
- singular vs plural 
- underscore vs hyphen

@rafaelcr @zone117x @pgray-hiro - is there anything that I missed?

@kyranjamie would you be able to contribute with some thoughts on this?

PR #882, opened 2021-12-08
feat: openapi lint config and grouping

agraebe:
contributing to the following issues:
- https://github.com/hirosystems/stacks-blockchain-api/issues/744
- https://github.com/hirosystems/stacks-blockchain-api/issues/762
- https://github.com/hirosystems/stacks-blockchain-api/issues/745
- https://github.com/hirosystems/docs/issues/27

cc @saralab

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -7,14 +7,67 @@ servers:
   - url: http://localhost:3999/
     description: Local
 info:
-  title: Stacks 2.0 Blockchain API
+  title: Stacks Blockchain API
   version: 'STACKS_API_VERSION'
   description: |
-    This is the documentation for the Stacks 2.0 Blockchain API.
+    Welcome to the API reference overview for the <a href="https://docs.hiro.so/get-started/stacks-blockchain-api">Stacks Blockchain API</a>.
 
-    It is comprised of two parts; the Stacks Blockchain API and the Stacks Core API.
+    <a href="https://hirosystems.github.io/stacks-blockchain-api/collection.json" download="stacks-api-collection.json">Download Postman collection</a>
 
-    <a href="collection.json" download="collection.json">Download Postman collection</a>
+tags:
+  - name: Accounts
+    description: Read-only endpoints to obtain Stacks account details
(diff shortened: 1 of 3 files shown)


docs/openapi.yaml
```diff
@@ -7,14 +7,67 @@ servers:
...
+    description: Read-only endpoints to obtain Stacking reward details
+    externalDocs:
+      description: Stacks Documentation - Stacking
+      url: https://docs.stacks.co/understand-stacks/stacking
+  - name: Transactions
+    description: Endpoints to obtain transaction details and to broadcast transactions to the network
+    externalDocs:
+      description: Hiro Documentation - Transactions
+      url: https://docs.hiro.so/get-started/transactions
+  - name: Tokens
```
rafaelcr, line 66: I think we should separate FTs from NFTs (or have additional specific tags instead of only 'Tokens'), given that we plan on adding a few new NFT endpoints and users are very specifically looking for those. Thoughts?

agraebe: I like this! will make the change

zone117x (commented):
Great looking changes! The vercel bot is gone again so we can't preview how the changes look when ran through redoc and the typescript-client generator. I'm guessing it broke after the repo org move..

agraebe:
> I'm guessing it broke after the repo org move..

how can we fix this? is this a low hanging fruit that could be part of this PR?

zone117x:
> > I'm guessing it broke after the repo org move..
> 
> how can we fix this? is this a low hanging fruit that could be part of this PR?

Charlie fixed vercel yesterday, you can see the comment now above. The doc builds aren't working though, likely an issue with the schema changes causing the npm build commands inside `/docs` and/or `/client` to fail

agraebe:
ah, I found the issue: some endpoints were assigned to two groups but it looks like that is not possible. To overcome that limitation, I have to remove the "Proxy" group again. Alternatively, we could remove the other group (eg transactions, accounts). However, that would make it difficult to discover important endpoints.

zone117x (approved):

Issue #885, opened 2021-12-09
transactions with transfers endpoint returning badly formatted value


aulneau:
Based on this spec: https://hirosystems.github.io/stacks-blockchain-api/#operation/get_account_transactions_with_transfers the data response should look like:
```json
"asset_identifier": "SP497E7RX3233ATBS2AB9G4WTHB63X5PBSP5VGAQ.boom-nfts::boom",
"value": {
  "hex": "0x00",
  "repr": "0"
},
```

However, the data is being returned as such:
```json
      "nft_transfers": [
        {
          "asset_identifier": "SPJW1XE278YMCEYMXB8ZFGJMH8ZVAAEDP2S2PJYG.stacks-giantpandas::stacks-giantpandas",
          "value": "\u0001\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0001]",
          "sender": "SP193B6A8GMXNFEH3WWCQC22QYYNSPHJTSGM80ZWY",
          "recipient": "SPNWZ5V2TPWGQGVDR6T7B6RQ4XMGZ4PXTEE0VQ0S.marketplace-v4"
        }
      ]
```

I believe the issue is this line here https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/api/routes/address.ts#L269

might be a bad copypasta

PR #886, opened 2021-12-09
fix: do not convert nft value object to string

aulneau:
This PR will hopefully resolve https://github.com/hirosystems/stacks-blockchain-api/issues/885, I followed where the events are converted from the db into what eventually ends up in the response:

https://github.com/hirosystems/stacks-blockchain-api/blob/04f3c68a9a00a0b16c2925f4150acd35b261ba2c/src/api/controllers/db-controller.ts#L273 

and it seems like it should already be formatted correctly when it [gets to this point](https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/api/routes/address.ts#L269). I am pretty unfamiliar with the API code so i apologize if this is completely incorrect. cc @zone117x


zone117x (approved):

rafaelcr (approved):
👍

PR #888, opened 2021-12-09
chore: fix jest not quickly finishing tests

zone117x:
jest would print `Jest did not exit one second after the test run has completed. This usually means that there are asynchronous operations that weren't stopped in your tests.`

This was caused by a long `setTimeout` in the CPU profiler endpoint that wasn't cleared when `GET /profile/cancel` was called.

Added support for passing an `AbortController` to the `timeout(...)` function. We're on Nodejs v14 which requires a 3rd party lib, we can get rid of it once we upgrade to Nodejs v16.


rafaelcr (approved):
LGTM

PR #892, opened 2021-12-09
chore: cache docker layers in to speed up stacks-node build CI step

zone117x:
Experimenting with github-cache support for intermediate docker layers in order to speed up the `stacks-node` rust build within the docker image build step


rafaelcr (approved):

CharlieC3:
The new `gha` cache is pretty awesome. We can likely use it in stacks-blockchain too.

PR #894, opened 2021-12-10
Fix contract-call tx bug in `address/<principal>/transactions_with_transfers`

zone117x:
Fix another bug introduced in #807.

Contract-call tx args were missing from the endpoint
`/extended/v1/address/<principal>/transactions_with_transfers`


rafaelcr (approved):
Nice

PR #901, opened 2021-12-13
chore: gh action caching for docker build layers

zone117x:
Enable github action caching for build layers. Minor changes to `Dockerfile` to cache the `apk add ...` and `npm install ...` steps.

Should bring image build time down from minutes to tens of seconds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,13 @@
 FROM node:14-alpine
 
 WORKDIR /app
-COPY . .
 
 RUN apk add --no-cache --virtual .build-deps alpine-sdk python3 git openjdk8-jre cmake
+COPY package.json package-lock.json ./
+RUN npm config set unsafe-perm true && npm install
+COPY . .
 RUN echo "GIT_TAG=$(git tag --points-at HEAD)" >> .env
-RUN npm config set unsafe-perm true && npm install && npm run build && npm prune --production
+RUN npm run build && npm prune --production
 RUN apk del .build-deps
 
 CMD ["node", "./lib/index.js"]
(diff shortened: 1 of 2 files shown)


rafaelcr (approved):
LGTM

zone117x:
Hmm this isn't working as intended. None of the following steps are being cached:
```
RUN apk add --no-cache --virtual .build-deps alpine-sdk python3 git openjdk8-jre cmake
COPY package.json package-lock.json ./
RUN npm config set unsafe-perm true && npm install
```
Not really sure why not, probably something weird with the docker context being detected as changed. I'll wait for @CharlieC3 to possibly provide some insight.

Dockerfile
```diff
@@ -1,11 +1,13 @@
...
 WORKDIR /app
-COPY . .
 
 RUN apk add --no-cache --virtual .build-deps alpine-sdk python3 git openjdk8-jre cmake
+COPY package.json package-lock.json ./
+RUN npm config set unsafe-perm true && npm install
+COPY . .
 RUN echo "GIT_TAG=$(git tag --points-at HEAD)" >> .env
-RUN npm config set unsafe-perm true && npm install && npm run build && npm prune --production
+RUN npm run build && npm prune --production
```
CharlieC3, line 10: I don't understand the effective changes implemented here. It looks like some Dockerfile directives were moved around and/or split into multiple directives. Can you clarify why this is needed?

CharlieC3 (commented):
Caching will often not work in this case due to the `COPY` directive, because it's very likely that shasum of what's being copied (e.g. the whole directory) is different than the previous commit. The rosetta dockerfile does not `COPY` much of anything. It just `git clone`'s the api's `master` branch every time, so it's a good candidate for this feature.

The `.git` directory is typically the most sensitive to change between commits for obvious reasons, and that directory is necessary when building this particular docker image. So we may not see this caching feature ever used.

zone117x:
Closing this PR for now. I opened an issue to track if we want to work on it in the future: https://github.com/hirosystems/stacks-blockchain-api/issues/1623

PR #905, opened 2021-12-14
fix: regression with `contract` http response `abi` field

zone117x:
Fix a regression introduced in #860 where the `smart_contracts` table `abi` column was changed from `text` to `jsonb`. The sql query responses also started returning a JSON object rather than a string, which ended up propagating to the http response for `/extended/v1/contract/<contract-id>`. 

This broke various products like the Explorer which attempted `JSON.parse(response.abi)`.


rafaelcr (approved):
Looks good!

PR #930, opened 2021-12-20
ci: fix docker tagging on release

CharlieC3:
Currently on releases, docker tagging is broken due to the recent refactoring in the GH actions workflow. This should help fix it by passing the release version directly to the metadata actions.


rafaelcr (approved):
Thanks @CharlieC3

Issue #924, opened 2021-12-19
Pagination seems broken on `extended/v1/tx/block_height/42041`


Jamil:
For blocks with a large number of transactions, the pagination returns the same transaction ID multiple times, when setting offsets. Try iterating through this block with offset in multiples of `96` or `200`; you never get all the transactions in the block, and some are duplicated.

Jamil:
Related: number of transactions per block is different in `https://stacks-node-api.mainnet.stacks.co/extended/v1/tx/block_height/42041` (shows a list of 987 transactions, or, after deduplication, 872 transactions). This is different from the `total: 870` returned by the `/tx/block_height` endpoint.

zone117x:
Thank you for the detailed report. Verified the issue, working on a fix.

zone117x:
@Jamil the two bugs you identified have been fixed in the `develop` branch. Note that the Hiro API deployment won't include these fixes until after the holidays.

PR #931, opened 2021-12-21
fix: incorrect tx list returned for block result 

zone117x:
Fixes the incorrect list of txs returned for a given block as described in #924

> Related: number of transactions per block is different in `https://stacks-node-api.mainnet.stacks.co/extended/v1/block/by_height/42041` (shows a list of 987 transactions, or, after deduplication, 872 transactions). This is different from the `total: 870` returned by the `/tx/block_height` endpoint.

The `getBlockWithMetadata` function, used for many endpoints to return a list of txids for a given block, was not filtering out non-canonical txs.


rafaelcr (approved):
LGTM

PR #933, opened 2021-12-21
fix: tx pagination failing due to lack of ordering #924

### LINKED ISSUE #924: Pagination seems broken on `extended/v1/tx/block_height/42041`

zone117x:
Closes #924

Paginating txs by block wasn't working correctly due to lack of ordering, so each page would have random-looking txs that could have been included in a previous page, as well as potentially never getting a full list of txs from ones randomly left out.

The first commit in this PR cleans up some long-standing TODOs around sql transaction integrity for fetching txs lists.


rafaelcr:
Looks very good so far 👍

rafaelcr (approved):
Nice

Issue #846, opened 2021-11-12
Implement new NFT endpoints


zone117x:
We've aligned on the need for more NFT related endpoints that revolve around ownership. These new endpoints have come out of research and discussion started from NFT metadata support. It's not currently practical for the API to implement metadata support for NFTs akin to a large scale OpenSea-like backend to power NFT marketplaces. However, something more reasonable and also useful we can implement are endpoints that can enumerate NFTs by their asset class, or get a list of NFTs owned by a given address, or show the history of ownership for a given NFT, etc. 

These endpoints will only include the type of data that is available in the db, e.g. contract addresses, NFT (Clarity) values, owner addresses, etc. They will _not_ include metadata that must be resolved via contract calls or resolving off-chain resources such as off-chain names/descriptions, images, arbitrary attributes, etc.

For more context on out-of-scope features see [Individual NFT (SIP-009) token metadata support](https://github.com/hirosystems/stacks-blockchain-api/issues/743), and [Complete NFT class metadata support](https://github.com/hirosystems/stacks-blockchain-api/issues/742).

Endpoints to implement:
* [x] Get all NFTs minted by a given asset class (smart contract principal + class name)
* [x] Get all NFTs owned by a given principal, can be filtered by asset class
* [x] Get ownership history of a given fully-qualified NFT asset class + NFT (Clarity) value

As mentioned above, the responses for these endpoints can include things like the asset class / address, txid, block height, NFT Clarity value.

The API _can_ include/parse information specific to a ratified SIP, however, only if it's possible with the data currently provided via the [Stacks Node event observer interface](https://github.com/blockstack/stacks-blockchain/blob/master/docs/event-dispatcher.md). E.g. SIP-specific contract call function signatures that the API can parse out of a tx, or contract log events. 

For example, [SIP-009](https://github.com/stacksgov/sips/blob/main/sips/sip-009/sip-009-nft-standard.md) specifies a contract trait function:
```
    ;; URI for metadata associated with the token 
    (get-token-uri (uint) (response (optional (string-ascii 256)) uint))
```
However, `token-uri` for a given NFT is not something we'd implement as part of the endpoints in this PR because it would require _resolving_ the data via contract-calls. If a contract log event for this data was also specified as part of a SIP, then it would be candidate for inclusion within these API endpoints.

zone117x:
@rafaelcr @agraebe @markmhx @kyranjamie let me know if I've missed any endpoints, I'll add them to the list.

[...]

PR #936, opened 2021-12-22
feat: add `/extended/v1/tokens/nft/holdings` endpoint

rafaelcr:
This patch creates the `/extended/v1/tokens/nft/holdings` endpoint which returns a paginated list of all NFTs owned by any principal (contract or STX address).

The endpoint can:
* Return the `tx_id` or complete `tx` object of the transaction that led that principal to own each NFT
* Filter by `asset_identifier` so users can receive a list of only one type of NFT (arties, megapont, etc.)
* Return anchored and unanchored results

Example responses (truncated):
```
GET /extended/v1/tokens/nft/holdings?principal=SPNWZ5V2TPWGQGVDR6T7B6RQ4XMGZ4PXTEE0VQ0S.marketplace-v3&offset=50

{
  "limit": 50,
  "offset": 50,
  "total": 1522,
  "results": [
    {
      "asset_identifier": "SPQZF23W7SEYBFG5JQ496NMY0G7379SRYEDREMSV.Candy::candy",
      "value": {
        "hex": "0x010000000000000000000000000000029d",
        "repr": "u669"
      },
      "tx_id": "0x3ff051cc6124480e9f9235e46e874c70a0cb678f065fca8a222a23d56e128b26"
    },
    {
      "asset_identifier": "SPQZF23W7SEYBFG5JQ496NMY0G7379SRYEDREMSV.Candy::candy",
      "value": {
        "hex": "0x0100000000000000000000000000000082",
        "repr": "u130"
      },
      "tx_id": "0x9ec8c2d6cc8ef590419c2328ece089b0a1dca87da2e789dfab9f41af6598bccb"
    },
    (truncated...)
```
```
GET /extended/v1/tokens/nft/holdings?principal=SP3BK1NNSWN719Z6KDW05RBGVS940YCN6X84STYPR&tx_metadata=true&asset_identifier=SP2X0TZ59D5SZ8ACQ6YMCHHNR2ZN51Z32E2CJ173.the-explorer-guild%3A%3AThe-Explorer-Guild

{
   "limit":50,
   "offset":0,
   "total":1,
   "results":[
      {
         "asset_identifier":"SP2X0TZ59D5SZ8ACQ6YMCHHNR2ZN51Z32E2CJ173.the-explorer-guild::The-Explorer-Guild",
         "value":{
            "hex":"0x0100000000000000000000000000000803",
            "repr":"u2051"
         },
         "tx":{
            "tx_id":"0x12e6f88724d0e630de26c376f172cf1781e25435e2b4ee54d36a949b244e429c",
            "nonce":7,
            "fee_rate":"812000",
            "sender_address":"SP3BK1NNSWN719Z6KDW05RBGVS940YCN6X84STYPR",
            "sponsored":false,
            "post_condition_mode":"deny",
            "post_conditions":[
               {
                  "type":"stx",
                  "condition_code":"sent_equal_to",
                  "amount":"100000000",
                  "principal":{
                     "type_id":"principal_standard",
                     "address":"SP3BK1NNSWN719Z6KDW05RBGVS940YCN6X84STYPR"
                  }
               }
            ],
            "anchor_mode":"any",
            "is_unanchored":false,
            "block_hash":"0x8a809916e35577337c3f9812a57d61666fde63932b1bdd1c4d1e2f24fb72d46c",
            (truncated...)
            "tx_type":"contract_call",
            "contract_call":{
               "contract_id":"SP2X0TZ59D5SZ8ACQ6YMCHHNR2ZN51Z32E2CJ173.the-explorer-guild-mint",
               "function_name":"claim",
               "function_signature":"(define-public (claim ))"
            }
         }
      }
   ]
}
```

Related to #846


docs/openapi.yaml
```diff
@@ -2851,6 +2854,68 @@ paths:
...
+          required: false
+          schema:
+            type: integer
+        - name: unanchored
+          in: query
+          description: whether or not to include tokens from unconfirmed transactions
+          required: false
+          schema:
+            type: boolean
+        - name: tx_metadata
```
zone117x, line 2903: Should this be something like a separate endpoint? Seems difficult to specify the spec/docs when the response payload schema can change based off a query param.

src/api/routes/tokens/tokens.ts
```diff
@@ -23,6 +29,76 @@ export function createTokenRouter(db: DataStore): express.Router {
...
+
+      const { results, total } = await db.getNftHoldings({
+        principal: principal,
+        assetIdentifiers: assetIdentifiers,
+        offset: offset,
+        limit: limit,
+        includeUnanchored: includeUnanchored,
+      });
+      const parsedResults: NonFungibleTokenHolding[] = await Promise.all(
+        results.map(async result => {
```
zone117x, line 70: Quite a lot of sql requests used up here. Should the `getNftHoldings` query JOIN against the txs table for these results? Or alternatively, fetch all txs by array in a single additional sql query?

zone117x: Also note this has two separate queries not contained with a a sql transaction, so the first one in `getNftHoldings` can return results for a block that ends up getting forked before the queries afterwards, resulting in either an error or mangled response data.

rafaelcr: > Quite a lot of sql requests used up here. Should the `getNftHoldings` query JOIN against the txs table for these results? Or alternatively, fetch all txs by array in a single additional sql query?

Interestingly, I found out that performing the JOIN against `txs` in the original query is much more expensive because it creates a lot of duplicate rows which then need to be trimmed by complex query techniques (e.g. once row produced per each `nft_event`<->`tx` possible combination, same problem we have in `/transactions`).

However, I agree with your idea that fetching by list might be the best way to go here. That fetch ends up being a `WHERE tx_id = ANY(x)` which is a much cheaper index lookup. I'll work on this.


> Also note this has two separate queries not contained within a single sql transaction (i.e. `BEGIN; <queries>; COMMIT;`).

Great point. I'll add this too.

docs/openapi.yaml
rafaelcr: You can see the response possibilities in Redoc but the UX is kinda bad because it's behind a couple of clicks (click on Responses > results here: https://deploy-preview-936--stacks-blockchain-api.netlify.app/#operation/get_nft_holdings). It's a good point, though, because I can't set two examples, only one.

@saralab what do you think it's best from a docs perspective? I took this behavior from some other metadata endpoints which can show/hide metadata fields in responses according to a query param, but maybe that's worth revisiting.

rafaelcr: Actually, scratch that 😎 
OpenAPI does support multiple response examples, check it out: https://deploy-preview-936--stacks-blockchain-api.netlify.app/#operation/get_nft_holdings

zone117x: Here is what the generated client code (`@stacks/blockchain-api-client`) looks like when using multiple response schemas:
image

There's a known bug with the generator where it can't create proper types with the complex `transaction` schema, so I tested the same multiple-response with another endpoint and getting the same result of `object`:

image

It looks like there's no way in OpenAPI to correlate/specify a query param with a response type:
https://stackoverflow.com/questions/36576447/swagger-specify-two-responses-with-same-code-based-on-optional-parameter

This isn't necessarily a deal breaker, but definitely something to consider.

rafaelcr: Yeah, I've been looking for a standard and haven't come up with anything solid. It's mostly based on opinion.

That said, this is the closest I found that endorses a dynamic response structure based on params (by Microsoft): https://docs.microsoft.com/en-us/onedrive/developer/rest-api/concepts/optional-query-parameters?view=odsp-graph-online

My opinion is that given that we're already changing responses in other endpoints we could just keep using that same practice in the future.

src/migrations/1640037852136_nft_custody_unanchored.ts
```diff
@@ -0,0 +1,30 @@
...
+import { MigrationBuilder, ColumnDefinitions } from 'node-pg-migrate';
+
+export const shorthands: ColumnDefinitions | undefined = undefined;
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  /**
+   * This view is identical to `nft_custody` but it will only refreshed during microblock updates
+   * to calculate an unanchored snapshot of NFT custody.
+   */
+  pgm.createMaterializedView('nft_custody_unanchored', {}, `
```
zone117x, line 11: I could be misreading -- but where is `nft_custody_unanchored` refreshed?

src/tests/token-tests.ts
```diff
@@ -0,0 +1,147 @@
...
+    const request6 = await supertest(api.server).get(
+      `/extended/v1/tokens/nft/holdings?principal=${addr2}`
+    );
+    expect(request6.status).toBe(200);
+    expect(request6.type).toBe('application/json');
+    const result6 = JSON.parse(request6.text);
+    expect(result6.total).toEqual(1);
+    expect(result6.results[0].asset_identifier).toEqual(assetId2);
+    expect(result6.results[0].tx_id).toEqual('0x5484');
+  });
```
zone117x, line 139: Should we also include tests that insert microblocks and query for `?unanchored=true` ?

rafaelcr: Done

src/migrations/1640037852136_nft_custody_unanchored.ts
rafaelcr: Whoops, good catch. I also found a bug so I had to revisit the refresh logic, check it out.

src/datastore/postgres-store.ts
```diff
@@ -4745,6 +4809,44 @@ export class PgDataStore
...
+        `
+        WITH anchored_height AS (SELECT MAX(block_height) AS anchored FROM nft_custody),
+          unanchored_height AS (SELECT MAX(block_height) AS unanchored FROM nft_custody_unanchored)
+        SELECT unanchored > anchored AS outdated
+        FROM anchored_height CROSS JOIN unanchored_height
+        `
+      );
+      if (outdated) {
+        await this.refreshMaterializedView(client, 'nft_custody');
+      }
```
zone117x, line 4846: The above code looks right. Are you confident the unanchored tests are good enough coverage or should we add some re-org tests as well?

zone117x (approved):
This looks great @rafaelcr! Had a last question about testing some logic around the upsert code related to re-orgs, but otherwise LGTM

src/datastore/postgres-store.ts
rafaelcr: Good looking out. I added more tests and a fix for only considering NFT events from canonical txs.

rafaelcr:
Added tests for re-orgs @zone117x here https://github.com/hirosystems/stacks-blockchain-api/pull/936/commits/4d8c54a0992332166b599a10bbae21338d9b2ea1
Can you take a quick look, please?

zone117x (approved):
Awesome, thanks @rafaelcr!

PR #939, opened 2021-12-22
fix: pin openAPI generator-cli version

zone117x:
Fix doc builds in CI by pinning the `openapi-generator` version. It had an update yesterday that broke our scripts.

Following the guide from https://github.com/OpenAPITools/openapi-generator-cli
> The first time you run the command `openapi-generator-cli` the last stable version of `OpenAPITools/openapi-generator` is downloaded by default. That version is saved in the file _openapitools.json_. Therefore you should include this file in your version control, to ensure that the correct version is being used next time you call the command.


rafaelcr (approved):
🏁

Issue #941, opened 2021-12-22
Install client dir as part of CI


CharlieC3:
**Describe the bug**
[CI failed during a release](https://github.com/hirosystems/stacks-blockchain-api/runs/4609298332?check_suite_focus=true#step:3:466) because as part of the release, `npm install` is ran inside the `client` directory. This was unexpected because CI passed in PRs leading up to the release.

Add a step in CI to run `npm install` in both the `client` and `docs` directories to catch any potential installation issues that would occur during a release.

PR #942, opened 2021-12-22
ci: test installing client and docs deps

CharlieC3:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/941

## Description
Adds a couple steps to the CI `test` job which installs dependencies in the `client` and `docs` directories. If there are any issues during the installation, they should now be caught before making their way to the `master` branch.

I also removed the notification jobs in the main CI workflow since they're not helpful anymore as Hiro doesn't use Slack.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

### DIFF
--- a/.github/workflows/stacks-blockchain-api.yml
+++ b/.github/workflows/stacks-blockchain-api.yml
@@ -13,23 +13,6 @@ on:
   workflow_dispatch:
 
 jobs:
-  notify-start:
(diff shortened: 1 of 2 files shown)


rafaelcr (approved):
Great

zone117x (approved):
ty!

PR #945, opened 2021-12-23
npm audit fixes, dependency cleanup

zone117x:
Fixes several high severity warnings reported by `npm audit`. The `package-lock.json` files were having issues with simple `npm audit fix` commands due to unused deps and deps with major version / breaking changes.

These have been removed or updated, with a couple small required code adjustments. 

Audit now passes on the root package.json, and client/package.json. Still working on docs/package.json which is using quite a bit of old and unmaintained deps.


rafaelcr (approved):
Thanks

PR #946, opened 2021-12-24
fix: create latest_contract_txs materialized view for unanchored blocks

rafaelcr:
Fixes a regression in the `/extended/v1/:addr/transactions` endpoint where if a smart contract had more than 50 transactions in the *anchored* block, and then it received some new transactions (for example 10) in the next *unanchored* block, they both were bundled into the same materialized view that has a limit of 50 cached transactions.

After this happened, if somebody requested the first 50 **anchored** transactions, the API sent back only 40 (50 minus the 10 unanchored txs at the next block).

The solution is to create a copy of the materialized view that is only refreshed in unanchored situations and then choose which one to query when the endpoint is called.

### DIFF
--- a/src/api/routes/address.ts
+++ b/src/api/routes/address.ts
@@ -221,7 +221,7 @@ export function createAddressRouter(db: DataStore, chainId: ChainID): RouterWith
       offset,
       blockHeight,
       atSingleBlock,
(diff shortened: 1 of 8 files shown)


rafaelcr:
This PR is obsoleted by #951 but it might still be worth to push this hotfix if required

PR #951, opened 2021-12-30
feat: add `principal_stx_txs` table to speed up `/transfers` endpoint

rafaelcr:
**TL;DR:**
1. Remove the `latest_contract_txs` materialized view
1. Create the `principal_stx_txs` table that keeps (`principal`,`tx_id`,`block_height`) records of transactions that affected the principal's STX balance.
1. Populate table with each new canonical transaction.
1. Use the new table to query `/transactions` much faster using a postgres index.
1. Add chain tip cache headers to `/transactions`

### Background

We're currently having a lot of performance issues in the `/transactions` endpoint because the current SQL query to retrieve results involves a very expensive cross-lookup between the `txs` and `stx_events` tables to identify and return transactions that affected the principal's STX balance. Lately, this endpoint has been used heavily by NFT issuers to determine the new minting and/or ownership of assets, typically by calling the endpoint every few seconds to get the newest 50 transactions (`limit=50`).

To mitigate this problem, we created and shipped a postgres materialized view that kept a record of the last 50 transactions for every smart contract (#849), refreshed once every block. This worked very well initially and greatly improved endpoint response times, but the problem started again when very popular NFT mints caused these contracts to have more than 50 new transactions every block (even hundreds), forcing NFT issuers to ask for multiple transaction pages or increase the `limit` parameter thus avoiding the materialized view and resurfacing the original problem. The fact that we started to retrieve full `abi` for smart contract calls (#807) on every request also exacerbated the situation.

### Solution

The new strategy proposed by this PR is to remove the materialized view in favor of a full table (`principal_stx_txs`) which will keep a record of all *canonical* transaction ids that affected every principal's STX balance throughout the full history of the chain. Using this table with the right index, we can query for STX-affecting transaction ids for any principal almost instantly and at any block height. Once we do that, we can join results against the `txs` table to get the full objects we need to return. Even though this table will have a large number of records, we will never need to query anything more than the index.

Additionally, this PR adds the chain tip cache headers to `/transactions` to avoid re-querying for the same smart contracts until the next block arrives.

### Performance

An initial performance comparison between the old (current) query and the new query using the new table suggests a great improvement. Note that these results are affected by the `abi` column lookup for smart contracts. Using `pgbench`:

Original query
```
transaction type: /Users/rafaelcr/Desktop/1.sql
scaling factor: 1
query mode: simple
number of clients: 96
number of threads: 8
duration: 15 s
number of transactions actually processed: 1956
latency average = 759.204 ms
initial connection time = 120.811 ms
tps = 126.448223 (without initial connection time)
```

New query
```
transaction type: /Users/rafaelcr/Desktop/2.sql
scaling factor: 1
query mode: simple
number of clients: 96
number of threads: 8
duration: 15 s
number of transactions actually processed: 3806
latency average = 381.744 ms
initial connection time = 111.835 ms
tps = 251.477347 (without initial connection time)
```


rafaelcr:
cc @saralab

src/tests/api-tests.ts
```diff
@@ -4699,6 +4699,45 @@ describe('api tests', () => {
...
+          },
+          event_count: 5,
+          events: [],
+          execution_cost_read_count: 0,
+          execution_cost_read_length: 0,
+          execution_cost_runtime: 0,
+          execution_cost_write_count: 0,
+          execution_cost_write_length: 0,
+        },
         {
```
zone117x, line 4741: Should we include the anchored unit tests from PR https://github.com/hirosystems/stacks-blockchain-api/pull/946 if we go ahead with merging this one and not the other?

zone117x (approved):
This looks like a huge win, nice work @rafaelcr!

src/tests/api-tests.ts
rafaelcr: Yep, absolutely. I'll add more tests and get back to you.

Issue #896, opened 2021-12-10
Mounting the API documentation on /doc


lgalabru:
When developers are using `clarinet integrate` we spawn a bitcoin-node, a stacks-node and a stacks-api (+ explorer) - the goal here is to create a network, 100% offline. 
We are already displaying the URL of the API in the Terminal UI - http://localhost:3999. When developers click on this URL, they'll see a health check, pretty much useless. 
Instead we would like to display http://localhost:3999/doc in the TUI, which would open a local version https://hirosystems.github.io/stacks-blockchain-api , so that developers can discover the API by themselves.
This could be guarded by an ENV VAR.

PR #954, opened 2022-01-03
Mount documentation on /doc

MHassanTariq:
## Description

This PR creates an endpoint to locally display OpenApi.yaml using [stagger-ui-express](https://www.npmjs.com/package/swagger-ui-express).
For details refer to issue #896 

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


src/api/init.ts
```diff
@@ -149,6 +152,10 @@ export async function startApiServer(opts: {
     res.redirect(`/extended/v1/status`);
   });
 
+  if (!isProdEnv) {
```
rafaelcr, line 155: We should serve this on prod too... thoughts @CharlieC3 @saralab ?

src/api/init.ts
```diff
@@ -43,6 +43,9 @@ import { createStatusRouter } from './routes/status';
 import { createTokenRouter } from './routes/tokens/tokens';
 import { createFeeRateRouter } from './routes/fee-rate';
 import { setResponseNonCacheable } from './controllers/cache-controller';
+import * as swaggerUi from 'swagger-ui-express';
+import * as yaml from 'yamljs';
+const swaggerDocument = yaml.load('docs/openapi.yaml');
```
rafaelcr, line 48: `/doc` should not return the YAML document but the actual Redoc HTML generated documentation

rafaelcr (changes requested):

src/api/init.ts
CharlieC3: I don't see why it shouldn't always be enabled. Is there any reason this enablement should be controlled by an ENV var?

zone117x (changes requested):
We're already generating a full Redoc/html directory from the openAPI+schemas via the scripts in `/docs/package.json`, shouldn't we just use that output and serve it?

zone117x (changes requested):
We're already generating a full Redoc/html directory from the openAPI+schemas via the scripts in `/docs/package.json`, shouldn't we just use that output and serve it?

zone117x:
@rafaelcr @MHassanTariq @saralab something that comes to mind after doing the `npm audit fix` [PR](https://github.com/hirosystems/stacks-blockchain-api/pull/945) -- we probably shouldn't expose this docs endpoint in production because we'd be serving html+js files as opposed to just a JSON REST API and therefore greatly increasing the attack surface of the API and the `*.stacks.co` domain properties. The much of the npm tooling used to generate these docs in `/docs/package.json` are years old, no longer maintained, and have critical security vulnerabilities reported by npm audit. So I think the default behavior should be if `NODE_ENV == production` then do not serve these files.

rafaelcr:
> greatly increasing the attack surface of the API and the `*.stacks.co` domain properties.

@zone117x That's a very good point, hadn't thought of it. I wonder if for production we could just do a redirect with nginx from `/docs` to https://docs.hiro.so/api ? Other people might run their API instances so I don't think it would be a good idea to make that redirect to Hiro's docs with express

package.json
```diff
@@ -21,9 +21,10 @@
     "test:integration:microblocks": "npm run devenv:deploy:pg -- -d && cross-env NODE_ENV=development jest --config ./jest.config.microblocks.js --coverage --no-cache --runInBand; npm run devenv:stop:pg",
     "test:integration:tokens": "npm run devenv:deploy -- -d && cross-env NODE_ENV=development jest --config ./jest.config.tokens.js --coverage --no-cache --runInBand; npm run devenv:stop",
     "git-info": "echo \"$(git rev-parse --abbrev-ref HEAD)\n$(git log -1 --pretty=format:%h)\n$(git describe --tags --abbrev=0)\" > ./.git-info",
-    "build": "npm run git-info && rimraf ./lib && tsc -p tsconfig.build.json",
+    "build": "npm run git-info && npm run build:api-documentation && rimraf ./lib && tsc -p tsconfig.build.json",
     "build:tests": "tsc -p tsconfig.json",
     "build:docs": "npm i --prefix docs && npm run build --prefix docs && npm i --prefix client && npm run generate:docs --prefix client",
+    "build:api-documentation": "npm i --prefix docs && npm run build --prefix docs",
```
zone117x, line 27: Any reason for this new command rather than using the existing `build:docs` one?

package.json
```diff
@@ -21,9 +21,10 @@
     "test:integration:microblocks": "npm run devenv:deploy:pg -- -d && cross-env NODE_ENV=development jest --config ./jest.config.microblocks.js --coverage --no-cache --runInBand; npm run devenv:stop:pg",
     "test:integration:tokens": "npm run devenv:deploy -- -d && cross-env NODE_ENV=development jest --config ./jest.config.tokens.js --coverage --no-cache --runInBand; npm run devenv:stop",
     "git-info": "echo \"$(git rev-parse --abbrev-ref HEAD)\n$(git log -1 --pretty=format:%h)\n$(git describe --tags --abbrev=0)\" > ./.git-info",
-    "build": "npm run git-info && rimraf ./lib && tsc -p tsconfig.build.json",
+    "build": "npm run git-info && npm run build:api-documentation && rimraf ./lib && tsc -p tsconfig.build.json",
```
zone117x, line 24: I'm conflicted about making this a build requirement for the root project. For example, one of the doc generators requires the java runtime. Seems like the dockerfiles (and whatever else) should just explicitly call a build:docs command after this initial one.

src/helpers.ts
```diff
@@ -25,6 +25,7 @@ export const isProdEnv =
   process.env.NODE_ENV === 'prod' ||
   !process.env.NODE_ENV ||
   (!isTestEnv && !isDevEnv);
+export const apiDocumentationUrl = process.env.API_DOCS_URL;
```
zone117x, line 28: Could you add an example and commented-out line and quick description in the `.env` file for this var? That's the main way we're "documenting" env vars right now.

package.json
MHassanTariq: But it is important that this command runs before the server starts. So we have the HTML ready to be rendered. Through this, upon running `dev:integrated`, it will run `build` which runs `build:api-documentation`. 

For explicitly running it, do you think I should print an error message like `Please run the command: npm run build:docs and restart your server` for endpoint `/doc`, in case the server starts without building documents (i.e. HTML file was not created)?

src/api/init.ts
```diff
@@ -149,6 +150,18 @@ export async function startApiServer(opts: {
     res.redirect(`/extended/v1/status`);
   });
 
+  // in case env is not production and url is not provided
+  if (apiDocumentationUrl) {
```
zone117x, line 154: These conditions should be inside a single route registration. Also, if the `/doc` endpoint is not enabled, it should return a 404 with a useful message about docs not being available.

src/api/init.ts
```diff
@@ -149,6 +150,18 @@ export async function startApiServer(opts: {
...
+    app.use('/doc', (req, res) => {
+      res.redirect(apiDocumentationUrl ?? '');
+    });
+  } else if (!isProdEnv) {
+    app.use('/doc', (req, res) => {
+      const apiDocumentationPath = path.join(__dirname + '../../../docs/.tmp/index.html');
```
zone117x, line 160: Is this directory included/available in the docker images?

package.json
zone117x: The other projects like Clarinet do not compile the API from source. The docker images are used, so those docker images should be responsible for ensuring the docs are built and included. For those working on this repo, it doesn't make sense to require that the core `build` script do this extra work. 

> For explicitly running it, do you think I should print an error message like ...

Yes if the doc endpoint is supposed to be served, but the `index.html` file doesn't exist, this would be a helpful error response.

src/api/init.ts
MHassanTariq: image

I created a new docker image after running `npm run build:docs` and analyzed its directories. This path is valid and directory is included in the docker image.

zone117x:
Looks like `package-lock.json` file changes need to be fully reverted

MHassanTariq:
Closing this PR due to merge failures which are resulting in random test failures. 
Continuing work in PR #991

Issue #955, opened 2022-01-04
Rosetta /account/balance call is returning an incremented nonce one block too early


kmur911:
**Describe the bug**
For these two transactions ([1](https://explorer.stacks.co/txid/0xe2a56741db5ea3bad7b50cd9df544ff60d64a94f13eecc433c779076cbf440dd?chain=testnet) [2](https://explorer.stacks.co/txid/0x65b5488940c36e91536f7057b14a892faabc10a2868fea57750f3fb582b172dd?chain=testnet)) we've seen the `/account/balance` call return the incremented nonce one block before the transaction was finalized.

For example, for the [first](https://explorer.stacks.co/txid/0xe2a56741db5ea3bad7b50cd9df544ff60d64a94f13eecc433c779076cbf440dd?chain=testnet) the explorer shows that the transaction landed (incrementing the nonce from 173->174) in block #21499. However, when calling `/account/balance` and getting back the result for block #21498, it returned a nonce of 174, when it should only have returned that at block #21499. It's worth pointing out that the returned balance at block #21498 was correct (i.e it didn't factor in the balance change from the transaction), it's only the nonce that seems to have been incremented one block prematurely.

For the [second](https://explorer.stacks.co/txid/0x65b5488940c36e91536f7057b14a892faabc10a2868fea57750f3fb582b172dd?chain=testnet) the explorer shows that the transaction landed (incrementing the nonce from 0->1) in block #26289. However, when calling `/account/balance` and getting back the result for block #26288, it returned a nonce of 1, when it should only have returned that at block #26289. Same as before, the balance at #26288 was correct (9.174474 STX), it didn't factor in the balance change from the transaction, it's only the nonce that seems to have been incremented one block prematurely.

**To Reproduce**
It's hard to reproduce this issue because it's intermittent, not happening for every account we control. However it is only happening on accounts where we're generating outgoing transactions, so I'm wondering if the node which broadcasts the transaction is jumping the gun on the nonce increment by 1 block. For example, we run 4 stacks nodes, and alternate between them when fetching `/account/balance` information. I wonder if we broadcast the transaction through one node, would the nonce returned by that node's `/account/balance` call increment by 1 a block earlier than the other nodes?

**Expected behavior**
`/account/balance` for all nodes should only return the incremented nonce at the block height in which the transaction is finalized for all nodes

rafaelcr:
[...]

PR #959, opened 2022-01-05
Hotfix/rosetta account nonce

zone117x:
Fixes https://github.com/hirosystems/stacks-blockchain-api/issues/955

> @rafaelcr yep you're right. That route queries postgres for info specific to a given block (if provided in the POST request), _except_ for the nonce fetch which unconditionally fetches "latest" from the stacks-node:
https://github.com/hirosystems/stacks-blockchain-api/blob/7a314437717e62aeaeb34d11de653364e1476acf/src/api/routes/rosetta/account.ts#L66-L70
> 
> 
> ```ts
> // BAD: queries stacks-node for account info at block tip, ignoring any specified request block
> const accountInfo = await new StacksCoreRpcClient().getAccount(accountIdentifier.address);
> const nonce = accountInfo.nonce ?? 0;
> ```

Created modified version of `db.getAddressNonces(...)` to determine the account nonce at a given height via postgres data.

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -1724,10 +1724,10 @@ export class PgDataStore
 
   async getAddressNonceAtBlock(args: {
     stxAddress: string;
-    blockIdentifer: BlockIdentifier;
+    blockIdentifier: BlockIdentifier;
   }): Promise<FoundOrNot<{ nonce: number }>> {
     return await this.queryTx(async client => {
-      const dbBlock = await this.getBlockInternal(client, args.blockIdentifer);
+      const dbBlock = await this.getBlockInternal(client, args.blockIdentifier);
       if (!dbBlock.found) {
         return { found: false };
       }
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -1722,6 +1722,30 @@ export class PgDataStore
     });
   }
 
+  async getAddressNonceAtBlock(args: {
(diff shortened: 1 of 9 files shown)


src/datastore/postgres-store.ts
```diff
@@ -1722,6 +1722,30 @@ export class PgDataStore
...
+    return await this.queryTx(async client => {
+      const dbBlock = await this.getBlockInternal(client, args.blockIdentifier);
+      if (!dbBlock.found) {
+        return { found: false };
+      }
+      const executedTxNonce = await client.query<{ nonce: number | null }>(
+        `
+        SELECT MAX(nonce) nonce
+        FROM txs
+        WHERE ((sender_address = $1 AND sponsored = false) OR (sponsor_address = $1 AND sponsored = true))
```
rafaelcr, line 1738: I'm curious if we need a new index for this since it's two AND conditions... I realize this is a hotfix to master so fixing here would be impossible if it's required, but we should keep this in mind if we're going to be querying for more sponsored txs in the future.

rafaelcr (approved):
LGTM but left one note for maybe another PR

src/datastore/postgres-store.ts
zone117x: Yep -- this was ripped from the similar query in develop where the index was added. The current indexes make this pretty quick so I'm not concerned. Once develop branch is merged it will be optimized further.

Issue #961, opened 2022-01-05
API is returning incorrect sequence number after update to v1.0.5


kmur911:
**Describe the bug**
I believe this is related to the fix shipped in https://github.com/hirosystems/stacks-blockchain-api/releases/tag/v1.0.5.

After updating the node api to v1.0.5 (redeploying using an existing node snapshot), making the call to /account/balance with the following POST body
```
{
  "network_identifier": {
    "blockchain": "stacks",
    "network": "testnet"
  },
  "account_identifier": {
    "address": "ST3XNSE77C2J2EJ7ESVXHTG3904C7GRNKH4CCM85K"
  },
  "block_identifier": {
    "hash": "0x74b6f26700ae744d2f4a21965424d65b1c74214ba48a33bae30f78605ef6ff9a",
    "index": 27041
  }
}
```

I'm getting the response
```
{
    "block_identifier": {
        "index": 27041,
        "hash": "0x74b6f26700ae744d2f4a21965424d65b1c74214ba48a33bae30f78605ef6ff9a"
    },
    "balances": [
        {
            "value": "120",
            "currency": {
                "symbol": "STX",
                "decimals": 6
            }
        }
    ],
    "metadata": {
        "sequence_number": 0
    }
}
```
even tho address [ST3XNSE77C2J2EJ7ESVXHTG3904C7GRNKH4CCM85K](https://explorer.stacks.co/address/ST3XNSE77C2J2EJ7ESVXHTG3904C7GRNKH4CCM85K?chain=testnet) has a nonce of 1 and has since [this](https://explorer.stacks.co/txid/0x603328f0d593786a31763db97814158a8d9d6535ebdad332f3810b6c24c4a0e9?chain=testnet) transaction in block 26650 3 days ago.
I wonder if the switch away from using the live nonce to pulling it from the local node db has exposed some inconsistency with how the nonces are tracked in the db for historical blocks?

**To Reproduce**
1. Update node to api v1.0.5
2. Make the POST to /account/balance with the body above
3. Verify that it returns the incorrect nonce

**Expected behavior**
The nonce should be 1 for this address (this issue is happening for multiple addresses, I've just included one for simplicity).

**Desktop (please complete the following information):**
stacks-blockchain-api v1.0.5
stacks-blockchain node 2.05.0.0.0

zone117x:
[...]

PR #964, opened 2022-01-07
fix(rosetta): off-by-one nonce returned with rosetta /account/balance…

zone117x:
Fixes https://github.com/hirosystems/stacks-blockchain-api/issues/961

The last used nonce was being returned as opposed to the next nonce.


rafaelcr (approved):
LGTM

PR #970, opened 2022-01-11
fix: reconnect broken pgnotify clients

### LINKED ISSUE #960: Database connection errors after DB restart
**Describe the bug**
The API logs errors on what seems to be all database interactions after a scenario where the database connection is severed (e.g. database restarts), and then re-established. Despite these errors logged, the API is still reporting 200-level status codes and is able to perform reads/writes to the database.

Below log snippets may be out of order, but meant to roughly show what's going on:

```
{"level":"error","message":"Error sending PgNotifier notification of type: microblockUpdate Client has encountered a connection error and is not queryable","stack":"Error: Client has encountered a connection error and is not queryable\n    at /app/node_modules/pg/lib/client.js:563:27\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)","timestamp":"2022-01-05T18:43:18.878Z"}
...
{"level":"error","message":"Error sending PgNotifier notification of type: addressUpdate Client has encountered a connection error and is not queryable","stack":"Error: Client has encountered a connection error and is not queryable\n    at /app/node_modules/pg/lib/client.js:563:27\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)","timestamp":"2022-01-05T18:54:53.457Z"}
...
{"level":"error","message":"Error sending PgNotifier notification of type: txUpdate Client has encountered a connection error and is not queryable","stack":"Error: Client has encountered a connection error and is not queryable\n    at /app/node_modules/pg/lib/client.js:563:27\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)","timestamp":"2022-01-05T18:54:53.645Z"}
...
# Still returns 200 status codes though despite errors
{"req":{"url":"/new_mempool_tx","headers":{"host":"my-api-host.com:3700","content-length":"642","date":"Wed, 05 Jan 2022 18:48:00 GMT","content-type":"application/json"},"method":"POST","httpVersion":"1.1","originalUrl":"/new_mempool_tx","query":{}},"res":{"statusCode":200},"responseTime":17,"level":"http","message":"HTTP POST /new_mempool_tx","timestamp":"2022-01-05T18:48:00.619Z"}
```

**To Reproduce**
1. Run an API + stacks-node + database
2. Restart the database
3. Observe API logs. It should log to severed connection, re-connect once the database boots back up, and continue printing error logs on every subsequent database interaction.

**Expected behavior**
Once the database connection is re-established, the API should not encounter any further database errors.

API Version: `1.0.4`

rafaelcr:
Configure `pg-listen` reconnection settings to retry indefinitely while the database is restarting.

Fixes #960

### DIFF
--- a/src/datastore/postgres-notifier.ts
+++ b/src/datastore/postgres-notifier.ts
@@ -54,17 +54,27 @@ type PgNotificationCallback = (notification: PgNotification) => void;
  * https://www.postgresql.org/docs/12/sql-notify.html
  */
 export class PgNotifier {
-  readonly pgChannelName: string = 'pg-notifier';
-  subscriber: Subscriber;
+  private readonly pgChannelName: string = 'pg-notifier';
+  private readonly clientConfig: ClientConfig;
+  private eventCallback?: PgNotificationCallback;
+  private subscriber?: Subscriber;
 
   constructor(clientConfig: ClientConfig) {
-    this.subscriber = createPostgresSubscriber(clientConfig);
+    this.clientConfig = clientConfig;
   }
 
   public async connect(eventCallback: PgNotificationCallback) {
-    this.subscriber.notifications.on(this.pgChannelName, message =>
-      eventCallback(message.notification)
-    );
+    this.eventCallback = eventCallback;
+    await this.doConnect();
+  }
+
+  async doConnect() {
+    this.subscriber = createPostgresSubscriber(this.clientConfig);
+    this.subscriber.notifications.on(this.pgChannelName, message => {
+      if (this.eventCallback) {
+        this.eventCallback(message.notification);
+      }
+    });
     this.subscriber.events.on('error', error => logError('Fatal PgNotifier error', error));
     await this.subscriber.connect();
     await this.subscriber.listenTo(this.pgChannelName);
@@ -99,15 +109,30 @@ export class PgNotifier {
   }
 
   public async close() {
-    await this.subscriber.unlisten(this.pgChannelName);
-    await this.subscriber.close();
+    await this.subscriber?.unlisten(this.pgChannelName);
+    await this.subscriber?.close();
   }
 
   private async notify(notification: PgNotification) {
-    await this.subscriber
-      .notify(this.pgChannelName, { notification: notification })
-      .catch(error =>
-        logError(`Error sending PgNotifier notification of type: ${notification.type}`, error)
-      );
+    if (this.subscriber) {
+      await this.subscriber
+        .notify(this.pgChannelName, { notification: notification })
+        .catch(async error => {
+          if (
+            error instanceof Error &&
+            error.message === 'Client has encountered a connection error and is not queryable'
+          ) {
+            // The postgres client has lost the connection or has otherwise become corrupt. We will attempt to
+            // reconnect so we can continue sending events.
+            // There's no point in calling `unlisten` before closing the current client since it requires a query.
+            await this.subscriber?.close();
+            await this.doConnect();
+            // Try again now.
+            await this.subscriber?.notify(this.pgChannelName, { notification: notification });
+          } else {
+            logError(`Error sending PgNotifier notification of type: ${notification.type}`, error);
+          }
+        });
+    }
   }
 }
--- a/src/datastore/postgres-notifier.ts
+++ b/src/datastore/postgres-notifier.ts
@@ -118,10 +118,7 @@ export class PgNotifier {
       await this.subscriber
         .notify(this.pgChannelName, { notification: notification })
         .catch(async error => {
-          if (
-            error instanceof Error &&
-            error.message === 'Client has encountered a connection error and is not queryable'
-          ) {
+          if (error instanceof Error && error.message.includes('not queryable')) {
             // The postgres client has lost the connection or has otherwise become corrupt. We will attempt to
             // reconnect so we can continue sending events.
             // There's no point in calling `unlisten` before closing the current client since it requires a query.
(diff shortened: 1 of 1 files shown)


src/datastore/postgres-notifier.ts
```diff
@@ -99,15 +109,30 @@ export class PgNotifier {
...
-      .catch(error =>
-        logError(`Error sending PgNotifier notification of type: ${notification.type}`, error)
-      );
+    if (this.subscriber) {
+      await this.subscriber
+        .notify(this.pgChannelName, { notification: notification })
+        .catch(async error => {
+          if (
+            error instanceof Error &&
+            error.message === 'Client has encountered a connection error and is not queryable'
```
rafaelcr, line 123: @zone117x this is horrible lol but unfortunately `pg` doesn't throw anything useful for me to catch here other than an `Error` with that specific string... suggestions are welcome for how I can improve this

zone117x: Ah, we have existing similarly horrible code here due to the pg lib not throwing very good Error types: https://github.com/hirosystems/stacks-blockchain-api/blob/a0fd2fe0baf396866f5ca8bb8b9ab91807f02349/src/datastore/postgres-store.ts#L673-L688

I'm actually confused now looking at the `pg` lib source. The error strings you've included are from `node_modules/pg/lib/native/client.js`, using `pg-native` which I thought was an opt-in mode that we don't use. 

For reference, the existing pg connection error handling code is checking strings defined in `node_modules/pg/lib/client.js` (non-native version). 

Does the pg-notifier lib opt-in to the native version, or is there some automatic process where sometimes it uses native when possible?

The error string seem different between the new two modes so that's the only concern I have here.

rafaelcr: I took that string from the error message @CharlieC3 reported:
```
{"level":"error","message":"Error sending PgNotifier notification of type: microblockUpdate Client has encountered a connection error and is not queryable","stack":"Error: Client has encountered a connection error and is not queryable\n    at /app/node_modules/pg/lib/client.js:563:27\n    at processTicksAndRejections (node:internal/process/task_queues:78:11)","timestamp":"2022-01-05T18:43:18.878Z"}
```
I basically went to that exact location in `pg/lib/client.js` to get the string

zone117x: Ah okay, looks good to me then!

rafaelcr:
Ready, but needs testing in a canary or specialized environment before merge

src/datastore/postgres-notifier.ts
```diff
@@ -99,15 +109,27 @@ export class PgNotifier {
...
-    await this.subscriber
-      .notify(this.pgChannelName, { notification: notification })
-      .catch(error =>
-        logError(`Error sending PgNotifier notification of type: ${notification.type}`, error)
-      );
+    if (this.subscriber) {
+      await this.subscriber
+        .notify(this.pgChannelName, { notification: notification })
+        .catch(async error => {
+          if (error instanceof Error && error.message.includes('not queryable')) {
```
zone117x, line 121: What do you think of some of the reconnect and error related events https://github.com/andywer/pg-listen#instanceeventsonreconnect-listener-attempt-number--void ?

src/datastore/postgres-notifier.ts
```diff
@@ -99,15 +109,27 @@ export class PgNotifier {
...
+            // The postgres client has lost the connection or has otherwise become corrupt. We will attempt to
+            // reconnect so we can continue sending events.
+            // There's no point in calling `unlisten` before closing the current client since it requires a query.
+            await this.subscriber?.close();
+            await this.doConnect();
```
zone117x, line 126: Does `doConnect` have retry logic? If so, it seems like we could hit this condition, then try to connect while the db is still unavailable, and end up in another error state?

rafaelcr: That's a good question. It doesn't have logic in this file but probably in the library. I'll take a look.

rafaelcr: It does try to reconnect on error and it retries an infinite number of times until it succeeds, so I think we're good: https://github.com/andywer/pg-listen/blob/master/src/index.ts#L93-L125

However, I can see a scenario where we lose some notifications that would come while a reconnection is still pending (the code as it is right now would only retry the original notification that triggered the reconnect).

I could add some sort of queue that would retry any notification on failure up to a fixed number of times, but that may be outside the scope of this PR. Thoughts?

zone117x: The dropped events seem okay to me in this scenario. Were you able to reproduce this behavior locally? Right now this error state results in the API still working (without websockets), we'd just want to ensure a possible regression around here didn't make things worse.

rafaelcr: I was just able to reproduce this locally, it came down to restarting the database just at the right time when a notification was about to be sent.

As it turns out, the library **does** reconnect when this error happens but it still logs the error. Adding this `close()`/`doConnect()` combination was revealed to be useless because `close()` actually **destroys** the whole error callback without finishing it 🤣 so `doConnect()` is never really called.

@CharlieC3 when you saw these errors on #960, did they continue to appear constantly from then on? Or did they just appear while the database was being restarted? My guess is that the series of errors you were seeing were from all the different message subscriptions that tried to notify at the same time or very close to each other, so this may actually be working as intended.

zone117x (approved):
LGTM! 

Note: this change also seems to help with one of the errors reported in https://github.com/hirosystems/stacks-blockchain-api/issues/1038

PR #971, opened 2022-01-12
fix: ensure `Cache-Control` is specified in 304 response

zone117x:
Fixes caching issue with certain CDNs (e.g. Cloudflare) where intermediate caching will only be performed if the `304 CACHE OK` response _also_ includes the `Cache-Control` headers.

Cloudflare will perform intermediate caching when the following page rules are applied for `/v1/extended/*`:
image


rafaelcr (approved):
👍

CharlieC3 (approved):
:shipit:

PR #972, opened 2022-01-12
feat: add write-only mode

### LINKED ISSUE #965: add write-only mode that only runs the event server to populate the pg database
The API currently has two operation modes:

1. `read-only` mode (#769) that enables us to run the API without the event server that consumes RPCs from a stacks node
2. `read-write` mode that runs both the event server and the API stack (default)

We'd like to add a third `write-only` mode that **only** runs the event server to populate the postgres database, so we can have dedicated write-only instances completely separate from the read-only APIs that are responding to external requests. This would also help all developers and community members that are interested in consuming data from the database (which already takes care of block re-orgs, etc.) without having to run the entire API frontend.

This would be the first step toward making the event server a separate micro-service and/or node module, and toward fragmenting the API into discrete micro-services.

rafaelcr:
- Add `writeonly` API mode
- Unify `readonly` and `offline` modes into a single ENV setting alongside `writeonly`
- Update README

Closes #965 

TODO: Test in a dedicated deployment to measure performance and any other considerations that could've been missed cc @CharlieC3

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,7 @@ import {
   isProdEnv,
   numberToHex,
   httpPostRequest,
-  isReadOnlyMode,
+  parseArgBoolean,
 } from './helpers';
 import * as sourceMapSupport from 'source-map-support';
 import { DataStore } from './datastore/common';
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -30,9 +30,53 @@ import { Socket } from 'net';
...
+    case 'readonly':
+      return StacksApiMode.readOnly;
+    case 'writeonly':
+      return StacksApiMode.writeOnly;
+    case 'offline':
+      return StacksApiMode.offline;
+    default:
+      break;
+  }
+  // Make sure we're backwards compatible if `STACKS_API_MODE` is not specified.
```
zone117x, line 70: Nice 👍

zone117x (approved):
Awesome! And great attention to the backwards-compat. LGTM!

CharlieC3 (approved):
No issues found in the deployment; it was able to sync to the chain tip and progress along. I did notice that it only uses about 20% of the CPU a full API normally would at idle, and about 10MB less memory as well.

Nice job @rafaelcr! :shipit:

PR #973, opened 2022-01-12
feat: exit immediately when stacks-node block desync detected

zone117x:
(Note: for now, this is primarily a testing PR for debugging desync issues)

This PR changes the behavior for when the API encounters a "desync" error with the `stacks-node` (event emitter). A desync being: when the stacks-node emits a `/new_block` event containing a pointer to a parent block that the API has _not_ received from any previous `/new_block` events. Typically this happens when some out-of-order or buggy shutdown sequence occurs between the API and `stacks-node`.

The API is unable to handle re-orgs and serve data in a "lossy" mode with missing blocks/txs. So prior to this PR, the API would return an http error response to the `stacks-node`, then the `stacks-node` would try to re-emit the same event, and this would just repeat indefinitely. This results in the db `event_observer_requests` table growing indefinitely, and a configured tsv output file growing indefinitely. 

With this PR, when a desync occurs, the program exits immediately with a specific error message and exit code.

### DIFF
--- a/.github/workflows/stacks-blockchain-api.yml
+++ b/.github/workflows/stacks-blockchain-api.yml
@@ -438,6 +438,7 @@ jobs:
         with:
           token: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Semantic Release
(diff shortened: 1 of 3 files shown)


smcclellan:
@zone117x This seems abandoned -- any reason not to close?

Issue #944, opened 2021-12-22
Print log and terminate when event-replay fails


CharlieC3:
**Describe the bug**
If event-replay fails, the error gets logged but it doesn't report that the replay failed. This makes it difficult to automatically detect through automated systems.

**Expected behavior**
In addition to the error, a line should print out noting that the event-replay process failed due to the aforementioned error.
And if it doesn't already, a failed event-replay should return a non-zero status code as well.

PR #980, opened 2022-01-17
Print additional log for failing event-replay

MHassanTariq:
## Description

This PR adds an additional error log specifying failing event-replay to help automatically detect such failure.  

For details refer to issue #944 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/event-stream/event-server.ts
+++ b/src/event-stream/event-server.ts
@@ -817,6 +817,7 @@ export async function startEventServer(opts: {
         res.status(200).json({ result: 'ok' });
       } catch (error) {
         logError(`error processing core-node /new_block: ${error}`, error);
+        logError(`Event-replay process failed due to the aforementioned error`);
         res.status(500).json({ error: error });
       }
     })
@@ -831,6 +832,7 @@ export async function startEventServer(opts: {
         res.status(200).json({ result: 'ok' });
       } catch (error) {
         logError(`error processing core-node /new_burn_block: ${error}`, error);
+        logError(`Event-replay process failed due to the aforementioned error`);
         res.status(500).json({ error: error });
       }
     })
@@ -845,6 +847,7 @@ export async function startEventServer(opts: {
         res.status(200).json({ result: 'ok' });
       } catch (error) {
         logError(`error processing core-node /new_mempool_tx: ${error}`, error);
+        logError(`Event-replay process failed due to the aforementioned error`);
         res.status(500).json({ error: error });
       }
     })
@@ -859,6 +862,7 @@ export async function startEventServer(opts: {
         res.status(200).json({ result: 'ok' });
       } catch (error) {
         logError(`error processing core-node /drop_mempool_tx: ${error}`, error);
+        logError(`Event-replay process failed due to the aforementioned error`);
         res.status(500).json({ error: error });
       }
     })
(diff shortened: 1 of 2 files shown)


src/event-stream/event-server.ts
```diff
@@ -817,6 +817,7 @@ export async function startEventServer(opts: {
         res.status(200).json({ result: 'ok' });
       } catch (error) {
         logError(`error processing core-node /new_block: ${error}`, error);
+        logError(`Event-replay process failed due to the aforementioned error`);
```
rafaelcr, line 820: These errors could happen also if there wasn't an event replay in progress, so I don't think these messages should be printed here.

IIRC, there's an `eventReplay` property in the `index.ts` file that shows whether or not we're actually performing one, so I think this logic should be handled over there.

MHassanTariq: Thanks for the help on this @rafaelcr 

I went through the documentation for [Event-Replay](https://github.com/hirosystems/stacks-blockchain-api#event-replay) and noticed that event-replay initiates on `export-events` or `import-events` commands are given. Therefore, I've logged the error in the catch. Can you check it once and let me know if that looks good to you? 

cc: @CharlieC3

src/index.ts
```diff
@@ -381,5 +381,6 @@ async function handleProgramArgs() {
 
 void handleProgramArgs().catch(error => {
   console.error(error);
+  console.error(`Event-replay process failed due to the aforementioned error`);
```
rafaelcr, line 384: You should only show this error when the command arg was `import-events`

rafaelcr (changes requested):

src/index.ts
rafaelcr: Alternatively, you can concat the operand to this string so we have an error that would work for all operations e.g.:
```js
`${args.operand} process failed`
```

src/index.ts
```diff
@@ -274,7 +274,7 @@ function initApp() {
     });
 }
 
-async function handleProgramArgs() {
+function getEventReplayArgs() {
```
rafaelcr, line 277: Change name to `getProgramArgs`.
Remember this function parses the args not only for event replay but for event export, etc.

rafaelcr (approved):
🏁

src/index.ts
```diff
@@ -381,5 +385,7 @@ async function handleProgramArgs() {
 
 void handleProgramArgs().catch(error => {
   console.error(error);
+  const { args } = getProgramArgs();
+  console.error(`${args.operand} process failed`);
```
zone117x, line 389: This causes the message `undefined process failed` to be printed when the API exits from an error when started regularly e.g. `node ./lib/index.js`

rafaelcr (changes requested):
My bad, missed @zone117x 's observation. Can you fix it please @MHassanTariq ?

src/index.ts
MHassanTariq: In that case, it won't go for events and [initiate the application](https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/index.ts#L331) instead. That has its own [error handler](https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/index.ts#L219-L228) which exits the program. 
Please let me know if I'm wrong here.

zone117x: Possibly, there's async and sync code mixed in here. Better to just check for null/undefined and avoid logging

zone117x (approved):
LGTM!

Issue #910, opened 2021-12-15
Improve `.git-info` handling


zone117x:
The API process will not start if the `.git-info` file is missing. Under normal circumstances, the file is auto-generated when `npm run build` is ran. 

However,  it looks relatively easy to run into an issue if the source it checked out in a way that doesn't include the repo git data, or can fail in a weird way if `npm run build` is ran on a machine without `git` installed.

The latter can be simulated by changing the npm script to something like:
```
"git-info": "echo \"$(gitnothere rev-parse --abbrev-ref HEAD)\n$(gitnothere log -1 --pretty=format:%h)\n$(gitnothere describe --tags --abbrev=0)\" > ./.git-info",
```

Might be worth looking into changing this a bit, like having our release CI commit the latest `.git-info`, or swapping the npm `git-info` script from a shell command to a small js snippet that can do more robust checks like "is git installed/available", "is repo git info available", etc.

_Originally posted by @zone117x in https://github.com/hirosystems/stacks-blockchain-api/pull/909#pullrequestreview-832999010_

PR #986, opened 2022-01-18
Replace shell script with javascript 

asimm241:
## Description
Added a javascript to check if git is installed locally or not. If yes, create a .git-info file, if no, throw an error. 

Fixes: #910 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Checklist
- [] Code is commented where needed
- [] `npm run test` passes
- [ ] Changelog is updated


git-info-script.mjs
```diff
@@ -0,0 +1,13 @@
+import { exec } from 'child_process';
+
+exec('git --version', (error, _stdout, stderr) => {
+  if (error) {
+    throw new Error(`"git is missing", please install git and retry\n${error.message}`);
+  }
+  if (stderr) {
+    throw new Error(`"git is missing", please install git and retry\n${stderr}`);
+  }
+  const gitInfo = `echo \"$(git rev-parse --abbrev-ref HEAD)\n$(git log -1 --pretty=format:%h)\n$(git describe --tags --abbrev=0)\" > ./.git-info`;
```
zone117x, line 10: If this is using *nix shell exec anyway, there's no point it moving it to a js file. Have you tested if this works on Windows?

git-info-script.mjs
```diff
@@ -0,0 +1,10 @@
+import  shell from 'shelljs';
```
rafaelcr, line 1: Change the name of this file to `git-info-generator.mjs` so intention is clearer

rafaelcr (changes requested):

git-info-generator.mjs
```diff
@@ -0,0 +1,10 @@
+import  shell from 'shelljs';
+
+if (!shell.which('git')) {
+  throw new Error(`"git is missing", please install git and retry`);
+}
+shell
+  .exec(
+    `echo \"$(git rev-parse --abbrev-ref HEAD)\n$(git log -1 --pretty=format:%h)\n$(git describe --tags --abbrev=0)\"`
```
zone117x, line 8: Any reason to be running shell code like this as opposed to using [`child_process.spawnSync(...)`](https://nodejs.org/docs/latest-v16.x/api/child_process.html#child_processspawnsynccommand-args-options)?

asimm241: It is done this way to support it on all platform including windows. Otherwise OS there would be different code for windows like [spawning-bat-and-cmd-files-on-windows](https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows)

zone117x:
```ts
const gitInfo = [
  'git rev-parse --abbrev-ref HEAD',
  'git log -1 --pretty=format:%h',
  'git describe --tags --abbrev=0',
].map(r => execSync(r, { encoding: 'utf8' }));
fs.writeFileSync('.git-info', gitInfo.join('\n'));
```
I think something like this would work just fine

asimm241: @zone117x What if the user has git installed but does not have tags fetch locally. should we throw an error in that case or ignore it?

zone117x: Good question, I think if git repo data is missing from disk then this should throw. @rafaelcr thoughts?

rafaelcr: Yes, that's a very good point @asimm241. IIRC, though, git doesn't return status 1 if the tags are missing from the repo so you'll probably have to check the output message directly.

git-info-generator.mjs
```diff
@@ -0,0 +1,22 @@
+import { execSync } from 'child_process';
+import { writeFileSync } from 'fs';
+import shell from 'shelljs';
+
+if (!shell.which('git')) {
```
zone117x, line 5: No need to pull in the shelljs dependency for this. Something like this should work:
```js
try {
  execSync('git --version');
} catch (error) {
  console.error(error.message);
  throw new Error(`git is missing, please install git and retry`);
}
```

tsconfig.json
```diff
@@ -18,6 +18,6 @@
       "@stacks/blockchain-api-client": ["client/src"]
     }
   },
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "git-info-generator.mjs"],
```
zone117x, line 21: What's the reason for this change?

git-info-generator.mjs
asimm241: do you think It would work on windows?

tsconfig.json
asimm241: oh, that's slipped in, I don't think we need that anymore.

.github/workflows/stacks-blockchain-api.yml
```diff
@@ -75,6 +75,9 @@ jobs:
             ${{ runner.os }}-build-${{ env.cache-name }}-
             ${{ runner.os }}-build-
             ${{ runner.os }}-
+        
+      - name: Fetch tags
+        run: git fetch --prune --unshallow --tags
```
rafaelcr, line 80: Instead of this step, add this to the `actions/checkout@v2` step:
```yml
with:
  fetch-depth: 0
```

rafaelcr (changes requested):

git-info-generator.mjs
```diff
@@ -0,0 +1,24 @@
+import { execSync } from 'child_process';
+import { writeFileSync } from 'fs';
+
+try {
+  execSync('git --version');
```
zone117x, line 5: Wrapping this and the rest of calls in the same try-catch can show a misleading error message. 

If git repo data isn't available it will show "git not installed' which is incorrect.

zone117x (approved):
LGTM! We should probably hold off merging this until we cut the next release.

rafaelcr (approved):

PR #988, opened 2022-01-18
fix: optimize indexes for `blocks`, `microblocks`, `txs`, `mempool_txs`

rafaelcr:
* Makes sure sorting indexes are correct, specially on `block_height`
* Removes all unused boolean indexes
* Adds indexes that should be there (block index, parent index, burn index, etc.)

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -3722,8 +3722,8 @@ export class PgDataStore
         `
         SELECT ${TX_COLUMNS}, ${abiColumn()}
         FROM txs
-        WHERE tx_id = $1 AND index_block_hash = $2 AND canonical = true AND microblock_canonical = true
-        ORDER BY block_height DESC
+        WHERE tx_id = $1 AND index_block_hash = $2
+        ORDER BY canonical DESC, microblock_canonical DESC, block_height DESC
         LIMIT 1
         `,
         [hexToBuffer(args.txId), hexToBuffer(args.indexBlockHash)]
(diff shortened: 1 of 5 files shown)


src/datastore/postgres-store.ts
```diff
@@ -2925,7 +2925,7 @@ export class PgDataStore
         SELECT ${TX_COLUMNS}, ${abiColumn()}
         FROM txs
         WHERE index_block_hash = $1 AND canonical = true AND microblock_canonical = true
-        ORDER BY microblock_sequence ASC, tx_index ASC
+        ORDER BY microblock_sequence DESC, tx_index DESC
```
zone117x, line 2928: This function is only used by a Rosetta endpoint right? Are you sure this has the intended effect?

rafaelcr: Yeah, I pushed this because thought this would optimize the query but I didn't realize ASC is the necessary order here. I'll revert this change.

src/migrations/1584619633448_txs.ts
```diff
@@ -154,29 +154,30 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
...
+  pgm.createIndex('txs', 'parent_index_block_hash');
+  pgm.createIndex('txs', 'sender_address');
+  pgm.createIndex('txs', 'smart_contract_contract_id');
   pgm.createIndex('txs', 'sponsor_address');
-  pgm.createIndex('txs', 'sponsored');
+  pgm.createIndex('txs', 'status');
   pgm.createIndex('txs', 'token_transfer_recipient_address');
-  pgm.createIndex('txs', 'contract_call_contract_id');
-  pgm.createIndex('txs', 'smart_contract_contract_id');
-  pgm.createIndex('txs', ['canonical', 'microblock_canonical']);
```
zone117x, line 171: Are you confident this is unnecessary? We have a lot of `WHERE canonical=true AND microblock_canonical=true` queries -- those never use this index?

rafaelcr: Yep, that's right. Since they're boolean properties, the filtering is done at the last step of the query plan (without using the index).

I tested this thoroughly with different queries and I never managed to get postgres to use this index.

rafaelcr: I'm other words, it's basically an index that has 4 possible values.

In a huge table it doesn't make a big difference.

zone117x (approved):
LGTM!

Issue #990, opened 2022-01-18
Add support for postgres read replicas


CharlieC3:
**Is your feature request related to a problem? Please describe.**
I'd like support added for postgres read replicas. This is different than the API's read-only mode, where an API only accepts HTTP traffic and does not write to the Postgres database. The API in read-only mode will still attempt to perform a SQL LISTEN command on the backend Postgres database at bootup. If read replicas are being used, LISTEN will fail on them and cause the API to crash.

**Describe the solution you'd like**
Support for read replicas should ensure any SQL commands which require a db primary, be directed to the primary. While other compatible commands (pretty much everything else when running a read-only API) is directed to a different URL which includes the read-replicas.

This can be achieved by storing the primary and read-replica URLs in two different ENV variables, and choosing the appropriate URL depending on the action being performed.

See [this page](https://www.postgresql.org/docs/12/hot-standby.html) for certain SQL actions which will cause errors in a read replica.

zone117x:
@CharlieC3 any idea of the data replication latency? One potential issue with hooking LISTENs to the primary is that the follow up queries against the replica might fail if the data is not yet available.

aulneau:
this is great news, was just going to come make this issue 🙏

PR #993, opened 2022-01-20
feat: use primary pg server for notifier

rafaelcr:
Adds primary pg server .env vars for `PgNotifier` to use.
Related to #990

### DIFF
--- a/.env
+++ b/.env
@@ -5,14 +5,28 @@ PG_PASSWORD=postgres
 PG_DATABASE=stacks_blockchain_api
 PG_SCHEMA=public
 PG_SSL=false
+# The connection URI below can be used in place of the PG variables above,
+# but if enabled it must be defined without others or omitted.
+# PG_CONNECTION_URI=
+
+# If your PG deployment implements a combination of primary server and read replicas, you should
+# specify the values below to point to the primary server. The API will use primary when
+# implementing LISTEN/NOTIFY postgres messages for websocket/socket.io support.
+# Any value not provided here will fall back to the default equivalent above.
+# PG_PRIMARY_HOST=
+# PG_PRIMARY_PORT=
+# PG_PRIMARY_USER=
+# PG_PRIMARY_PASSWORD=
+# PG_PRIMARY_DATABASE=
+# PG_PRIMARY_SCHEMA=
+# PG_PRIMARY_SSL=
+# The connection URI below can be used in place of the PG variables above,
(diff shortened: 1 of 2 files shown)


zone117x (approved):
This looks good to me. I think we'd need to add more docs around usage if/when other parties would want to use this feature, since the pg replica setup needs to be setup in particular away to ensure they are synced for the sql queries that follow a NOTIFY.

But we can add that later if needed.

rafaelcr:
Thanks, I agree @zone117x that's a good point... maybe I could expand the .env comments a bit to mention this if @CharlieC3 has a good basic suggestion for replica setup

CharlieC3:
@rafaelcr There's a few required settings in both the primary and read-replica's configs that will allow for synchronous replication. Below are some trimmed down snippets for both a primary and its single replica. If a db cluster has more than 1 replica, it would only have to change slightly.

Primary config snippet:
```
#------------------------------------------------------------------------------
# CONNECTIONS AND AUTHENTICATION
#------------------------------------------------------------------------------

listen_addresses = '*'
port = '5432'

#------------------------------------------------------------------------------
# WRITE-AHEAD LOG
#------------------------------------------------------------------------------

wal_level = 'replica'
fsync = 'on'
synchronous_commit = 'on'

#------------------------------------------------------------------------------
# REPLICATION
#------------------------------------------------------------------------------

synchronous_standby_names = '1 ("stacks_node_api")'
hot_standby = 'on'
```

Replica config snippet:
```
#------------------------------------------------------------------------------
# CONNECTIONS AND AUTHENTICATION
#------------------------------------------------------------------------------

listen_addresses = '*'
port = '5432'

#------------------------------------------------------------------------------
# WRITE-AHEAD LOG
#------------------------------------------------------------------------------

wal_level = 'replica'
fsync = 'on'

#------------------------------------------------------------------------------
# REPLICATION
#------------------------------------------------------------------------------

primary_conninfo = 'host=<PRIMARY HOST NAME> port=5432 user=<MY REPLICATION USER> password=<MY_PASSWORD> application_name=stacks_node_api'
promote_trigger_file = '/tmp/postgresql.trigger.5432'
hot_standby = 'on'
```

These configs don't include _everything_, like authentication settings or tunings, we should refer devs to the postgres docs and [PGTune](https://pgtune.leopard.in.ua/) for that.

aulneau:
I've got this running and seems to be working well, thanks so much @rafaelcr :~)

PR #996, opened 2022-01-24
feat: select default/primary pg server for all queries

rafaelcr:
Maintains two postgres pools (default and primary) and selects the appropriate one depending on each query's purpose: `INSERT` and `UPDATE` operations go to primary, `SELECT` goes to default.

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -681,22 +681,30 @@ function getSqlQueryString(query: QueryConfig | string): string {
   }
 }
 
+enum PgServer {
+  default,
+  primary,
+}
+
 export class PgDataStore
   extends (EventEmitter as { new (): DataStoreEventEmitter })
   implements DataStore {
+  readonly primaryPool: Pool;
   readonly pool: Pool;
   readonly notifier?: PgNotifier;
-  readonly eventReplay: boolean;
+  readonly isEventReplay: boolean;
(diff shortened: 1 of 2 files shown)


zone117x (commented):
I like this approach as it involves a minimal amount of changes, but I'm somewhat worried about the maintainability. Doing a thorough review of this PR involves looking through essentially every query function and determining if it involves a write operation, same thing in the future with each new change to the pg datastore class. We would almost want a new suite of unit tests that ran all queries with a read-only permissioned pg connection to ensure correct behavior, although that seems like a lot of work.

I'm wondering if it makes sense to split the datastore interface in two: one with functions that write, the other with functions that read. It would be a pretty large change, think something like that is worth it?

src/datastore/postgres-store.ts
```diff
@@ -685,18 +692,22 @@ export class PgDataStore
   extends (EventEmitter as { new (): DataStoreEventEmitter })
   implements DataStore {
   readonly pool: Pool;
+  readonly primaryPool: Pool;
```
zone117x, line 695: Should there be new code added to the `close()` function to destroy this pool instance?

rafaelcr:
Closing this PR. I'm working on a new one that refactors `postgres-store.ts` into separate default/primary files and starts using https://github.com/porsager/postgres for pg connections and queries.

PR #1012, opened 2022-01-31
ci: update CI workflow and remove slack notifs

CharlieC3:
## Description
* Changed the name of the main workflow to just "CI". It was originally given the same name as the repository when GH Actions was much newer, and was a requirement for something hacky we were doing back then. It's not necessary anymore.
* Removed the Slack notify jobs from the CI workflow.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

No

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: stacks-blockchain-api
+name: CI
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


rafaelcr:
@CharlieC3 we recently reversed the change that made the CI run only once instead of twice per PR push because we thought it conflicted with semantic release (AFAIK it didn't), could we bring that back in this PR too?
```yml
on:
  push:
    branches:
      - '**'
```

CharlieC3:
@rafaelcr I recall our conversation on that, but it looks like it already is running twice:
https://github.com/hirosystems/stacks-blockchain-api/pull/1012/checks

Is that what you intended? Or do you mean to you want it to only run once per commit when a PR is open?

rafaelcr:
@CharlieC3 I'd like it to run only once per PR commit, if possible. If not, that's ok.

CharlieC3:
@rafaelcr Misread your first message, done!

rafaelcr (approved):
Thanks @CharlieC3

Issue #1003, opened 2022-01-25
Testnet faucet endpoint should respect multi-cast config


CharlieC3:
**Describe the bug**
Currently, the faucet endpoint does use the multi-cast feature of the API. So faucet requests won't get multi-casted to overridden targets in the multi-cast configuration.

**To Reproduce**
1. Send a faucet request to a testnet API with multicast configured with multiple targets
2. Observe the faucet request only gets routed to the default backend proxy stacks-node.

**Expected behavior**
Testnet faucet requests should get routed to all targets configured in the mult-cast config file.

API version: 1.0.7

zone117x:
We looks like we haven't ran into problems with this in a couple years (maybe due to improved mempool propagation in the stacks-node?). Going to close for now, please feel free to re-open.

PR #1014, opened 2022-02-01
feat: tx multi-cast support for testnet stx faucet

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1003

TODO: draft, pending unit tests

### DIFF
--- a/src/core-rpc/client.ts
+++ b/src/core-rpc/client.ts
@@ -285,19 +285,14 @@ export class StacksCoreRpcClient {
         errors: [],
       };
     }
+    // Use Set to remove dupes
     const endpoints = [
-      // The primary proxy endpoint (the http response from this one will be returned to the client)
-      this.createUrl('v2/transactions'),
-      ...extraEndpoints.map(e => this.createUrl('v2/transactions', { endpoint: e })),
+      ...new Set([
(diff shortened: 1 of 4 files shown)


src/core-rpc/client.ts
```diff
@@ -208,6 +268,60 @@ export class StacksCoreRpcClient {
...
+      ...extraEndpoints.map(e => this.createUrl('v2/transactions', { endpoint: e })),
+    ];
+    // Remove dupes
+    for (let i = endpoints.length - 1; i >= 0; i--) {
+      for (let f = endpoints.length - 1; f >= 0; f--) {
+        if (f !== i && endpoints[i] === endpoints[f]) {
+          endpoints.splice(f, 1);
+        }
+      }
+    }
```
rafaelcr, line 300: To avoid this O(n^2) loop, you could feed the endpoints to a `Map<string, bool>` (taking the keys afterwards) or a `Set` from the start.

I'm also not sure on what JS would do if you modify the `endpoints` array while you're iterating it. Some other languages I've come across freak out with an OOB exception when this happens.

zone117x: Map would be better except for not preserving order (the first endpont is treated special). But yeah I'll look for a better way to do this, maybe a Set somehow.

Issue #883, opened 2021-12-08
Endpoint to get all events of certain type


aulneau:
Similar to https://github.com/hirosystems/stacks-blockchain-api/issues/875 (maybe could be added as a param to that endpoint), being able to fetch a paginated list of all events (not specific to a tx_id) of a given event_type would be very helpful.

rafaelcr:
Thanks @aulneau, could you give us a bit more context on what use case you need this endpoint for? We'd like to figure out the best way to solve this and/or what additional data we'd need to return on this new endpoint

aulneau:
@rafaelcr sure thing! I would like to be able to fetch data for all "stx_lock" events, or data for all "smart_contract_log" where x = y. Ideally I'd also be able to search for all events where principal Z is a participant. eg "all smart_contract_log events where principal z was a participant in the transaction"

There are really limitless use cases for this kind of querying. Lmk if i can provide more information.

PR #1027, opened 2022-02-02
Events-only endpoints for address and tx_id

### LINKED ISSUE #875: Expose events-only tx endpoint
Hi there!

Currently when i'm working with transaction data, (eg fetching all transactions for a given principal), none of the events data is included in the transaction object. As such I need to do a secondary fetch for each transaction to the `/extended/v1/tx` endpoint to grab only the events.

I was thinking it'd be really helpful to have an endpoint at something like `/extended/v1/tx-events` that could take either a single `tx_id`, or an array of `tx_ids`.

Curious what you all think, or if there is a better solution!

asimm241:
## Description
Add endpoints to return events for principal or transaction-id 

closes #875 #883 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

_NO_

## Testing information

API unit tests are added

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated
- [ ] Tag 1 of @rafaelcr


docs/openapi.yaml
```diff
@@ -3169,3 +3169,87 @@ paths:
               $ref: ./api/info/get-fee-rate-request.schema.json
             example:
               $ref: ./api/info/get-fee-rate-request.example.json
+
+  /extended/v1/events/address/{principal}:
```
rafaelcr, line 3173: These two endpoints should be mixed into one general-purpose endpoint that allows you to filter for events by principal, tx_id, type, etc:
```
/extended/v1/tx/events
```

docs/openapi.yaml
```diff
@@ -3169,3 +3169,87 @@ paths:
...
+    get:
+      summary: Get address events
+      description: Get list of events belong to a particular address
+      tags:
+        - Accounts
```
rafaelcr, line 3178: Move to `Transactions`

docs/openapi.yaml
```diff
@@ -3169,3 +3169,87 @@ paths:
...
+      operationId: get_address_events
+      parameters:
+        - name: principal
+          in: path
+          description: Stacks address or a Contract identifier (e.g. `SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0.get-info`)
+          required: true
```
rafaelcr, line 3184: Add a `tx_id` parameter. Users should specify at least one of these two.

docs/openapi.yaml
```diff
@@ -3169,3 +3169,87 @@ paths:
...
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: ./api/address/get-address-events.schema.json
+              example:
+                $ref: ./api/address/get-address-events.example.json
+
+  /extended/v1/events/tx/{tx_id}:
```
rafaelcr, line 3215: Remove this endpoint

src/api/routes/events.ts
```diff
@@ -0,0 +1,94 @@
...
+      DbEventTypeId.StxAsset,
+      DbEventTypeId.FungibleTokenAsset,
+      DbEventTypeId.NonFungibleTokenAsset,
+      DbEventTypeId.StxLock,
+    ]; //no filter provided , return all types of events
+  }
+  return eventTypeFilter;
+}
+
+export function createEventsRouter(db: DataStore): express.Router {
```
rafaelcr, line 35: Move these functions to the Transactions router

src/datastore/postgres-store.ts
```diff
@@ -4270,6 +4330,341 @@ export class PgDataStore
...
+      for (const eventType of args.eventTypeFilter) {
+        switch (eventType) {
+          case DbEventTypeId.StxLock:
+            const stxLockResults = await client.query<StxLockEventQueryResult>(
+              `
+              SELECT
+                event_index, tx_id, tx_index, block_height, canonical, locked_amount, unlock_height, locked_address
+              FROM stx_lock_events
+              WHERE locked_address = $1 AND microblock_canonical = true AND event_index BETWEEN $2 AND $3
+              `,
```
rafaelcr, line 4352: A few comments on these queries:
1. Querying for events in separate `query()` calls is a bad idea for pagination because you want to paginate the results, not filter against the `event_index` property of every table. You should switch to a single query in which you make `UNION` of all the results and then paginate over those **sorted** results (it will be very important to sort by `block_height, microblock_sequence, tx_index, event_index`).
2. Make sure to filter both `canonical` AND `microblock_canonical`

rafaelcr (changes requested):

src/datastore/postgres-store.ts
asimm241: I followed the same approach as this https://github.com/hirosystems/stacks-blockchain-api/blob/f1abc291d955525983d40da11f8f5c638cee756c/src/datastore/postgres-store.ts#L4077

should we look into this as well?

rafaelcr: Interesting @asimm241.

Where is that function being used? We could evaluate whether or not to refactor that one (later, in another PR) based on the current use.

asimm241: @rafaelcr I don't think we can do union here since all events have different columns.

src/api/init.ts
```diff
@@ -7,6 +7,10 @@ import { v4 as uuid } from 'uuid';
 import * as cors from 'cors';
 import * as WebSocket from 'ws';
 import * as SocketIO from 'socket.io';
+import { ChainID } from '@stacks/transactions';
+import * as pathToRegex from 'path-to-regexp';
+import * as expressListEndpoints from 'express-list-endpoints';
+import { createMiddleware as createPrometheusMiddleware } from '@promster/express';
```
zone117x, line 13: Can you revert unnecessary whitespace / stylistic changes like this?

src/api/query-helpers.ts
```diff
@@ -222,3 +223,55 @@ export function validatePrincipal(stxAddress: string) {
...
   }
 }
+
+export function parseAddressOrTxId(
+  req: Request,
+  res: Response,
+  next: NextFunction
+): { address: string; txId: undefined } | { address: undefined; txId: string } | never {
+  const address = req.query.address;
+  const txId = req.query.txId;
```
zone117x, line 233: This should be `req.query.tx_id` rather than `req.query.txId` right?

src/api/query-helpers.ts
```diff
@@ -222,3 +223,55 @@ export function validatePrincipal(stxAddress: string) {
...
+      eventTypeFilter = [
+        DbEventTypeId.SmartContractLog,
+        DbEventTypeId.StxAsset,
+        DbEventTypeId.FungibleTokenAsset,
+        DbEventTypeId.NonFungibleTokenAsset,
+        DbEventTypeId.StxLock,
+      ]; //no filter provided , return all types of events
+    }
+  } catch (error) {
+    handleBadRequest(res, next, `invalid 'event type'`);
```
zone117x, line 273: The try/catch around this whole block is meant to handle the error thrown by `parseEventTypeStrings()` right? If so, it should be scoped to that function call such that it doesn't overlap with the other `handleBadRequest` instead of the try body.

src/api/routes/tx.ts
```diff
@@ -250,6 +254,36 @@ export function createTxRouter(db: DataStore): express.Router {
     })
   );
 
+  router.get(
+    '/events',
+    asyncHandler(async (req, res, next) => {
+      const limit = parseTxQueryEventsLimit(req.query['limit'] ?? 96);
+      const offset = parsePagingQueryInput(req.query['offset'] ?? 0);
+
+      const { address, txId } = parseAddressOrTxId(req, res, next);
```
zone117x, line 263: Seems like `address` and `txId` input should be validated here for 400 (bad request) before handing off to datastore layer.

docs/openapi.yaml
```diff
@@ -3169,3 +3169,51 @@ paths:
...
               $ref: ./api/info/get-fee-rate-request.example.json
+
+  /extended/v1/tx/events:
+    get:
+      summary: Get transaction or address events
+      description: Get list of events belong to a particular transaction or an address
+      tags:
+        - Transactions
+      operationId: get_filtered_events
+      parameters:
```
rafaelcr, line 3180: Add an example to each of these parameters

rafaelcr (changes requested):

docs/api/transaction/get-transaction-events.example.json
```diff
@@ -0,0 +1,67 @@
+{
+  "limit": 96,
+  "offset": 0,
+  "events": [
```
rafaelcr, line 4: Please change this example to be a real response instead of using mock data ("0x1234", etc.)

docs/generated.d.ts
```diff
@@ -762,6 +763,12 @@ export type NonFungibleTokenHolding = NonFungibleTokenHoldingWithTxId | NonFungi
  * Describes the minting of a Non-Fungible Token
  */
 export type NonFungibleTokenMint = NonFungibleTokenMintWithTxId | NonFungibleTokenMintWithTxMetadata;
+export type TransactionEvent1 =
```
rafaelcr, line 766: Can you remove the a 1 at the end here? I think you'll need to edit the TransactionEvent entity schema somewhere

docs/api/transaction/get-transaction-events.example.json
rafaelcr: Is this a real response? If so, why are the `event_index` entries out of order for the same tx?

docs/openapi.yaml
```diff
@@ -3169,3 +3169,60 @@ paths:
...
+        - name: tx_id
+          in: query
+          description: Hash of transaction
+          required: false
+          schema:
+            type: string
+            example: 29e25515652dad41ef675bd0670964e3d537b80ec19cf6ca6f1dd65d5bc642c5
```
rafaelcr, line 3188: Add the `0x` at the start for consistency

src/datastore/common.ts
```diff
@@ -680,6 +680,13 @@ export interface DataStore extends DataStoreEventEmitter {
     offset: number;
   }): Promise<{ results: DbEvent[] }>;
 
+  getTransactionEvents(args: {
```
rafaelcr, line 683: Can you add a JSdoc comment to this function?
```js
/**
 * Example...
 * @param ...
 */
```

rafaelcr (changes requested):
It's looking very good, just a few style comments

src/datastore/postgres-store.ts
```diff
@@ -4310,6 +4373,156 @@ export class PgDataStore
...
   }
 
+  /**
+   * It retrieves filtered events from the db based on transaction, principal or event type. Note: It does not accept both principal and txId at the same time
+   * @param args - addressOrTxId: filter for either transaction id or address
+   * @param args - eventTypeFilter: filter based on event types ids
+   * @param args - limit: returned that many rows
+   * @param args - offset: skip that any rows
+   * @returns returns array of events
+   */
```
rafaelcr, line 4383: Looks good, but move it to `common.ts` so the comment appears in tooltips everywhere

src/api/routes/tx.ts
```diff
@@ -250,6 +254,26 @@ export function createTxRouter(db: DataStore): express.Router {
...
+      const eventTypeFilter = parseEventTypeFilter(req, res, next);
+
+      const { results } = await db.getTransactionEvents({
+        addressOrTxId: principalOrTxId,
+        eventTypeFilter,
+        offset,
+        limit,
+      });
+      const response = { limit, offset, events: results.map(e => parseDbEvent(e)) };
+      res.status(200).json(response);
```
rafaelcr, line 273: Can you add the chain tip cache headers to this response? You can look at the `/block` endpoint for an example on how to do that

src/tests/api-tests.ts
```diff
@@ -9812,6 +9812,574 @@ describe('api tests', () => {
     expect(mempoolTxResult2.body).toEqual(expectedMempoolResult2);
   });
 
+  test('Address events', async () => {
```
rafaelcr, line 9815: Set the test name to "/tx/events" so we can find it easily

src/tests/api-tests.ts
```diff
@@ -9812,6 +9812,574 @@ describe('api tests', () => {
...
+          },
+        },
+      ],
+    };
+
+    expect(addressEvents.status).toBe(200);
+    expect(addressEvents.body).toEqual(expectedResponse);
+  });
+
+  test('Address events: empty events returned', async () => {
```
rafaelcr, line 9898: "/tx/events empty events returned" and so on

rafaelcr (changes requested):
Almost there

rafaelcr (approved):
LGTM 🏁 
Can you hold off on merging until we've shipped this next release though?

asimm241:
@rafaelcr bumping it up! this PR is rebased and ready to merge.

rafaelcr (changes requested):
One small style nit before merging

rafaelcr (approved):
Great @asimm241 thank you

rafaelcr:
Can you fix the new conflicts so we can merge, please? @asimm241

PR #1028, opened 2022-02-02
feat: add `chain_tip` materialized view to track chain tip stats

rafaelcr:
* Calculate block height, leading block hashes, leading microblock hash, and transaction counts in a materialized view
* Substitute all `block_height` calculations for a SELECT against the new view
* Substitute all `COUNT(*)` queries for their respective stats in the view


zone117x (approved):
Awesome, looking forward to seeing these `count()` optimizations in prod!

PR #1034, opened 2022-02-03
chore: configure dependabot to only open PRs for security issues

zone117x:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#open-pull-requests-limit
> By default, Dependabot opens a maximum of five pull requests for version updates. Once there are five open pull requests, new requests are blocked until you merge or close some of the open requests, after which new pull requests can be opened on subsequent updates. Use open-pull-requests-limit to change this limit. This also provides a simple way to temporarily disable version updates for a package manager.
> 
> This option has no impact on security updates, which have a separate, internal limit of ten open pull requests.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,9 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    # Disable version updates for npm dependencies
+    # https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#open-pull-requests-limit
+    open-pull-requests-limit: 0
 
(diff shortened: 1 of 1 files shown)


rafaelcr (approved):

CharlieC3 (approved):
Not sure if this was your intention, but even without this change it sounds like Dependabot is still able to open security PRs, and won't be blocked by the 5 PR dependency limit. This would really only be helpful if you find the dependency updates not useful.

zone117x:
> This would really only be helpful if you find the dependency updates not useful.

We got a barrage of them last few days, none of the non-security related ones were helpful lol

PR #1037, opened 2022-02-03
chore: add additional readme badges

zone117x:
Add readme badges for:
* Latest github release version w/ link
* Docker image downloads w/ link (using the legacy `blockstack` user because it has more downloads, but linking to the updated `hirosystems` image)
* NPM client library to increase discoverability

Updated a couple references from `blockstack/stacks-blockchain-api-standalone` to `hirosystems/stacks-blockchain-api-standalone`

image


rafaelcr (approved):
Very cool 👍

PR #1042, opened 2022-02-07
feat: change string and hex column indices to Hash method

### LINKED ISSUE #1041: remove unused postgres indices
PgHero reports the following indices are unused:
```
blocks_burn_block_hash_index
blocks_burn_block_height_index
blocks_index_block_hash_index
blocks_parent_block_hash_index
blocks_parent_index_block_hash_index
burnchain_rewards_canonical_index
contract_logs_block_height_index
contract_logs_canonical_microblock_canonical_index
contract_logs_contract_identifier_index
contract_logs_microblock_canonical_index
contract_logs_microblock_sequence_index
contract_logs_parent_index_block_hash_index
faucet_requests_ip_index
ft_events_asset_identifier_index
ft_events_microblock_canonical_index
ft_events_microblock_sequence_index
ft_events_parent_index_block_hash_index
ft_metadata_name_index
ft_metadata_symbol_index
ft_metadata_tx_id_index
mempool_txs_status_index
microblocks_index_block_hash_index
microblocks_microblock_hash_index
microblocks_microblock_sequence_index
miner_rewards_block_hash_index
miner_rewards_canonical_index
miner_rewards_from_index_block_hash_index
names_canonical_index
names_microblock_canonical_index
names_parent_index_block_hash_index
names_tx_index_index
namespaces_canonical_index
namespaces_microblock_canonical_index
nft_events_microblock_sequence_index
nft_events_parent_index_block_hash_index
nft_metadata_contract_id_index
nft_metadata_name_index
nft_metadata_tx_id_index
reward_slot_holders_address_index
reward_slot_holders_canonical_index
smart_contracts_abi_index
smart_contracts_block_height_index
smart_contracts_microblock_canonical_index
smart_contracts_microblock_sequence_index
smart_contracts_parent_index_block_hash_index
smart_contracts_tx_id_index
stx_events_canonical_microblock_canonical_index
stx_events_microblock_canonical_index
stx_events_microblock_sequence_index
stx_events_parent_index_block_hash_index
stx_lock_events_canonical_index
stx_lock_events_microblock_canonical_index
subdomains_canonical_microblock_canonical_index
subdomains_microblock_canonical_index
subdomains_parent_index_block_hash_index
subdomains_resolver_index
txs_microblock_sequence_tx_index_index
txs_parent_index_block_hash_index
txs_status_index
```
We should remove them to speed up database writes

rafaelcr:
Change method of all string/hex column indices to [Hash](https://www.postgresql.org/docs/14/indexes-types.html#INDEXES-TYPES-HASH) instead of the default B-Tree to optimize `WHERE` statements that do direct comparisons using `=`.

Some indices that were previously removed (like `tx_id` on the `txs` table) were brought back using Hash because that allowed the query planner to prefer them instead of table constraints which always use B-Tree indexing.

Also:
* Make sure all `block_height` indices are sorted `DESC`
* Remove unused indices
* Rename migration files to fit common standard
* Remove unnecessary `down` instructions that are automatically inferred by `pgm`
* Split BNS migrations into one file per table
* Turn off `PgNotifier` on event import/export connections
* Handle cases where an input TSV file contains duplicate node messages
* Warn when a duplicate microblock was trying to be inserted into the database

Fixes #1041


src/migrations/1584604583726_blocks.ts
```diff
@@ -75,11 +75,6 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
   });
 
   pgm.createIndex('blocks', 'block_hash');
-  pgm.createIndex('blocks', 'burn_block_hash');
-  pgm.createIndex('blocks', 'burn_block_height');
-  pgm.createIndex('blocks', 'index_block_hash');
```
zone117x, line 80: Are the unused indexes being reported by an instance that watches both read and write queries? For example, `index_block_hash` is queried in a few places during ingestion and re-org handling. It's often along-side other columns like
```sql
        FROM blocks
        WHERE block_height = $1 AND index_block_hash = $2
```
So maybe the query planner just isn't using the index for this column?

zone117x: Similarly, does the tool ever suggest missing _composite_ indexes? Like in this case, it could potentially be a net gain if there was a `(block_height, index_block_hash)` index?

rafaelcr: > Are the unused indexes being reported by an instance that watches both read and write queries?
Yes, PgHero shows a report for primary and read-replicas

> Similarly, does the tool ever suggest missing composite indexes?
Not AFAIK... I haven't found a section that suggests new indices, but we could definitely try that index if you think so

zone117x (commented):
Should we do perf tests on this change before merging to develop? Otherwise looks good!

rafaelcr:
@zone117x yes, agreed 👍🏻 
I'll open an issue on devops so we can hopefully test this next week

CharlieC3:
I'm getting the following error when running an event replay to a DB.
```
error: duplicate key value violates unique constraint "temp_event_observer_requests_pkey"
    at Parser.parseErrorMessage (/app/node_modules/pg-protocol/src/parser.ts:369:69)
    at Parser.handlePacket (/app/node_modules/pg-protocol/src/parser.ts:188:21)
    at Parser.parse (/app/node_modules/pg-protocol/src/parser.ts:103:30)
    at Socket.<anonymous> (/app/node_modules/pg-protocol/src/index.ts:7:48)
    at Socket.emit (node:events:520:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
  length: 305,
  severity: 'ERROR',
  code: '23505',
  detail: 'Key (id)=(718095) already exists.',
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: 'COPY temp_event_observer_requests, line 718095',
  schema: 'pg_temp_5',
  table: 'temp_event_observer_requests',
  column: undefined,
  dataType: undefined,
  constraint: 'temp_event_observer_requests_pkey',
  file: 'nbtinsert.c',
  line: '670',
  routine: '_bt_check_unique'
}
```

I'm not positive it's related to the changes from this PR, I believe I saw a similar error come up a couple weeks ago that I dismissed thinking it was something I did wrong.

This is the command I'm using to initiate the event replay:
`node ./lib/index.js import-events --file ${STACKS_EXPORT_EVENTS_FILE} --wipe-db --force`

rafaelcr:
@CharlieC3 I just tried an event replay with the TSV I downloaded today and I saw the same error for both `develop` and this PR's branch, so it's not related to the new indices.

Furthermore, I compared to another replay I just did from an old TSV (december 2021) and that one did get imported correctly, suggesting that there is something wrong going on with the TSV. My guess is that an event is included twice (might be more but it errors out when it sees the first one).

A solution might be to add an `ON CONFLICT` clause to the event import that simply ignores repeated events. What do you think?

CharlieC3:
@rafaelcr Doesn't the import code already have logic in place to ignore duplicate entries? I recall seeing log entries for certain events in the TSV being ignored over the course of an event-replay -- I thought it was for duplicate entries but I could be wrong.

Is there a chance that the duplicate event's contents are different than the original one in the database? Meaning, their primary keys are the same and thus create the conflict, but the other columns between the two entries are different for some legitimate reason, caused by a re-org or something.

If there's no chance of this happening (or is very unlikely/should not be happening), then I think what you suggested is something we should be doing as it would improve the API's reliability at no downside. Otherwise the TSV may be unrecoverable if events were written to it incorrectly by some other bug.

rafaelcr:
@CharlieC3 
> Doesn't the import code already have logic in place to ignore duplicate entries?

It does but it's around duplicate mempool txs, blocks, etc., we don't currently check for duplicate TSV entries

> Is there a chance that the duplicate event's contents are different than the original one in the database?

I don't think that could be the case but I'll let @zone117x chime in here

> If there's no chance of this happening (or is very unlikely/should not be happening), then I think what you suggested is something we should be doing as it would improve the API's reliability at no downside. 

Yep, I actually just tested this approach and works perfectly to ignore duplicate TSV entries, I'll update this PR soon

rafaelcr:
@CharlieC3 @zone117x update... the replay did manage to crash at 98% complete with this error
```
{"code":"23505","constraint":"unique_microblock_hash","detail":"Key (microblock_hash)=(\\x6e7e3efd5075dfa4f90d4ec1510612060c64f3594d2a8a16b66bcfacf4520d23) already exists.","file":"nbtinsert.c","length":288,"level":"error","line":"670","message":"Error processing core node microblock message duplicate key value violates unique constraint \"unique_microblock_hash\"","name":"error","routine":"_bt_check_unique","schema":"public","severity":"ERROR","stack":"error: duplicate key value violates unique constraint \"unique_microblock_hash\"\n    at Parser.parseErrorMessage (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/parser.ts:369:69)\n    at Parser.handlePacket (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/parser.ts:188:21)\n    at Parser.parse (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/parser.ts:103:30)\n    at Socket.<anonymous> (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/index.ts:7:48)\n    at Socket.emit (events.js:400:28)\n    at addChunk (internal/streams/readable.js:293:12)\n    at readableAddChunk (internal/streams/readable.js:267:9)\n    at Socket.Readable.push (internal/streams/readable.js:206:10)\n    at TCP.onStreamRead (internal/stream_base_commons.js:188:23)","table":"microblocks","timestamp":"2022-02-15T22:45:09.023Z"}
{"code":"23505","constraint":"unique_microblock_hash","detail":"Key (microblock_hash)=(\\x6e7e3efd5075dfa4f90d4ec1510612060c64f3594d2a8a16b66bcfacf4520d23) already exists.","file":"nbtinsert.c","length":288,"level":"error","line":"670","message":"error processing core-node /new_microblocks: error: duplicate key value violates unique constraint \"unique_microblock_hash\" duplicate key value violates unique constraint \"unique_microblock_hash\"","name":"error","routine":"_bt_check_unique","schema":"public","severity":"ERROR","stack":"error: duplicate key value violates unique constraint \"unique_microblock_hash\"\n    at Parser.parseErrorMessage (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/parser.ts:369:69)\n    at Parser.handlePacket (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/parser.ts:188:21)\n    at Parser.parse (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/parser.ts:103:30)\n    at Socket.<anonymous> (/Users/rafaelcr/src/stacks/stacks-blockchain-api/node_modules/pg-protocol/src/index.ts:7:48)\n    at Socket.emit (events.js:400:28)\n    at addChunk (internal/streams/readable.js:293:12)\n    at readableAddChunk (internal/streams/readable.js:267:9)\n    at Socket.Readable.push (internal/streams/readable.js:206:10)\n    at TCP.onStreamRead (internal/stream_base_commons.js:188:23)","table":"microblocks","timestamp":"2022-02-15T22:45:09.023Z"}
{"level":"error","message":"HTTP POST /new_microblocks","req":{"headers":{"connection":"close","content-length":"124957","content-type":"application/json","host":"127.0.0.1:64881"},"httpVersion":"1.1","method":"POST","originalUrl":"/new_microblocks","query":{},"url":"/new_microblocks"},"res":{"statusCode":500},"responseTime":11,"timestamp":"2022-02-15T22:45:09.023Z"}
Error: Bad status response status code 500: Internal Server Error
    at IncomingMessage.<anonymous> (/Users/rafaelcr/src/stacks/stacks-blockchain-api/src/helpers.ts:342:29)
    at IncomingMessage.emit (events.js:412:35)
    at endReadableNT (internal/streams/readable.js:1334:12)
    at processTicksAndRejections (internal/process/task_queues.js:82:21) {
  requestUrl: 'http://127.0.0.1:64881/new_microblocks',
  statusCode: 500,
  response: '{"error":{"length":288,"name":"error","severity":"ERROR","code":"23505","detail":"Key (microblock_hash)=(\\\\x6e7e3efd5075dfa4f90d4ec1510612060c64f3594d2a8a16b66bcfacf4520d23) already exists.","schema":"public","table":"microblocks","constraint":"unique_microblock_hash","file":"nbtinsert.c","line":"670","routine":"_bt_check_unique"}}'
}
import-events process failed
```
Should I attempt to recover gracefully from this error? Or does this mean something else could be going on?

zone117x:
The `/new_microblocks` event is unreliable. It's already is known to send out-of-order microblocks, as well as produce microblock streams with unrecoverable gaps / missing microblocks. In both these cases, the API detects then drops/ignores them, resulting in those microblocks not showing up until the next `/new_block` event. That's often why we see microblock related flapping when traffic is split between 2 different environments. 

The above case seems like another variant that hasn't yet happened before, where the same microblock is emitted twice. It seems safe to detect and ignore this condition as well.

rafaelcr:
@CharlieC3 I finally have a successful replay, but I think the last 2 percent of the log is worth a look. For example, one particular `/new_microblocks` message was received hundreds of times. See [replay-microblocks.txt](https://github.com/hirosystems/stacks-blockchain-api/files/8083165/replay-microblocks.txt)

@zone117x can you help me with another quick review?

Thanks

CharlieC3:
@rafaelcr Does the API write an event to the TSV file even if it's rejected from the database? Wondering if the stacks-node fed it duplicate data over a bunch over erroneous restarts.

src/datastore/postgres-store.ts
```diff
@@ -941,26 +945,40 @@ export class PgDataStore
...
-        const importStream = client.query(
-          pgCopyStreams.from(`COPY temp_event_observer_requests FROM STDIN`)
-        );
+        const importStream = client.query(pgCopyStreams.from(`COPY temp_raw_tsv FROM STDIN`));
         await pipelineAsync(readStream, importStream);
+        await client.query(`
+          INSERT INTO temp_event_observer_requests
+          SELECT *
+          FROM temp_raw_tsv
+          ON CONFLICT DO NOTHING;
```
zone117x, line 980: I'm looking at how the tsv file could result in duplicate `id`s, since those should just be a mirror of the unique primary key in the `event_observer_requests` table:
https://github.com/hirosystems/stacks-blockchain-api/blob/f4649fd91ce559a14da849757685a7098fcbbf27/src/datastore/postgres-store.ts#L876-L909

It looks like the tsv file can be appended to with the _expected_ `id` before `INSERT INTO event_observer_requests ...` transaction is fully committed, like if the API or pg shutdown after the file write, but before the sql commit finishes. There's probably room to improve the logic there to reduce or eliminate the chance of that happening. But, with the existing behavior, it looks like the _newer_ tsv entries should actually overwrite the older ones on conflict. 

Thoughts?

zone117x:
@CharlieC3 
> Does the API write an event to the TSV file even if it's rejected from the database? Wondering if the stacks-node fed it duplicate data over a bunch over erroneous restarts.

Yes, it's the first thing the API does on event receipt so that it's behaving as close to a "store and forward" layer as possible.

src/datastore/postgres-store.ts
```diff
@@ -881,42 +881,42 @@ export class PgDataStore
...
+    if (insertResult.rowCount !== 1) {
+      throw new Error(
+        `Unexpected row count ${insertResult.rowCount} when storing event_observer_requests entry`
+      );
+    }
+    const exportEventsFile = process.env['STACKS_EXPORT_EVENTS_FILE'];
+    if (exportEventsFile) {
+      const result = insertResult.rows[0];
+      const tsvRow = [result.id, result.receive_timestamp, result.event_path, result.jsonb];
+      fs.appendFileSync(exportEventsFile, tsvRow.join('\t'));
```
rafaelcr, line 910: What do you think about this approach @zone117x ? Basically, I'm trying to finalize the query before writing synchronously into the TSV file.

My only worry here is that there could be a race condition for example if we get a barrage of events (like mempool txs), all the queries settle correctly, but then the `appendFileSync` calls fall out of order and the TSV file ends up with unordered rows. That is a possibility with the current code too, I believe, but it would only happen if we're running several processes per write-only pod.

zone117x: Cool this looks good to me, seems like best way to reduce odds of race conditions. Ultimately, it will still be possible for the process to crash or be forced killed between a sql commit and file write, but no way to really get around that.

zone117x:
@rafaelcr should we merge this and combine testing efforts for the v3.0.0 release?

zone117x (approved):

PR #1043, opened 2022-02-08
feat: add usage details for each pg connection via application_name

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1046

Use the pg connection `application_name` string property to specify usage context. This was previously left undefined which makes it difficult to determine which API instances are using connections or what purpose a give connection is serving.

This is a free-form value, the PR uses the convention:
`<app instance name>:<optional datastore instance scope>;<connection instance scope>`

The app instance name is read from the env var `PG_APPLICATION_NAME` (or `PG_PRIMARY_APPLICATION_NAME` when applicable), which could be set to something like the pod name.

The usage scopes are specified in code, e.g. `event-replay`, `notifier`, `init-connection-poll`, the run-mode, etc.

Examples:
`api-pod-blue-3:import-events;datastore-crud`
`api-pod-blue-3:datastore-readonly;init-connection-poll`
`api-pod-blue-3:datastore-readonly;datastore-crud`
`api-pod-blue-3:datastore-readonly;notifier`

### DIFF
--- a/.env
+++ b/.env
@@ -5,6 +5,10 @@ PG_PASSWORD=postgres
 PG_DATABASE=stacks_blockchain_api
 PG_SCHEMA=public
 PG_SSL=false
+
+# Can be any string, use to specify a use case specific to a deployment
+PG_APPLICATION_NAME=stacks_api_app
+
 # The connection URI below can be used in place of the PG variables above,
 # but if enabled it must be defined without others or omitted.
 # PG_CONNECTION_URI=
--- a/.env
+++ b/.env
@@ -7,7 +7,7 @@ PG_SCHEMA=public
 PG_SSL=false
 
 # Can be any string, use to specify a use case specific to a deployment
-PG_APPLICATION_NAME=stacks_api_app
+PG_APPLICATION_NAME=stacks-blockchain-api
 
 # The connection URI below can be used in place of the PG variables above,
 # but if enabled it must be defined without others or omitted.
--- a/src/tests/datastore-tests.ts
+++ b/src/tests/datastore-tests.ts
@@ -68,11 +68,11 @@ describe('in-memory datastore', () => {
   });
 });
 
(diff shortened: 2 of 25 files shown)


src/tests/datastore-tests.ts
```diff
@@ -702,7 +729,7 @@ describe('postgres datastore', () => {
     expect([...addrDResult]).toEqual([]);
   });
 
-  test.only('pg block store and retrieve', async () => {
+  test('pg block store and retrieve', async () => {
```
zone117x, line 732: I think tests in this PR are broken because of this. This `test.only` slipped through a few months ago and since then some of these tests seem to have been broken.

zone117x: Yep, that is the issue https://github.com/hirosystems/stacks-blockchain-api/issues/1045

CharlieC3 (commented):
In the PR description you mention the optional use of  the`PG_PRIMARY_APPLICATION_NAME` env var, but I don't see this env var in code. Is this being dynamically generated somewhere?

Overall looks good and shouldn't be difficult to specify unique pod names!

.env
```diff
@@ -5,6 +5,10 @@ PG_PASSWORD=postgres
 PG_DATABASE=stacks_blockchain_api
 PG_SCHEMA=public
 PG_SSL=false
+
+# Can be any string, use to specify a use case specific to a deployment
+PG_APPLICATION_NAME=stacks_api_app
```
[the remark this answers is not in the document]
zone117x: Good catch!

> In the PR description you mention the optional use of thePG_PRIMARY_APPLICATION_NAME env var, but I don't see this env var in code. Is this being dynamically generated somewhere?

Yep, the bit of code reading the pg env vars is dynamic already (it checks both the `PRIMARY` and non-primary).

rafaelcr (approved):
LGTM, very nice

PR #1044, opened 2022-02-08
chore: re-enable & repair datastore tests

zone117x:
Around 5 months ago [a `test.only` slipped](https://github.com/hirosystems/stacks-blockchain-api/pull/759/files#diff-77355b7b19cf4bc905ea87dc6722e76b275c022ddb3bc77155ff48337cca5cd2) into the `datastore-tests.ts` file which disabled several critical tests. 

Since then, several of these tests have broken in subsequent PRs.

This PR re-enables the tests, and fixes the broken ones (WIP).


src/datastore/postgres-store.ts
```diff
@@ -5459,18 +5464,22 @@ export class PgDataStore
         WITH
         -- getAddressTxs
         stx_txs AS (
-          SELECT tx_id, ${COUNT_COLUMN}
+          SELECT tx_id
           FROM principal_stx_txs AS s
           WHERE principal = $1 AND ${blockCond}
+        ` +
+          // TODO: this also needs to sort by microblock_sequence DESC, tx_index DESC, but the
+          // columns don't currently exist on the table. this will be a breaking change to fix
```
zone117x, line 5472: Opened an issue for this TODO https://github.com/hirosystems/stacks-blockchain-api/issues/1050

src/datastore/postgres-store.ts
```diff
@@ -492,7 +492,7 @@ interface MempoolTxQueryResult {
   coinbase_payload?: Buffer;
 
   // sending abi in case tx is contract call
-  abi?: string;
+  abi: unknown | null;
```
zone117x, line 495: Note: using `unknown` to represent the JSON object here for lack of a better type. Another option is `Record<string, unknown>`, which typescript/linters often recommend (e.g. when trying to use the `Object` type).

We could also use `ClarityAbi` from stacks.js, although I'm somewhat wary of depending on an external lib's type in this area.

@rafaelcr I don't have strong feelings one way or the other, let me know what you think.

rafaelcr: I was curious about this yesterday too, actually, but I arrived at the same conclusion that `unknown` might be ok for now... I think it's safe, we can re-evaluate later if there's a better alternative

rafaelcr (approved):
LGTM

Issue #1007, opened 2022-01-26
Add strict FT/NFT metadata flag


rafaelcr:
We currently have flags that enable FT/NFT metadata processing in `.env`, but this does not guarantee the metadata is available when required.
We should add another flag that toggles a strict mode (`STACKS_API_METADATA_STRICT_MODE`) that, when enabled, it will try again forever if the metadata retrieval error is recoverable. If it's not, then try a given number of times before failing.

Relevant codepaths:
https://github.com/hirosystems/stacks-blockchain-api/blob/f29919b5fb4b4275a07f4be7aec1ef2f66a6ec52/src/event-stream/tokens-contract-handler.ts#L629
https://github.com/hirosystems/stacks-blockchain-api/blob/f29919b5fb4b4275a07f4be7aec1ef2f66a6ec52/src/event-stream/tokens-contract-handler.ts#L645

PR #1049, opened 2022-02-08
fix: add token metadata modes to warn on missing rosetta FT metadata

rafaelcr:
Adds a `STACKS_API_TOKEN_METADATA_MODE` .env value so users can control how the API should handle missing token metadata (warn or error). In the future, #1007 would add a `strict` mode to this value.

TODO: Add unit tests

### DIFF
--- a/src/event-stream/tokens-contract-handler.ts
+++ b/src/event-stream/tokens-contract-handler.ts
@@ -48,6 +48,17 @@ const METADATA_MAX_PAYLOAD_BYTE_SIZE = 1_000_000; // 1 megabyte
 
 const PUBLIC_IPFS = 'https://ipfs.io';
 
+export enum TokenMetadataMode {
+  /**
+   * Default mode. If a required token metadata is not found, the API will issue a warning.
+   */
+  warning,
+  /**
+   * If a required token metadata is not found, the API will throw an error.
+   */
+  error,
+}
+
 export function isFtMetadataEnabled() {
   const opt = process.env['STACKS_API_ENABLE_FT_METADATA']?.toLowerCase().trim();
   return opt === '1' || opt === 'true';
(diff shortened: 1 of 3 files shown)


src/event-stream/tokens-contract-handler.ts
```diff
@@ -58,6 +69,19 @@ export function isNftMetadataEnabled() {
...
 
+/**
+ * Determines the token metadata mode based on .env values.
+ * @returns TokenMetadataMode
+ */
+export function tokenMetadataMode(): TokenMetadataMode {
+  switch (process.env['STACKS_API_TOKEN_METADATA_MODE']) {
+    case 'warning':
+      return TokenMetadataMode.warning;
+    default:
```
zone117x, line 80: Default should be `warning` right?

rafaelcr: Whoops, nice catch!

src/rosetta-helpers.ts
```diff
@@ -1012,7 +1016,11 @@ export async function getValidatedFtMetadata(
   const tokenContractId = assetIdentifier.split('::')[0];
   const ftMetadata = await db.getFtMetadata(tokenContractId);
   if (!ftMetadata.found) {
-    throw new Error(`FT metadata not found for token: ${assetIdentifier}`);
+    if (tokenMetadataMode() === TokenMetadataMode.warning) {
+      logger.warn(`FT metadata not found for token: ${assetIdentifier}`);
+    } else {
+      throw new Error(`FT metadata not found for token: ${assetIdentifier}`);
```
zone117x, line 1022: This is fine for the hotfix, but I think in a future PR we should check if a given FT's ABI matches the SIP-09 ABI before throwing. 

It actually might need to go further and check 1) does the token match SIP-09 ABI, and 2) is the metadata missing because of a recoverable error e.g. network timeouts vs a Clarity contract error. Otherwise a single SIP-09 compliant looking FT could still have some contract error and completely block the strict mode from working.

rafaelcr: Good point. I'll add a TODO comment to that effect

zone117x (approved):
LGTM! Should we open an issue for the TODO?

rafaelcr:
Thanks, I'll address the TODO while solving #1007

PR #1056, opened 2022-02-11
fix: principal_stx_txs sorting

### LINKED ISSUE #1050: Add additional columns for sorting `principal_stx_txs` table
A hotfix in PR https://github.com/hirosystems/stacks-blockchain-api/pull/1044/files had to move the LIMIT/OFFSET clauses to the outer sql query to fix an issue with out-of-order txs returned by the `getAddressTxs(...)` function:
https://github.com/hirosystems/stacks-blockchain-api/blob/c483485c9c9b632d9c1360dd6d347d9d2b83e57f/src/datastore/postgres-store.ts#L5469-L5487

This degrades the performance of endpoints using this query. The `principal_stx_txs` table needs a breaking change to add the additional columns so the more performant query can be restored.

rafaelcr:
* Add `microblock_sequence` and `tx_index` columns to `principal_stx_txs`
* Sort via these columns when querying for `/transactions`

Closes #1050

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -4414,25 +4414,34 @@ export class PgDataStore
     }
     const insertPrincipalStxTxs = async (principals: string[]) => {
       principals = [...new Set(principals)]; // Remove duplicates
-      const columnCount = 3;
+      const columnCount = 5;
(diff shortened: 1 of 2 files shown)


zone117x (approved):
Fix looks good 👍

PR #1067, opened 2022-02-15
ci: update gh-pages workflow

CharlieC3:
## Description

Updates the name of the referenced CI workflow to kick off the gh-pages workflow.
This should fix the problem where the gh-pages aren't updated after a release.


rafaelcr (approved):
LGTM

Issue #975, opened 2022-01-13
Pin exact package.json dependencies


zone117x:
We're currently using `^x.y.z` versioning (any version in the same major range) in `package.json`. We've ran into a few issues in the past with deps not following semver completely. We're also prone to breakage from other js-ecosystem idiosyncrasies, e.g. catching a human-readable string in an `Error` objects due to common lack of proper error handling in many libraries. 

We should pin all dependencies to exact versions, e.g. `x.y.z` rather than `^x.y.z`. 

The repo already has bots setup to report dependency vulnerability / security related issues, so no need to worry there.

MHassanTariq:
Please let me know if there is anything remaining regarding this issue. Otherwise, I think we can close it. Thanks!
@zone117x @rafaelcr

PR #1068, opened 2022-02-16
Pin exact dependencies

MHassanTariq:
## Description

This PR updates the dependencies and uses exact ones. Required dependencies have been updated for the proper working of the project. 

For details refer to issue #975 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


rafaelcr (approved):
LGTM

zone117x (changes requested):
The package-lock.json file is showing this change downgrading many dependencies to older versions. If a dependency is included with `^1.0.0`, then typically its highest semver compatible version will be used, e.g. `1.5.0`.

So simply removing the `^` char from package.json is going to result in version downgrades.

MHassanTariq:
I'll do that for all libraries. I did it only for the ones that would otherwise break the project. 
Thanks for pointing it out!

zone117x (approved):
LGTM, thanks @MHassanTariq!

PR #1069, opened 2022-02-18
Add the missing "routes" path to the overview guide for completeness

sabbyanandan:
## Description
As a user, while reviewing the architecture overview, I noticed the missing reference to the path in which the Express.js route may have been residing, so I wanted to make it more explicit as to where they are living in this doc-change.

Please disregard and close this PR if I misinterpreted the intent. Perhaps the correct directory reference would still be helpful for the general audience.

## Type of Change


zone117x (approved):
LGTM, thanks @sabbyanandan!

PR #1086, opened 2022-02-28
fix: deactivate bns table indices before subdomain import

### LINKED ISSUE #1081: Optimize subdomain import
The current subdomain import code is working very slowly in the newer local db deployments, taking close to 6 hours to complete. There are ~2,200,000 subdomains to be imported for every event replay.

One theory as to why this is slower now is that the table indices for the `namespaces`, `subdomains` and `names` tables now use Hash as their method to speed up lookups, but that is more expensive to produce for every `INSERT` we do during import.

A possible solution would be to find a way to delay the index recalculation on those tables until all the items are imported, either by destroying and then re-creating the indices on finish, or by using temporary tables that keep the same structure without the indices to later just copying the data from there to the correct tables.

rafaelcr:
This patch temporarily disables indices on the `subdomains`, `zonefiles`, `namespaces`, and `names` tables before subdomain import, and also removes unused indices from these tables.

From postgres docs on [populating a database](https://www.postgresql.org/docs/current/populate.html#POPULATE-RM-INDEXES):

> If you are adding large amounts of data to an existing table, it might be a win to drop the indexes, load the table, and then recreate the indexes.

Instead of dropping and re-creating, this code just disables and reindexes them after import is complete via the use of the `indisready` and `indisvalid` [index state flags](https://www.postgresql.org/docs/current/catalog-pg-index.html).

On local tests, it now takes ~2 seconds to import 10,000 subdomains when before this change it took ~30 seconds per 10,000.

Fixes #1081

### DIFF
--- a/src/import-v1/index.ts
+++ b/src/import-v1/index.ts
@@ -430,6 +430,15 @@ export async function importV1BnsData(db: PgDataStore, importDir: string) {
   const client = await db.pool.connect();
   try {
     await client.query('BEGIN');
+    // Temporarily disable BNS table indices to speed up INSERTs.
+    await client.query(`
+      UPDATE pg_index
+      SET indisready = false, indisvalid = false
+      WHERE indrelid = ANY (
+        SELECT oid FROM pg_class
+        WHERE relname IN ('subdomains', 'zonefiles', 'namespaces', 'names')
+      )
+    `);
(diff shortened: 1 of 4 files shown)


zone117x (approved):
Neat postgres feature!

PR #1089, opened 2022-03-02
fix: send address tx updates correctly on microblocks

### LINKED ISSUE #1085: address-transaction update is not emitted for microblocks
Hi, 
I had opened a feature request https://github.com/hirosystems/stacks-blockchain-api/issues/952 to receive updates while listening to address-transactions of a contract and a contract call tx makes it in a microblock, I should receive updates.

this was confirmed delivered in https://github.com/hirosystems/stacks-blockchain-api/pull/956/commits/c1b98d0ed720b1260e7a70ac761b7472fb03a20b

But this doesn't seem to be the case, neither socket.io-client or websocket client receive these events. 

I do see the code that's supposed to send these updates to notifier
https://github.com/hirosystems/stacks-blockchain-api/blob/9cae1ca01664db2e9ae6955d1a9a21accbe08159/src/datastore/postgres-store.ts#L1153

So I'm not sure why I'm not able to receive these event notifications with below setup:
>client.subscribeAddressTransactions(contract, event => {})

rafaelcr:
`block_height` was being sent incorrectly to the ws/socket.io handlers when a transaction was confirmed in a microblock, causing the call to `getAddressTxsWithAssetTransfers` to return empty results.

This patch also fixes some race conditions and flaky tests that happened from socket messages not being awaited correctly.

Fixes #1085

### DIFF
--- a/src/api/routes/ws/socket-io.ts
+++ b/src/api/routes/ws/socket-io.ts
@@ -44,22 +44,22 @@ export function createSocketIORouter(db: DataStore, server: http.Server) {
     if (subscriptions) {
       // TODO: check if init topics are valid, reject connection with error if not
       const topics = [...[subscriptions]].flat().flatMap(r => r.split(','));
-      topics.forEach(topic => {
+      topics.forEach(async topic => {
         prometheus?.subscribe(socket, topic);
-        void socket.join(topic);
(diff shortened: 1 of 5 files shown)


zone117x (approved):
Nice catch!

PR #1090, opened 2022-03-02
chore: add log entries warning of slow bns reindex during import

rafaelcr:


CharlieC3 (approved):
:shipit: Thanks for adding this!

PR #1091, opened 2022-03-03
fix bns download script

hozzjss:
## Description
update the script to download bns data to absolute path if user is in different path than `/`


rafaelcr:
This change looks good @hozzjss thank you 👍 
However, could you change the target branch to `develop` so we can merge it + bundle it with the next release?

rafaelcr (approved):

Issue #1000, opened 2022-01-25
expand `/rosetta/v1/construction/preprocess` examples to include contract calls


rafaelcr:
all of the current examples are about STX transfers, so we should add a contract call example and probably a token transfer example.

MHassanTariq:
@rafaelcr I couldn't find any example for STX transfers in [/rosetta/v1/construction/preprocess](https://github.com/hirosystems/stacks-blockchain-api/blob/develop/docs/openapi.yaml#L2191-L2216). Can you clarify which examples are you talking about?

PR #1093, opened 2022-03-07
Example and test for contract call in `/rosetta/v1/construction/preprocess`

MHassanTariq:
## Description

This PR adds a test, and example in the documentation for `construction/preprocess` for `stack_stx`.
Test was added to increase code coverage for Rosetta tests.

For details refer to issue #1000 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/docs/api/rosetta/rosetta-construction-preprocess-response.example.json
+++ b/docs/api/rosetta/rosetta-construction-preprocess-response.example.json
@@ -1,0 +1,16 @@
+{
+  "options": {
+    "sender_address": "STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6",
+    "type": "stack_stx",
+    "number_of_cycles": 5,
+    "amount": "5000",
+    "symbol": "STX",
+    "decimals": 6,
+    "pox_addr": "1Xik14zRm29UsyS6DjhYg4iZeZqsDa8D3",
+    "suggested_fee_multiplier": 1,
+    "max_fee": "12380898",
+    "size": 260,
+    "memo": "SAMPLE MEMO"
+  },
+  "required_public_keys": [{ "address": "STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6" }]
+}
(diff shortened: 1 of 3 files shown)


docs/api/rosetta/rosetta-construction-preprocess-response.example.json
```diff
@@ -0,0 +1,16 @@
...
+    "amount": "5000",
+    "symbol": "STX",
+    "decimals": 6,
+    "pox_addr": "1Xik14zRm29UsyS6DjhYg4iZeZqsDa8D3",
+    "suggested_fee_multiplier": 1,
+    "max_fee": "12380898",
+    "size": 260,
+    "memo": "SAMPLE MEMO"
+  },
+  "required_public_keys": [{ "address": "STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6" }]
```
rafaelcr, line 15: Can you add the correct line breaks to this array/object?

rafaelcr:
@MHassanTariq this looks good, however, it would be best to add another example of a regular contract call (like miamicoin operation, NFT mint, etc.) so users can see how to call arbitrary smart contracts. Can you add that?
Let me know if you need help with setting different examples for 1 request in openapi (you can look at the `/nft/history` endpoint for a concrete example)

asimm241:
> @MHassanTariq this looks good, however, it would be best to add another example of a regular contract call (like miamicoin operation, NFT mint, etc.) so users can see how to call arbitrary smart contracts. Can you add that? Let me know if you need help with setting different examples for 1 request in openapi (you can look at the `/nft/history` endpoint for a concrete example)

@rafaelcr I don't think rosetta supports arbitrary contract calls right now. 
We only support two **hardcoded** contract calls to `pox` contracts. 
 1. `stack-stx` 
 2.  `delegate-stx` 
 Where do you think we can provide an arbitrary contract?

rafaelcr (approved):
LGTM

rafaelcr (approved):

Issue #873, opened 2021-12-03
Optimize Clarity value serialization


zone117x:
The latest CPU profiling against production points to over 50% of CPU time spent encoding/decoding raw Clarity values. For example, generating `repr` strings for contract-call tx args, parsing "length-prefixed" strings from raw tx segments, etc. 

(Technically some of this actually parsing binary formats in the STX wire protocol / tx encoding format, but I'm lumping them together because they likely require the same kind of optimization approach).

![image](https://user-images.githubusercontent.com/1447546/144636150-9c36d280-502c-4e08-985d-21877ab2f847.png)

Parsing lots of small binary objects in js is slow. An approach to optimizing this is to _not_ do this in js. Much faster parser implementations already exist in rust, which we can use as either a native nodejs addon, or use WASM compiled versions. For example, the Clarity-repl.js has WASM that could be used.

PR #1094, opened 2022-03-08
Feat/native stacks decoding

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/873

* New native (rust-based) NodeJS addon library here: https://github.com/zone117x/stacks-encoding-native-js/actions/runs/1952832451
  * Includes native functions for decoding Clarity values, binary transaction blobs, post-condition binary blobs, Stacks address blobs, and more. All of which use ASM / hardware acceleration where available.
* The stacks-blockchain-api branch of the refactor using this new library is here: https://github.com/hirosystems/stacks-blockchain-api/tree/feat/native-stacks-decoding

All tests pass locally with when the new `stacks-encoding-native-js` library is compiling locally. Draft PR while working on the CI for the npm package for the lib.


rafaelcr (commented):
Looks amazing so far 🔥 Can't wait to try it out!
Were you thinking of tackling the TODO items in this same PR too? Or should we merge this first?

src/api/serializers/post-conditions.ts
```diff
@@ -30,109 +26,77 @@ const assetPrincipalTypeMap = {
 } as const;
 
 function serializePostConditionPrincipal(
-  principal: TxPostConditionPrincipal
+  principal: import('stacks-encoding-native-js').PostConditionPrincipal
```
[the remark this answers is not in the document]
zone117x: Ah just a refactoring artifact, nice catch, will fix 👍

zone117x:
> Were you thinking of tackling the TODO items in this same PR too? Or should we merge this first?

I'm aiming to finish at least a few of the critical TODOs around testing in this PR. Especially to make sure no drop in code coverage.

zone117x:
@CharlieC3 @rafaelcr codecov is on the fritz. Some of the files shown as a drop in coverage aren't touched by this PR, others that had a boost in coverage aren't listed. Their website errors while trying to view the diff. 

I can continue adding tests _unrelated_ to this PR's changes in order to boost coverage, but not sure if that makes sense.

zone117x:
Update:
I've made progress on comprehensive benchmarking capabilities for _read_ traffic analysis. I've put together scripts and some new features that provided the ability to run a local API instance using a read-only mode, populated with a recent mainnet event-replay TSV. 

A new benchmarking and regression testing script reads from a list of GET http endpoints, sampled from real-world traffic in Kibana. It initiates a "start CPU profiling" request to the API, then iterates through the sampled endpoints using curl, going as fast as the server can response.

It also checks for non-200 responses which was useful in catching a couple edge-cases where this PR had a couple encoding issues.

Once all requests have been issued, a "stop CPU profiling" request is sent, with the results dumped to disk (for later analysis with flamegraph tools). The profiling supports both the existing NodeJS CPU profiling features, as also a new profiling function integrated with the `stacks-encoding-native-js` lib which outputs data useful for analyzing performance of the native/rust code in this library.

Initial local analysis shows this PR results in around a 75% reduction in CPU time spent on Stacks encoding/decoding (one of the largest remaining CPU hogs identified in https://github.com/hirosystems/stacks-blockchain-api/issues/873). Additionally, all the STX address encoding/decoding is performed within the native lib and thus the LRU cache introduced in https://github.com/hirosystems/stacks-blockchain-api/pull/841 is no longer necessary -- this should result in something around a 50% reduction in memory usage (need to confirm some data with @CharlieC3). 

The rust/native profiling also shows what appears to be low-hanging fruit where some optimizations in the stacks-node libraries and the native addon could be optimized further and bring the CPU time down even more. 

Next steps:
* Commit benchmarking & regression testing scripts/files to this repo (probably under the `/utils` dir), with some instructions on usage for future regression testing.
* Rust optimizations for some of easy win code paths shown by the profiler.
* Get a canary version this PR deployed to ensure local benchmarks match prod.

src/inspector-util.ts
```diff
@@ -342,8 +356,118 @@ export async function startProfilerServer(
...
+    let profilerResults: Buffer;
+    try {
+      profilerResults = stopProfiler();
+    } catch (error: any) {
+      console.error(error);
+      res.status(500).end(error);
+      return;
+    }
+    const fileName = `profile-${Date.now()}.svg`;
+    const nativeProfilerFilePath = path.join('/Users/matt/Downloads/profiler-results/', fileName);
```
rafaelcr, line 431: Can you make this a relative path?

utils/package.json
```diff
@@ -21,6 +22,7 @@
     "pg-copy-streams": "^5.1.1",
     "pg-cursor": "^2.6.0",
     "pg-listen": "^1.7.0",
+    "stacks-encoding-native-js": "file:../../stacks-encoding-native-js",
```
rafaelcr, line 25: Shouldn't this point to your repo? I know it's temporary but just in case

rafaelcr, line 25: Shouldn't this point to your repo? I know it's temporary but just in case

zone117x: This debug code has since been converted into unit tests in the stacks-encoding-native projects, so I just removed them from this `utils/` dir.

src/inspector-util.ts
zone117x: This was debug code, just removed it!

src/event-replay/reverse-file-stream.ts
```diff
@@ -34,9 +42,16 @@ export class ReverseFileStream extends stream.Readable {
...
+
       // Ignore empty/trailing lines, `readable.push('')` is not recommended
       this.lineBuffer = this.lineBuffer.filter(line => line.length > 0);
+      const remainderHasPrefixEnding = this.remainder.startsWith('\n');
       this.remainder = this.lineBuffer.shift() ?? '';
+
+      // Preserve the line-ending char for the remainder if one was at the read boundary
+      if (remainderHasPrefixEnding) {
+        this.remainder = '\n' + this.remainder;
+      }
```
zone117x, line 54: @rafaelcr found and fixed a bug here where multiple lines were merged and emitted if a read chunk starts with a `\n` character

src/datastore/common.ts
```diff
@@ -144,13 +143,17 @@ export interface BaseTx {
...
   status: DbTxStatus;
   type_id: DbTxTypeId;
   /** Only valid for `contract_call` tx types */
   contract_call_contract_id?: string;
   contract_call_function_name?: string;
+  // TODO(perf): use hex string since that is what we already get from deserializing event payloads
+  //   and from the pg-node adapter (all the pg js libs use text mode rather than binary mode)
   /** Hex encoded Clarity values. Undefined if function defines no args. */
   contract_call_function_args?: Buffer;
   abi?: string;
```
zone117x, line 159: With the cpu optimizations from this PR, profiling indicates around a 3rd of the cpu time is spent on the postgres adapter layer (converting postgres data to and from js data). Which is great news! Because it indicates the rest of the code is now using relatively little cpu. 

One of the slow paths around the pg data layer is converting between Buffers and hex strings. The nodejs pg library we use (and the newer one we are considering to use) both operated in "text" mode rather than binary mode, which means even though many columns are stored in `bytea` format, they are still converted to a hex string before being sent to the nodejs layer, and the adapter then has to parse it into a buffer. Same thing for writing buffers from nodejs to pg, the adapter encodes Buffers to a hex string before sending the queries to pg. 

In many of these cases, we already _start off_ with a hex string and convert it into a Buffer when submitting pg queries, and when we get a Buffer from the pg adapter, we re-encode it back into a hex string to send to client http requests. So it looks like we'd cut down the cpu time a decent amount (need to do more benchmarking to know how much) by just avoiding the intermediate Buffer type all-together.

src/tests/synthetic-stx-txs-tests.ts
```diff
@@ -0,0 +1,347 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import { DecodedTxResult, TxPayloadTypeID } from 'stacks-encoding-native-js';
+import { CoreNodeBlockMessage } from '../event-stream/core-node-message';
+import { parseMessageTransaction } from '../event-stream/reader';
+
+// Test processing of the psuedo-Stacks transactions, i.e. the ones that
+// originate on the Bitcoin chain, and have a `raw_tx == '0x00'.
+
+test('test synthetic tx token transfer 1', () => {
```
rafaelcr, line 11: Can you wrap these tests inside a `describe('synthetic stx txs', () =>`?
So we can filter them out easily on Jest in the future

rafaelcr (approved):
LGTM, this looks amazing, can't wait to try it out with prod traffic... I only left a style nit comment on tests

src/tests/synthetic-stx-txs-tests.ts
zone117x: Done!

rafaelcr (approved):
LGTM 💯

PR #1101, opened 2022-03-14
feat: prune garbage collected (256 blocks old) mempool txs

### LINKED ISSUE #1059: prune mempool transactions if older than 256 blocks from chain tip
**Describe the bug**
(Sleepi asked me to file this issue...)
I will cause they are still in there. 🙏 
https://stacks-node-api.mainnet.stacks.co/extended/v1/tx/mempool?offset=400
Most of them are receipt on 04Feb
"receipt_time_iso": "2022-02-04T09:57:40.000Z"

rafaelcr:
A new `receipt_block_height` column was required in the `mempool_txs` table to record at which height we saw the tx first.
Closes #1059

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -2184,7 +2184,7 @@ export class PgDataStore
     const deletedTxResults = await client.query<{ tx_id: Buffer }>(
       `UPDATE mempool_txs
       SET pruned = TRUE, status = $2
-      WHERE receipt_block_height < $1
+      WHERE pruned = FALSE AND receipt_block_height < $1
       RETURNING tx_id`,
       [cutoffBlockHeight, DbTxStatus.DroppedApiGarbageCollect]
     );


src/datastore/postgres-store.ts
```diff
@@ -2158,6 +2163,33 @@ export class PgDataStore
...
+      `SELECT (block_height - $1) AS block_height FROM chain_tip`,
+      [blockThreshold]
+    );
+    if (cutoffResults.rowCount != 1) {
+      return { deletedTxs: [] };
+    }
+    const cutoffBlockHeight = cutoffResults.rows[0].block_height;
+    // Delete every mempool tx that came before that block.
+    const deletedTxResults = await client.query<{ tx_id: Buffer }>(
+      `DELETE FROM mempool_txs
```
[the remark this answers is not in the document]
rafaelcr: Done @zone117x 👍

src/datastore/postgres-store.ts
```diff
@@ -2180,8 +2180,10 @@ export class PgDataStore
     }
     const cutoffBlockHeight = cutoffResults.rows[0].block_height;
     // Delete every mempool tx that came before that block.
+    // TODO: Use DELETE instead of UPDATE once we implement a non-archival API replay mode.
     const deletedTxResults = await client.query<{ tx_id: Buffer }>(
-      `DELETE FROM mempool_txs
+      `UPDATE mempool_txs
+      SET pruned = TRUE
```
zone117x: Do we also want to set the status like here?
https://github.com/hirosystems/stacks-blockchain-api/blob/8f56087e8ff399dd6359111fa623b5a76fcc8204/src/datastore/postgres-store.ts#L3466-L3468

We could use `DroppedStaleGarbageCollect` (the one we receive from the stacks-node), although maybe a new status code for "dropped by API because stale" would be more useful for later analysis of the table?

rafaelcr: I think Github might have sent you an old commit email, here's the one where I set it to `DroppedStaleGarbageCollect ` @zone117x https://github.com/hirosystems/stacks-blockchain-api/pull/1101/commits/ef08c0f76184b18e3218ee207165890c9b519b33 😄

rafaelcr: however, your idea of a new status is great (garbage collected by the API)... I'll add that

zone117x: Haha gotcha, looks great!

src/datastore/postgres-store.ts
```diff
@@ -2158,6 +2163,35 @@ export class PgDataStore
...
+    if (cutoffResults.rowCount != 1) {
+      return { deletedTxs: [] };
+    }
+    const cutoffBlockHeight = cutoffResults.rows[0].block_height;
+    // Delete every mempool tx that came before that block.
+    // TODO: Use DELETE instead of UPDATE once we implement a non-archival API replay mode.
+    const deletedTxResults = await client.query<{ tx_id: Buffer }>(
+      `UPDATE mempool_txs
+      SET pruned = TRUE, status = $2
+      WHERE receipt_block_height < $1
```
zone117x, line 2187: Should this `where` condition exclude rows that were already pruned? Otherwise I think it would overwrite existing `status` columns, right?

rafaelcr: Oh, great catch. You're right.

zone117x (approved):
Looks great! Because we have a change in a migration file, we should merge this with the "BREAKING CHANGE" conventional commit directive.

PR #1102, opened 2022-03-14
fix: disable notifier for tests that don't need it

rafaelcr:


zone117x (approved):
Very helpful, thanks!

Issue #1077, opened 2022-02-24
Rosetta tests are flaky


rafaelcr:
The `test-rosetta` suite is flaky at times, for example: https://github.com/hirosystems/stacks-blockchain-api/runs/5314495421?check_suite_focus=true#step:8:2456

The error output shows
```
    console.log
      Error getting pox info on try 97 Error: Response 500: Internal Server Error fetching http://127.0.0.1:20443/v2/pox - Failed to query peer info
          at StacksCoreRpcClient.fetchText (/home/runner/work/stacks-blockchain-api/stacks-blockchain-api/src/core-rpc/client.ts:133:13)
          at runMicrotasks (<anonymous>)
          at processTicksAndRejections (node:internal/process/task_queues:96:5)
          at StacksCoreRpcClient.fetchJson (/home/runner/work/stacks-blockchain-api/stacks-blockchain-api/src/core-rpc/client.ts:112:26)
          at StacksCoreRpcClient.getPox (/home/runner/work/stacks-blockchain-api/stacks-blockchain-api/src/core-rpc/client.ts:150:20)
          at standByForPoxToBeReady (/home/runner/work/stacks-blockchain-api/stacks-blockchain-api/src/tests-rosetta/api.ts:113:9)
          at Object.<anonymous> (/home/runner/work/stacks-blockchain-api/stacks-blockchain-api/src/tests-rosetta/api.ts:3076:5)

      at standByForPoxToBeReady (tests-rosetta/api.ts:117:17)
          at runMicrotasks (<anonymous>)
```

This could be due to the way those tests are crafted, by running an Event Server in parallel to the tests and thus being susceptible to timing or race conditions.

Usually, re-running all jobs from Github actions fixes the issue temporarily.

MHassanTariq:
@rafaelcr do you recommend updating db manually with data to test? That will remove the integration in tests. We can use this approach for all tests other than construction apis. We might have to consider writing those tests separately. 
An example is in my PR [here](https://github.com/hirosystems/stacks-blockchain-api/pull/1072/files#diff-cb5a6340146571d0092cd38e8d8516f087e054d3730aa7bd67636649a4f30fd8R3292-R3511).

PR #1103, opened 2022-03-14
Refactored Rosetta tests - separated construction tests

MHassanTariq:
## Description

This PR separates the rosetta tests from construction tests in `api.ts`. I have also refactored and removed integration for other tests. However, construction tests require integration with `EventServer`, therefore, those are not refactored at all. 

For details refer to issue #1077

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,10 +250,6 @@ jobs:
       - name: Run tests
         run: npm run test:rosetta
 
-      - name: Re-run tests
-        run: npm run test:rosetta
-        if: failure()
-
       - name: Print integration environment logs
         run: cat docker-compose-logs.txt
         if: failure()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,6 +250,10 @@ jobs:
       - name: Run tests
         run: npm run test:rosetta
 
+      - name: Re-run tests
+        run: npm run test:rosetta
+        if: failure()
+
       - name: Print integration environment logs
         run: cat docker-compose-logs.txt
         if: failure()
--- a/src/tests-rosetta/construction.ts
+++ b/src/tests-rosetta/construction.ts
@@ -1,0 +1,2492 @@
+import { PgDataStore, cycleMigrations, runMigrations } from '../datastore/postgres-store';
+import { PoolClient } from 'pg';
+import * as BigNum from 'bn.js';
+import { ApiServer, startApiServer } from '../api/init';
+import * as supertest from 'supertest';
+import { startEventServer } from '../event-stream/event-server';
+import { Server } from 'net';
+import { DbTx, DbTxStatus } from '../datastore/common';
+import * as assert from 'assert';
+import {
+  AnchorMode,
+  AuthType,
+  bufferCV,
+  ChainID,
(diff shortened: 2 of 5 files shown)


rafaelcr:
@MHassanTariq this looks good, but were you able to find the cause for the flakyness? If the construction tests were not refactored, do we still have a risk of flakyness in those?

MHassanTariq:
@rafaelcr I initially spent a lot of time diagnosing the actual problem. However, I wasn't able to pinpoint the exact cause as they were not breaking regularly. Therefore, I refactored the tests to the extent where I could remove the maximum integration between them.

rafaelcr:
Great news @MHassanTariq ... I think the strategy then would be to follow @CharlieC3 's advice and add a mechanism to Github actions that re-tries the test suite if it detects that same error. Can you add that to this PR?

CharlieC3:
> Great news @MHassanTariq ... I think the strategy then would be to follow @CharlieC3 's advice and add a mechanism to Github actions that re-tries the test suite if it detects that same error. Can you add that to this PR?

I think the long-term goal would be to use a custom GH runner, or figure out a way to change how the tests are ran to consume less memory. Otherwise besides the retries, we could try running **just these tests** (not the whole pipeline) on a MacOS GH runner. We get fewer of those GH runners, [but they offer double the amount of memory](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

MHassanTariq:
@rafaelcr sure. I'll look into it.
@CharlieC3 thanks for the suggestion and link!

MHassanTariq:
@CharlieC3 and @rafaelcr I've tried re-running the tests for `rosetta` once. I think it fixes the problem temporarily like @CharlieC3 mentioned. 
Can you please go through the [changes](https://github.com/hirosystems/stacks-blockchain-api/pull/1103/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR253-R255) in `ci.yml` and let me know if it looks good?

src/tests-rosetta/construction.ts
```diff
@@ -0,0 +1,2492 @@
...
+        tries++;
+        await  new StacksCoreRpcClient().getPox();
+        return Promise.resolve();
+      }
+      catch(error){
+        console.log('Error getting pox info on try ' + tries, error);
+        timeout(100);
+      }
+    }
+  }
```
rafaelcr, line 109: Echoing what we discussed on the call, this function could be improved to wait for PoX in a robust way

rafaelcr (changes requested):

.github/workflows/ci.yml
```diff
@@ -250,6 +250,10 @@ jobs:
       - name: Run tests
         run: npm run test:rosetta
 
+      - name: Re-run tests
+        run: npm run test:rosetta
+        if: failure()
```
rafaelcr, line 255: This will not be necessary anymore

rafaelcr (approved):

PR #1115, opened 2022-03-18
feat!: add mempool endpoint cache handlers

### LINKED ISSUE #879: Implement http cache-control for mempool-sensitive routes
Follow up to https://github.com/hirosystems/stacks-blockchain-api/pull/834#issuecomment-986670782

Many of the endpoints can check the mempool, which means a "state hash" of the mempool needs to be used in conjunction with the chain-tip. This can be done with a query like
```sql
SELECT digest(
(
    SELECT array_agg(tx_id)
    FROM (
        SELECT tx_id
        FROM mempool_txs
        WHERE pruned = false
        ORDER BY receipt_time DESC
    ) m
)::text, 'sha256')
```
Could also be a materialized view that is refreshed on mempool table updates. Note: this is not memory-friendly query as this concats all the mempool txids at once rather than stream/aggregate the rows, but the mempool table is small enough now to not cause a problem (yet). The latest version of pg has an aggregator function (`bit_xor`) that could be used with something like:
```sql
select bit_xor(tx_short_id)
from (
    SELECT ('x' || encode(tx_id, 'hex'))::bit(64)::bigint tx_short_id
    FROM mempool_txs
    WHERE pruned = false
    ORDER BY receipt_time DESC
) m
```
However, google cloud has not yet released support for pg14 -- should be easy enough to refactor the query once available.

rafaelcr:
This PR adds `ETag` cache support for mempool endpoints.
- Creates a new `mempool_digest` materialized view that calculates a pending mempool tx digest string only for pg14+. A version check is added to disable this string for earlier pg versions.
- Adds a new `mempool` entry to response locals to keep and validate this `ETag`.
- Installs the new cache handler in all mempool-sensitive endpoints, and adds chain tip handlers to endpoints that didn't have them before.
- Upgrade the postgres version used in our docker compose files to 14

Closes #879

### DIFF
--- a/src/api/controllers/cache-controller.ts
+++ b/src/api/controllers/cache-controller.ts
@@ -24,6 +24,10 @@ export enum ETagType {
   mempool = 'mempool',
 }
 
+const ETAG_EMPTY = -1;
+/** An ETag can be a status string or an empty value */
+type ETag = string | typeof ETAG_EMPTY;
+
 interface ETagCacheMetrics {
   chainTipCacheHits: prom.Counter<string>;
   chainTipCacheMisses: prom.Counter<string>;
@@ -78,15 +82,11 @@ export function setResponseNonCacheable(res: Response) {
  * to the response locals.
  */
 export function setETagCacheHeaders(res: Response, etagType: ETagType = ETagType.chainTip) {
-  const etag: string | undefined = res.locals[etagType];
+  const etag: ETag | undefined = res.locals[etagType];
   if (!etag) {
-    // An empty mempool ETag means either no pending mempool txs or that the API is running
-    // on old pg versions. No error reporting is necessary in that case.
-    if (etagType === ETagType.chainTip) {
-      logger.error(
-        `Cannot set cache control headers, no etag was set on \`Response.locals[${etagType}]\`.`
-      );
-    }
+    logger.error(
+      `Cannot set cache control headers, no etag was set on \`Response.locals[${etagType}]\`.`
+    );
     return;
   }
   res.set({
@@ -159,7 +159,7 @@ async function checkETagCacheOK(
   db: DataStore,
   req: Request,
   etagType: ETagType
-): Promise<string | undefined | typeof CACHE_OK> {
+): Promise<ETag | undefined | typeof CACHE_OK> {
   const metrics = getETagMetrics();
   let etag: string;
   switch (etagType) {
@@ -173,9 +173,10 @@ async function checkETagCacheOK(
       break;
     case ETagType.mempool:
       const digest = await db.getMempoolTxDigest();
-      if (!digest.found) {
-        // This would only happen if the `mempool_digest` materialized view hasn't been refreshed.
-        return;
+      if (!digest.found || !digest.result.digest) {
+        // An empty mempool digest is not a fatal error, since this can easily happen if the mempool is empty
+        // or if the `bit_xor` digest couldn't be calculated by the pg server.
+        return ETAG_EMPTY;
       }
       etag = digest.result.digest;
       break;
@@ -242,7 +243,7 @@ export function getETagCacheHandler(
     } else {
       // Request does not have a valid cache. Store the etag for later
       // use in setting response cache headers.
-      const etag: string | undefined = result;
+      const etag: ETag | undefined = result;
       res.locals[etagType] = etag;
       next();
     }
--- a/src/api/controllers/cache-controller.ts
+++ b/src/api/controllers/cache-controller.ts
@@ -1,17 +1,27 @@
 import { RequestHandler, Request, Response } from 'express';
 import * as prom from 'prom-client';
-import { FoundOrNot, logger } from '../../helpers';
-import { DataStore, DbChainTip } from '../../datastore/common';
+import { logger } from '../../helpers';
+import { DataStore } from '../../datastore/common';
 import { asyncHandler } from '../async-handler';
 
 const CACHE_OK = Symbol('cache_ok');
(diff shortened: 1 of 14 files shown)


src/migrations/1647573802137_mempool_digest.ts
```diff
@@ -0,0 +1,39 @@
...
+          SELECT ('x' || encode(tx_id, 'hex'))::bit(64)::bigint tx_short_id
+          FROM mempool_txs
+          WHERE pruned = false
+      ) m
+    `);
+  } else {
+    // TODO: Create a backwards-compatible view.
+    // We could use postgres' `digest` function here, but that requires enabling the `pgcrypto`
+    // extension which might not be possible for some users.
+    pgm.createMaterializedView('mempool_digest', {}, `SELECT NULL AS digest`);
```
rafaelcr, line 33: I'm interested in your thoughts here @zone117x ... right now, running the API on old pg versions would just make all mempool cache code moot to keep the current behavior.

I could also add a minimum recommended version of 14 to the README

src/api/controllers/cache-controller.ts
```diff
@@ -1,18 +1,29 @@
...
+ */
+export enum ETagType {
+  /** ETag based on the latest `index_block_hash` or `microblock_hash`. */
+  chainTip = 'chain_tip',
+  /** ETag based on a digest of all pending mempool `tx_id`s. */
+  mempool = 'mempool',
+}
+
 interface ChainTipCacheMetrics {
   chainTipCacheHits: prom.Counter<string>;
```
zone117x, line 28: May be useful to expand this to include counter specific to an etag/cache type (mempool vs chaintip)

src/api/controllers/cache-controller.ts
```diff
@@ -48,31 +59,30 @@ export function setResponseNonCacheable(res: Response) {
...
-      `Cannot set cache control headers, no chain tip was set on \`Response.locals[CHAIN_TIP_LOCAL]\`.`
-    );
-    return;
-  }
-  if (!chainTip.found) {
+export function setETagCacheHeaders(res: Response, etagType: ETagType = ETagType.chainTip) {
+  const etag: string | undefined = res.locals[etagType];
+  if (!etag) {
+    // An empty mempool ETag means either no pending mempool txs or that the API is running
+    // on old pg versions. No error reporting is necessary in that case.
```
zone117x, line 69: I have mild concerns with maintainability if we get rid of this error condition / logging. It's pretty easy to incorrectly setup cache control on a given endpoint, and this is one of the areas that is able to detect that. 

Seems like it would be useful to retain that behavior. What do you think?

src/api/routes/address.ts
```diff
@@ -107,10 +111,12 @@ interface AddressAssetEvents {
 
 export function createAddressRouter(db: DataStore, chainId: ChainID): express.Router {
   const router = express.Router();
-  const cacheHandler = getChainTipCacheHandler(db);
+  const cacheHandler = getETagCacheHandler(db);
+  const mempoolCacheHandler = getETagCacheHandler(db, ETagType.mempool);
 
   router.get(
     '/:stx_address/stx',
+    cacheHandler,
```
zone117x, line 119: Seems like there could be a significant amount of potential CACHE-OK responses marked as expired in requests that do _not_ include `unanchored=true`, and are only invalidated by the chaintip as opposed to the mempool.

We could definitely merge in this PR before addressing that, but something to keep in mind. We'd probably have to do some refactoring on how we pass around cache state.

src/migrations/1647573802137_mempool_digest.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { MigrationBuilder, ColumnDefinitions } from 'node-pg-migrate';
+
+export const shorthands: ColumnDefinitions | undefined = undefined;
+
+async function pgServerMajorVersion(pgm: MigrationBuilder): Promise<number | undefined> {
+  const result = await pgm.db.query(`SHOW server_version`);
```
zone117x, line 7: Just did some searching around this command.. WTF? Strange that this is essentially a free form field as opposed to a strict semver or similar.

Would it be better to check for a specific features availability? Could help in the future if we have features dependent on enabled modules as opposed to server version. Something like 

```ts
async function isBitXorAvailable(pgm: MigrationBuilder) {
  try {
    await pgm.db.query('select bit_xorr(1)');
    return true;
  } catch (error: any) {
    if (error.code === '42883' /* UNDEFINED_FUNCTION */) {
      return false;
    }
    throw error;
  }
}
```

rafaelcr: Great idea. I'll do this

src/api/routes/address.ts
rafaelcr: Great point.

I could read the query params inside the chain tip `cacheHandler` to check if it's an explicit anchored request and check the etag based on that. I'd need to add another `ETagType` like `anchored_chain_tip` that only keeps the latest `index_block_hash`. I don't think it would be that hard to do.

I'll open an issue for this.

src/api/controllers/cache-controller.ts
zone117x: Yeah that sounds like a good approach. The previous behavior was simply "check if this locals property was set, regardless if empty value", then error if the property is not set (since that would a bug / 500 error).

src/api/routes/address.ts
rafaelcr: Opened https://github.com/hirosystems/stacks-blockchain-api/issues/1122

zone117x (approved):
Great work @rafaelcr!

PR #1125, opened 2022-03-25
feat: add `pruned` event import mode that ignores some historical events

### LINKED ISSUE #1117: Only process mempool TSV events after a certain block delta threshold

rafaelcr:
Adds a `--mode` option to the `import-events` process, that allows a new `pruned` value to ignore some node events if outside of the (block height - 256) window.

To determine the TSV's block height, a helper was added which streams the file backwards until it finds the last `/new_block` event, from which it takes the `block_height` value.

On local tests, an `import-events` baseline duration of 1:24:17.75 was made for archival mode. The `pruned` mode performed a successful event replay (without BNS import) in:
* 1:12:18.26 by pruning `/new_mempool_tx` and `/drop_mempool_tx` (15% reduction)
* 51:15.56 by pruning `/new_mempool_tx`, `/drop_mempool_tx`, and `/new_microblocks` (40% reduction)

Closes #1117


src/event-replay/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+ * the `block_height` reported by that event. Even though the block produced by that event might
+ * end up being re-org'd, it gives us a reasonable idea as to what the Stacks node thought
+ * the block height was the moment it was sent.
+ * @param filePath - TSV path
+ * @returns `number` found block height, 0 if not found
+ */
+export async function findTsvBlockHeight(filePath: string): Promise<number> {
+  const blockHeightWaiter = waiter<number>();
+  const reverseStream = fsr(filePath, { flags: 'r' });
+  reverseStream.on('data', data => {
```
zone117x, line 16: I don't think this callback approach supports stream backpressure (as in it has unbounded memory usage). 

If `reverseStream` implements the right nodejs stream interface, you could use an async iterator like `for await (const data of reverseStream) { ... }` which should support backpressure and disposal correctly, otherwise pipeline with a transform stream would also work. 

Check out https://nodejs.org/en/docs/guides/backpressuring-in-streams/

package.json
```diff
@@ -120,6 +120,7 @@
     "express": "^4.17.1",
     "express-list-endpoints": "^5.0.0",
     "express-winston": "^4.1.0",
+    "fs-reverse": "0.0.3",
```
zone117x, line 123: What do you think about vendoring this dep? It seems no longer maintained, and repo owner archived it. If we have to define the types anyway, we could just pull in the code and use ts directly? I didn't find any good alternative libs in a cursory search.

src/event-replay/reverse-file-stream.ts
```diff
@@ -0,0 +1,47 @@
...
+  _read(size: number): void {
+    while (this.lineBuffer.length === 0 && this.position > 0) {
+      // Read `size` bytes from the end of the file.
+      const length = Math.min(size, this.position);
+      const buffer = Buffer.alloc(length);
+      this.position = this.position - length;
+      fs.readSync(this.fileDescriptor, buffer, 0, length, this.position);
+
+      // Split into lines to fill the `lineBuffer`.
+      this.remainder = buffer + this.remainder;
```
zone117x, line 30: So this `[type Buffer] + [type string]` I think works because implicit `Buffer.toString([encoding])` is called, and the default encoding is utf8, but probably better to be explicit. I think normally typescript/linters would error here but we must have some rule misconfigured or disabled.

zone117x (approved):
Awesome feature @rafaelcr!

PR #1131, opened 2022-03-29
Add NFT FT metadata flag

MHassanTariq:
## Description

This PR adds a new flag to identify nft/ft strict mode is enabled to keep retrying forever if the metadata retrieval error is recoverable. If it's not, then try a given number of times before failing. 
For details refer to issue #1007 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/event-stream/tokens-contract-handler.ts
+++ b/src/event-stream/tokens-contract-handler.ts
@@ -27,6 +27,7 @@ import {
   logError,
   logger,
   parseArgBoolean,
+  parseArgNumber,
   parseDataUrl,
   REPO_DIR,
   stopwatch,
@@ -46,8 +47,6 @@ type ReadOnlyContractCallResponse =
   | { found: true; value: ClarityValue }
   | { found: false; error: string; retriable: false }
   | { found: false; retriable: true };
-
-const METADATA_RETREIVAL_INTERVAL = 3000;
 
 /**
  * The maximum number of token metadata parsing operations that can be ran concurrently before
@@ -82,6 +81,15 @@ export enum TokenMetadataErrorMode {
 
 function isMetadataStrictModeEnabled() {
   return parseArgBoolean(process.env['STACKS_API_METADATA_STRICT_MODE']);
+}
+
+function getMetadataRetrievalInterval() {
+  const interval = parseArgNumber(process.env['STACKS_API_METADATA_RETRIEVAL_INTERVAL']);
+  if (!interval) {
+    // returning 3000ms for undefined, null and 0ms
+    return 3000;
+  }
+  return interval;
 }
 
 export function isFtMetadataEnabled() {
@@ -677,7 +685,10 @@ export class TokensContractHandler {
       });
     } catch (error: any) {
       // In case there is a network error during fetch.
-      console.log(`Error calling read only contract function value: ${error.message}`);
+      logger.warn(
+        `[token-metadata] Error calling read only contract function value for ${functionName}`,
+        error
+      );
       return { found: false, retriable: true };
     }
 
@@ -705,7 +716,10 @@ export class TokensContractHandler {
       return { found: true, value: parseValue };
     } catch (error: any) {
       // In case there was a connection was interrupted
-      console.log(`Error parsing read only contract function response: ${error.message}`);
+      logger.warn(
+        `[token-metadata] Error parsing read only contract function response for ${functionName}`,
+        error
+      );
       return { found: false, retriable: true };
     }
   }
@@ -718,11 +732,13 @@ export class TokensContractHandler {
     while (!response.found && response.retriable) {
       // In case of strict mode enabled, retry should be aggresive. Otherwise, queued for later.
       if (!isMetadataStrictModeEnabled()) {
-        await timeout(METADATA_RETREIVAL_INTERVAL);
+        await timeout(getMetadataRetrievalInterval());
       }
       response = await this.makeReadOnlyContractCall(functionName, functionArgs);
     }
-    if (response.found) return response.value;
+    if (response.found) {
+      return response.value;
+    }
     // In case the issue is permanent.
     throw new Error(response.error);
   }
--- a/src/event-stream/tokens-contract-handler.ts
+++ b/src/event-stream/tokens-contract-handler.ts
@@ -61,7 +61,7 @@ export enum TokenMetadataErrorMode {
 
 const MAX_RETRIES_METADATA_RETREIVAL = 3;
 
-export function isMetadataStrictModeEnabled() {
+function isMetadataStrictModeEnabled() {
   const opt = process.env['STACKS_API_ENABLE_FT_METADATA']?.toLowerCase().trim();
   return opt === '1' || opt === 'true';
 }
--- a/src/event-stream/tokens-contract-handler.ts
+++ b/src/event-stream/tokens-contract-handler.ts
@@ -17,6 +17,7 @@ import {
   getAddressFromPrivateKey,
   makeRandomPrivKey,
   parseReadOnlyResponse,
+  ReadOnlyFunctionResponse,
   TransactionVersion,
   uintCV,
   UIntCV,
@@ -674,13 +675,16 @@ export class TokensContractHandler {
           'Content-Type': 'application/json',
         },
       });
-    } catch {
+    } catch (error: any) {
       // In case there is a network error during fetch.
+      console.log(`Error calling read only contract function value: ${error.message}`);
       return { found: false, retriable: true };
     }
 
     if (!response.ok) {
-      if (response.status >= 500) return { found: false, retriable: true };
+      if (response.status >= 500) {
+        return { found: false, retriable: true };
+      }
       const msg = await response.text().catch(() => '');
       return {
         found: false,
@@ -689,10 +693,19 @@ export class TokensContractHandler {
       };
     }
     try {
-      const value = await response.json();
-      return { found: true, value: parseReadOnlyResponse(value) };
-    } catch {
+      const value: ReadOnlyFunctionResponse = await response.json();
+      const parseValue = parseReadOnlyResponse(value);
+      if (parseValue.type === ClarityType.ResponseErr) {
+        return {
+          found: false,
+          retriable: false,
+          error: `Function execution failed with error: ${parseValue.value}`,
+        };
+      }
+      return { found: true, value: parseValue };
+    } catch (error: any) {
       // In case there was a connection was interrupted
+      console.log(`Error parsing read only contract function response: ${error.message}`);
       return { found: false, retriable: true };
     }
   }
@@ -701,11 +714,13 @@ export class TokensContractHandler {
     functionName: string,
     functionArgs: ClarityValue[]
   ): Promise<ClarityValue> {
-    let response: ReadOnlyContractCallResponse = { found: false, retriable: true };
+    let response = await this.makeReadOnlyContractCall(functionName, functionArgs);
     while (!response.found && response.retriable) {
-      response = await this.makeReadOnlyContractCall(functionName, functionArgs);
       // In case of strict mode enabled, retry should be aggresive. Otherwise, queued for later.
-      if (!isMetadataStrictModeEnabled()) await timeout(METADATA_RETREIVAL_INTERVAL);
+      if (!isMetadataStrictModeEnabled()) {
+        await timeout(METADATA_RETREIVAL_INTERVAL);
+      }
+      response = await this.makeReadOnlyContractCall(functionName, functionArgs);
     }
     if (response.found) return response.value;
     // In case the issue is permanent.
--- a/src/event-stream/tokens-contract-handler.ts
+++ b/src/event-stream/tokens-contract-handler.ts
@@ -17,19 +17,25 @@ import {
   getAddressFromPrivateKey,
   makeRandomPrivKey,
   parseReadOnlyResponse,
-  ReadOnlyFunctionOptions,
   TransactionVersion,
   uintCV,
   UIntCV,
 } from '@stacks/transactions';
 import { GetStacksNetwork } from '../bns-helpers';
-import { logError, logger, parseArgBoolean, parseDataUrl, REPO_DIR, stopwatch } from '../helpers';
+import {
+  logError,
+  logger,
+  parseArgBoolean,
+  parseDataUrl,
+  REPO_DIR,
+  stopwatch,
+  timeout,
+} from '../helpers';
 import { StacksNetwork } from '@stacks/network';
 import PQueue from 'p-queue';
 import * as querystring from 'querystring';
 import fetch from 'node-fetch';
 import { Evt } from 'evt';
-import { reject } from 'bluebird';
 
 /**
  * This response is for the readOnlyContractCall function and time interval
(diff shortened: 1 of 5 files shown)


src/event-stream/tokens-contract-handler.ts
```diff
@@ -623,10 +631,35 @@ class TokensContractHandler {
...
   }
 
+  async retryReadOnlyContractCall(
+    functionName: string,
+    functionArgs: ClarityValue[]
+  ): Promise<ClarityValue> {
+    let retryCounter = 0;
+    while (true) {
+      retryCounter++;
+      console.log('retryCounter: ', retryCounter);
```
rafaelcr, line 641: Remove this log

rafaelcr (changes requested):

src/event-stream/tokens-contract-handler.ts
```diff
@@ -623,10 +631,35 @@ class TokensContractHandler {
...
+      } catch (error: any) {
+        console.log('failed with error ', error);
+        if (!isMetadataStrictModeEnabled()) {
+          throw error;
+        } else if (
+          !error.message.includes('Error calling read-only function') &&
+          retryCounter === MAX_RETRIES_METADATA_RETREIVAL
+        ) {
+          throw error;
+        }
```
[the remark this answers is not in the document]
MHassanTariq: @rafaelcr the main problem identifying recoverable and fatal errors is that I don't get the status code for `callReadOnlyFunction`. It is implemented [here](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/builders.ts#L1289) in stacks/transaction. 

One approach would be to find the status code from the [error message](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/builders.ts#L1321) and keep retrying incase the status code is 500. What do you think about this approach?

Also, regarding the disabled strict mode, do you suggest we still try 3 times before throwing the error?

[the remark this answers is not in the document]
zone117x: I don't think there's any particularly compelling reason to use an existing helper function for making the call read-only function network request. It's not much code/complexity to implement the fetch here in order to have more fine grained control and observability into error responses.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -319,9 +335,10 @@ class TokensContractHandler {
     } else if (isCompliantNft(args.smartContractAbi)) {
       this.tokenKind = 'nft';
     } else {
-      throw new Error(
-        `TokenContractHandler passed an ABI that isn't compliant to FT or NFT standards`
-      );
+      this.tokenKind = 'nft';
+      // throw new Error(
+      //   `TokenContractHandler passed an ABI that isn't compliant to FT or NFT standards`
+      // );
```
zone117x, line 341: Was this change intentional? It looks like a bug.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +643,77 @@ class TokensContractHandler {
...
+      functionName
+    );
+    const args = functionArgs.map(arg => cvToHex(arg));
+
+    const body = JSON.stringify({
+      sender: this.address,
+      arguments: args,
+    });
+
+    const response = await fetchPrivate(url, {
```
zone117x, line 662: My understanding is that both this `await` and the below `await response.json()` can throw if there's a network error, e.g. dns resolution failure, host unreachable, connection interrupted, etc.

The error handling below checking `response.ok` and `response.status` is only applicable the request was successful at the http protocol layer (i.e. headers were received without interrupt, then the body was received without interrupt).

zone117x: So I think that would require a try/catch block around these statements. You should be able to test these kind of network failures in a unit test. There should be a test for network unreachable (like using a bogus hostname), another test should be for connection interrupt i.e. have an http server reply with the headers, flush connection data, then force close the socket without sending the http response body.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +643,77 @@ class TokensContractHandler {
...
+      let response: ReadOnlyContractCallResponse = { found: false, retriable: true };
+      if (isMetadataStrictModeEnabled()) {
+        while (!response.found && response.retriable) {
+          response = await this.makeReadOnlyContractCall(functionName, functionArgs);
+        }
+        if (response.found) return resolve(response.value);
+        // incase the issue is not temporary.
+        else if (!response.retriable) return reject(new Error(response.error));
+      } else {
+        const interval = setInterval(async () => {
```
zone117x, line 697: The `setInterval` usage here is a bit confusing to reason about. It looks like essentially a while-loop in the form of callbacks, right? In that case, just a regular while loop with an `await timeout(interval)` would be a lot more clear.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +648,76 @@ class TokensContractHandler {
...
-      functionName: functionName,
-      functionArgs: functionArgs,
-      network: this.stacksNetwork,
-    };
-    return await callReadOnlyFunction(txOptions);
+    let response: ReadOnlyContractCallResponse = { found: false, retriable: true };
+    while (!response.found && response.retriable) {
+      response = await this.makeReadOnlyContractCall(functionName, functionArgs);
+      // In case of strict mode enabled, retry should be aggresive. Otherwise, queued for later.
+      if (!isMetadataStrictModeEnabled()) await timeout(METADATA_RETREIVAL_INTERVAL);
```
zone117x, line 708: Style nit: expression blocks should always have brackets. I.e.
```
if (<condition>) {
  <expression>
}
```
The syntax `if (<condition>) <expression>` is error prone and _should_ be a linter error but looks like that's broken or misconfigured right now.

zone117x, line 708: Won't the way this is coded make calls to `retryReadOnlyContractCall` _always_ await the timeout, even if successful the first time?

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +648,76 @@ class TokensContractHandler {
...
+      return {
+        found: false,
+        retriable: false,
+        error: `Error calling read-only function. Response ${response.status}: ${response.statusText}. Attempted to fetch ${url} and failed with the message: "${msg}"`,
+      };
+    }
+    try {
+      const value = await response.json();
+      return { found: true, value: parseReadOnlyResponse(value) };
+    } catch {
```
zone117x, line 694: This empty `catch` statement without an error binding is strange. It looks like it was historically non-standard syntax, then was standardized in ES2019?
Regardless of if this is valid syntax, it indicates poor error handling. Errors should almost never be swallowed. In this case, it should at least log something useful.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +648,76 @@ class TokensContractHandler {
...
+          'Content-Type': 'application/json',
+        },
+      });
+    } catch {
+      // In case there is a network error during fetch.
+      return { found: false, retriable: true };
+    }
+
+    if (!response.ok) {
+      if (response.status >= 500) return { found: false, retriable: true };
```
zone117x, line 683: The lack of comments around this condition requires reviewers to do a lot of research. It's very unclear to me if this check is correct, too aggressive, or not aggressive enough.

 For example:
* What happens if the stacks-node RPC server is responsive, but hasn't yet synced to the block with this smart-contract? Is there a 2xx OK status but error in the body? Is there a non-2xx status but indication that the error is temporary until synced? Etc.
* Can the stacks-node ever respond with 2xx OK status but with an http response payload indicating some kind of error condition?
* What is the http response for a readonly contract-call request that was successfully processed by the stacks-node RPC server and successfully evaluated by the Clarity VM, but resulted in a Clarity function error result? Does it return 2xx OK but have an error response body?


At the least, these need tested and documented. Ideally there would also be integration tests against the stacks-node as well.

zone117x (commented):
It looks like the current goal of this PR is to make it so that FT metadata is _eventually_ resolved when in strict mode (although see my previous comments where it looks like that goal is not accomplished). 

However, this doesn't solve issue #1007
> guarantee the metadata is available when required

If a given FT is detected and in the process of resolving, and at the same time something like a Rosetta call is made to fetch info about that FT, it will result in critical errors for the typical systems that ingest Rosetta data (probably also other FT related endpoints as well). A solution to this involves any endpoints querying FT metadata to detect when the FT metadata is not-yet-resolved, and to await the resolve result. This can be done in a number of ways, e.g. polling/watching the pg db, or checking the `TokensContractHandler` instance queue, etc.

This PR can be scoped down to "ensure FT metadata is _eventually_ resolved", which is mostly fine for our products like the wallet and explorer, however, there should be a follow-up issue/PR for ensuring FT metadata is never missing from endpoints when requested.

cc @rafaelcr for additional thoughts

src/event-stream/tokens-contract-handler.ts
MHassanTariq: Yes, as long as the server is responding, it will return 200. I have also added a test without deploying the contract call. It throws an error with the message `Unchecked(NoSuchContract({contract_id))` [here](https://github.com/hirosystems/stacks-blockchain-api/pull/1131/files#diff-1be0acee815b65d2c2026b016d6bfc8eff2e08f184b79c128001d50680f71421R184-R203).

Yes, the same test is applicable for 2nd question. It returns headers with 200 and OK. However, fails during `response.json()` since the contract doesn't exist.

Added another [test](https://github.com/hirosystems/stacks-blockchain-api/pull/1131/files#diff-1be0acee815b65d2c2026b016d6bfc8eff2e08f184b79c128001d50680f71421R156-R180) for testing the third question. The contract function returns the object with `ClarityType.ResponseErr` as type and value for it.

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +649,90 @@ class TokensContractHandler {
...
+        return {
+          found: false,
+          retriable: false,
+          error: `Function execution failed with error: ${parseValue.value}`,
+        };
+      }
+      return { found: true, value: parseValue };
+    } catch (error: any) {
+      // In case there was a connection was interrupted
+      console.log(`Error parsing read only contract function response: ${error.message}`);
```
zone117x, line 708: These should use the `logger` instance and match the convention used in the rest of the module, e.g.
```js
logger.warn(
  `[token-metadata] ...<information about the error and contract ID>...`,
  error
);
```

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +649,90 @@ class TokensContractHandler {
...
-    return await callReadOnlyFunction(txOptions);
+    let response = await this.makeReadOnlyContractCall(functionName, functionArgs);
+    while (!response.found && response.retriable) {
+      // In case of strict mode enabled, retry should be aggresive. Otherwise, queued for later.
+      if (!isMetadataStrictModeEnabled()) {
+        await timeout(METADATA_RETREIVAL_INTERVAL);
+      }
+      response = await this.makeReadOnlyContractCall(functionName, functionArgs);
+    }
+    if (response.found) return response.value;
```
zone117x, line 725: Nit: brace style

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +649,90 @@ class TokensContractHandler {
...
+      response = await fetchPrivate(url, {
+        method: 'POST',
+        body,
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      });
+    } catch (error: any) {
+      // In case there is a network error during fetch.
+      console.log(`Error calling read only contract function value: ${error.message}`);
```
zone117x, line 680: Same here, should use the `logger` instance

src/event-stream/tokens-contract-handler.ts
```diff
@@ -626,27 +649,90 @@ class TokensContractHandler {
...
+      const msg = await response.text().catch(() => '');
+      return {
+        found: false,
+        retriable: false,
+        error: `Error calling read-only function. Response ${response.status}: ${response.statusText}. Attempted to fetch ${url} and failed with the message: "${msg}"`,
+      };
+    }
+    try {
+      const value: ReadOnlyFunctionResponse = await response.json();
+      const parseValue = parseReadOnlyResponse(value);
```
zone117x, line 697: Is `parseReadOnlyResponse` guaranteed to only ever throw if the error is actually retriable and will eventually succeed?

src/event-stream/tokens-contract-handler.ts
```diff
@@ -7,27 +7,48 @@ import {
...
+/**
+ * This response is for the readOnlyContractCall function and time interval
+ * for calls if strict mode is disabled
+ */
+type ReadOnlyContractCallResponse =
+  | { found: true; value: ClarityValue }
+  | { found: false; error: string; retriable: false }
+  | { found: false; retriable: true };
+
+const METADATA_RETREIVAL_INTERVAL = 3000;
```
zone117x, line 50: This should be configurable, probably with another env var.

src/event-stream/tokens-contract-handler.ts
zone117x: For example, I think the stacks-node has a configurable max cost for these readonly function calls, so what happens if someone writes a contract with a function that is too expensive, will it cause this to throw and retry forever?

MHassanTariq: It will throw error in 2 situations:
1- If the response is not ok. Meaning, there was a server side issue and not an issue with the function. This check is given [here](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/utils.ts#L256). 
2- If the `ClarityValue` returned from the function is not formatted correctly. In that case, the errors are thrown by `deserializeCV`. For instance, [Unrecognize Clarity Type](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/clarity/deserialize.ts#L39), [Clarity Type unavailable](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/clarity/deserialize.ts#L111-L113), [Invalid Clarity name](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/clarity/types/tupleCV.ts#L15), [Invalid byte length for Contract Name](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/clarity/types/principalCV.ts#L59), [Invalid Buffer length](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/clarity/types/bufferCV.ts#L11), [Range Errors for uintCV](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/clarity/types/intCV.ts#L36-L42), and [Range Errors for intCV](https://github.com/hirosystems/stacks.js/blob/master/packages/transactions/src/clarity/types/intCV.ts#L17-L24). These errors could be an issue that will keep retrying and never resolve, considering the function will always return the same result for the same `functionArgs` provided.

It should retry in the first case. But since we’re already catering to that issue with `retriable: true` [here](https://github.com/hirosystems/stacks-blockchain-api/pull/1131/files#diff-94b60ecbb28bcac15bf49d0f6601de1d8273ff795f81ac60f19a70eed6b429a7R684-R687), so we don’t actually need to worry about this error here. 
However, for the second case, I believe I should split these lines in `try block` such that,
`const value: ReadOnlyFunctionResponse = await response.json();` would be a retriable issue, considering if the headers were sent but connection broke before sending the body to the client.
`const parseValue = parseReadOnlyResponse(value);` should not be a retriable issue. Instead, it should log the actual error found, parsing the response.
Let me know if I’m wrong here. 
Thanks!

zone117x: From what I can tell, this endpoint can return an object:
```ts
interface ReadOnlyFunctionErrorResponse {
  okay: false;
  cause: string;
}
```
What areas handle the retrying or _not_ retrying for when this is `okay: false`.

What happens if the query cost exceeds the limits of the read-only-call parameters defined at
https://github.com/stacks-network/stacks-blockchain/blob/26bfd5fcdc1f25106288f18ced05290b569f6abb/testnet/stacks-node/src/config.rs#L1477-L1481
?

Issue #1119, opened 2022-03-23
inconsistent tx data across public API instances


pseudozach:
Running `/extended/v1/address/SP2507VNQZC9VBXM7X7KB4SF4QJDJRSWHG4V39WPY.stxswap_v8/transactions` endpoint gives different values
- https://stacks-node-api.stacks.co/extended/v1/address/SP2507VNQZC9VBXM7X7KB4SF4QJDJRSWHG4V39WPY.stxswap_v8/transactions: "total": 642
- [my node](http://api.lnswap.org:3999/extended/v1/address/SP2507VNQZC9VBXM7X7KB4SF4QJDJRSWHG4V39WPY.stxswap_v8/transactions): "total": 656 
- https://mainnet.syvita.org/extended/v1/address/SP2507VNQZC9VBXM7X7KB4SF4QJDJRSWHG4V39WPY.stxswap_v8/transactions: "total": 686 -> seems to be the most realistic
- https://stacks-node-api.alexlab.co/extended/v1/address/SP2507VNQZC9VBXM7X7KB4SF4QJDJRSWHG4V39WPY.stxswap_v8/transactions: "total": 686,

- I'm guessing similar inconsistent results will happen for other data as well which could be a problem as user facing apps depend on this data.
- Note that in my case I've indeed done `event replay` to upgrade from 2.05 to 2.05.1 for the lnswap node so it could be related.
- Another (possibly related) issue is that I've seen same txid twice in the stacks.co instance:
Loop through and save all transactions for the above endpoint into a json file
You can see this txid: `0x04bcbe61f525f795ca736e7ca44213081763f907f39be378d22e0db263deb995` twice on 2 blockheights `"block_height": 45852,` and `"block_height": 45853,` and same nonce `245`

cc @rafaelcr

caoer:
@rafaelcr 
I think it might not be caused from stacks-node or reply. I did see the error logs while replay the events regarding sending duplicated events and inserting the same tx_id into the mempool. 

however, with [a simple script](https://gist.github.com/caoer/7e35e137d2de1efaedd318233bb4d245) to check for the missing txs between 1.0.x and 3.0.x. we can find out what is missing in the newer API: 

it looks like that the tx is only missing from `/extended/v1/address/{principal}/transactions` endpoint, however, we can still get the tx directly by calling `extended/v1/tx/{tx_id}`, the transactions is not missing in the db, but we just can not get it from transactions list API endpoint.

[...]

PR #1135, opened 2022-04-01
fix!: duplicate block re-org scenario in `/transactions`

rafaelcr:
This PR fixes a re-org scenario in which the Stacks node sent the same block message twice i.e. with the same `block_height` and containing the same transactions but with a different `index_block_hash`. The second block was consequently being marked as `canonical = FALSE` causing the `principal_stx_txs` table to do the same with its related transactions.

To fix this, the `principal_stx_txs` table now also tracks the `index_block_hash` and `microblock_hash` so it can better response to re-orgs.

Related to #1119


src/datastore/postgres-store.ts
```diff
@@ -1714,9 +1718,11 @@ export class PgDataStore
     // Update `principal_stx_txs`
     await client.query(
       `UPDATE principal_stx_txs
-      SET canonical = $1, microblock_canonical = $2
-      WHERE tx_id = ANY ($3)`,
-      [args.isCanonical, args.isMicroCanonical, updatedMbTxs.map(tx => hexToBuffer(tx.tx_id))]
+      SET microblock_canonical = $1, canonical = $2, index_block_hash = $3
```
zone117x, line 1721: Have you performed a full (archival) event-replay with this PR? I'm wondering if the `principal_stx_txs` unique constraint might get violated during the re-org handling around here.

zone117x (approved):
Left a question about testing this change with a full event-replay, otherwise this fix looks good!

src/datastore/postgres-store.ts
rafaelcr: Just performed one, and it was all good 👍

PR #1146, opened 2022-04-13
chore(ci): upload code coverage to coveralls.io

zone117x:
Upload code coverage report to coveralls.io which we can use as a sanity check against the buggy reports we've been getting from codecov, and possibly switch over if it ends up working better in general.


rafaelcr (approved):
Nice 👌

PR #1148, opened 2022-04-19
feat: refactor pg classes, switch to postgres.js

rafaelcr:
This PR refactors the `postgres-store.ts` file and divides it into several single-purpose files/classes under the `src/datastore` directory:
* `pg-store.ts` - the `PgStore` class which contains all functions which only read data from the DB
* `pg-write-store.ts` - the `PgWriteStore` class which extends `PgStore` to add insert/update functions for node event ingestion
* `connection.ts` - helpers that handle DB connections
* `migrations.ts` - helpers that manage DB migrations
* `common.ts` - db types
* `helpers.ts` - parsers, utilities, db query result types

This change also includes the switch from the [node-postgres](https://github.com/brianc/node-postgres) library (no longer maintained) to [postgres.js](https://github.com/porsager/postgres) to speed up data processing.

TODO:
- [ ] Archival event replay tests
- [x] Unit test migration
- [ ] postgres.js migration
- [ ] Inactive/canary testing


src/datastore/pg-write-store.ts
```diff
@@ -0,0 +1,2280 @@
...
+      canonical: minerReward.canonical,
+      recipient: minerReward.recipient,
+      coinbase_amount: minerReward.coinbase_amount.toString(),
+      tx_fees_anchored: minerReward.tx_fees_anchored.toString(),
+      tx_fees_streamed_confirmed: minerReward.tx_fees_streamed_confirmed.toString(),
+      tx_fees_streamed_produced: minerReward.tx_fees_streamed_produced.toString(),
+    };
+    const result = await sql`
+      INSERT INTO miner_rewards ${sql(values)}
+      RETURNING *
```
zone117x, line 421: There's a lot of instances of using `RETURNING *` when only the row count is needed. This can be much more efficient by removing `RETURNING *` and using `result.count` instead of `result.length`.

rafaelcr: Great catch, thanks. Must've missed that from the docs (or it isn't there)

zone117x (approved):
This is an awesome refactor! I've been building off this branch in https://github.com/hirosystems/stacks-blockchain-api/pull/1166, and so far it's been working well. We need to do QA smoke tests with the wallets/explorer, then some prod canary testing. Otherwise LGTM!

rafaelcr:
Thanks @zone117x , I'm going to merge this PR and send another one to change the package dependency once postgres.js issues a release

PR #1159, opened 2022-05-02
fix: `/v1/names/[:name]` name resolution

### LINKED ISSUE #1158: Fix /v1/names/[NAME] from returning old owner addresses
As originally reported with https://github.com/hirosystems/stacks-wallet-web/issues/2394#issuecomment-1114258999

rafaelcr:
This PR fixes a bug where the owner of a BNS name was not correctly reported through the `/v1/names/[:name]` endpoint.

The `names` table does not really track true name ownership, given that it is mostly updated when there is a zonefile change or a set of specific contract calls, causing re-orgs not being accounted for correctly. The `nft_custody` view is now used to get the latest owner given that a name is also an NFT and we get all events for it.

Incidentally, this also fixes a previously unknown bug where `nft_custody` was not being refreshed on a deep chain re-org.

Closes #1158

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -244,7 +244,7 @@ export async function startApiServer(opts: {
       const router = express.Router();
       router.use(cors());
       router.use('/namespaces', createBnsNamespacesRouter(datastore));
-      router.use('/names', createBnsNamesRouter(datastore));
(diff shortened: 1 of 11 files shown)


zone117x (approved):
Looks great! Thanks @rafaelcr

PR #1161, opened 2022-05-04
fix: additional logging for tx multicasting results

zone117x:
Add logging for http success status responses in tx multicasting


zone117x:
IIRC we made this change to debug some behavior, and we no longer need this.

PR #1165, opened 2022-05-10
feat: add strict ft/nft metadata processing mode for better error handling

rafaelcr:
This PR solves #1007 by making the following changes to the FT/NFT metadata processor:

1. Separates `tokens-contract-handler.ts` into separate queue, handler, and helper files under `src/token-metadata`
2. Creates a `RetryableTokenMetadataError` class and tags each retryable error inside the metadata processor with this class (if applicable) so we can decide what to do later in each case
3. Adds `STACKS_API_TOKEN_METADATA_STRICT_MODE` env var that controls how we handle retryable errors:
    * If enabled, contracts with retryable errors will be processed again indefinitely
    * If disabled, processing will be retried for that contract until `STACKS_API_TOKEN_METADATA_MAX_RETRIES` is reached
4. Starts listening to `blockUpdate` events inside the metadata processor so we can try to drain the queue with every new block. This allows us to retry failed contracts.


src/migrations/1621511823100_token_metadata_queue.ts
```diff
@@ -28,6 +28,11 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
     processed: {
       type: 'boolean',
       notNull: true,
+    },
+    retry_count: {
+      type: 'integer',
+      notNull: true,
+      default: 0,
```
zone117x, line 35: If this is moved to a new migration file (e.g. `alter table .. add column ...`) we can merge and deploy to beta testing without event-replay.

src/tests-tokens/strict-mode-tests.ts
```diff
@@ -0,0 +1,266 @@
+import * as nock from 'nock';
```
zone117x, line 1: Another useful couple tests could be around the `performFetch` function error conditions when called by `getMetadataFromUri`. For example, exceeding `METADATA_FETCH_TIMEOUT_MS` and `METADATA_MAX_PAYLOAD_BYTE_SIZE`.

src/token-metadata/tokens-contract-handler.ts
```diff
@@ -0,0 +1,481 @@
...
+
+    let contractCallDecimals: number | undefined;
+    const decimalsResult = await this.readUIntFromContract('get-decimals');
+    if (decimalsResult) {
+      contractCallDecimals = Number(decimalsResult.toString());
+    }
+
+    let metadata: FtTokenMetadata | undefined;
+    if (contractCallUri) {
+      metadata = await this.getMetadataFromUri<FtTokenMetadata>(contractCallUri);
```
zone117x, line 197: I'm wondering what to do for conditions where this external URL cannot be fetched/parsed _but_ the above contract-calls to get `decimals`, `symbol`, etc succeeded. 

You could imagine the third-party server for this metadata being unavailable for an extended period of time or even go offline indefinitely, while most apps could still make do with the information from the contract-call (primarily `decimals`).

rafaelcr: Great point... IMO for now (this PR) we should keep retrying but we could consider later adding another retry limit just for external metadata that is also enforced during strict mode.

zone117x: I'd hazard a guess that there's probably already cases of a 3rd party urls for this already permanently unavailable, and if these are considered retriable errors, then it seems like it could result in two critical situations:
1) The _more_ critical data and already resolved data from the contract-call, like `decimals`, will never be written to the db since the 3rd party url fetch would cause the function to exit here before finishing.
2) If `TOKEN_METADATA_PARSING_CONCURRENCY_LIMIT` (5) or more of these exist, the queue could be stuck indefinitely trying to process them, possibly never getting to process others.

I agree we don't have to solve this entirely in this PR, but I think we should try to prevent either of the above. It could just be something like `metadata = await this.getMetadataFromUri<FtTokenMetadata>(contractCallUri)` is wrapped in a for loop where it can retry fetching a few times, but once it fails those, it is A) _not_ considered retriable, and B) it lets the function continue (leaving `metadata` as undefined).

The only real downside of this approach would be that `image` urls would not be available if the 3rd party url is unresponsive for the duration of the fixed set of retries -- but the fault for that is much more attributable to the 3rd party than us.

zone117x: We could even fill in one of the "image failed to load" placeholders from internet explorer to indicate this kind of failure 😆 

image

rafaelcr: 🤣 lmao that would be great.

I think you're absolutely right though, I didn't consider the queue being permanently full from these dead URLs being retried forever.

Your idea for a for loop sounds great, I'll add that to this PR.

src/tests-tokens/strict-mode-tests.ts
```diff
@@ -257,6 +262,60 @@ describe('token metadata strict mode', () => {
...
+      result: cvToHex(stringCV('http://indigo.com/nft.jpeg', 'ascii')),
+    };
+    nock('http://127.0.0.1:20443')
+      .post(
+        '/v2/contracts/call-read/SP176ZMV706NZGDDX8VSQRGMB7QN33BBDVZ6BMNHD/project-indigo-act1/get-token-uri'
+      )
+      .reply(200, mockTokenUri);
+    nock('http://indigo.com')
+      .get('/nft.jpeg')
+      .delay(METADATA_FETCH_TIMEOUT_MS + 1000)
```
zone117x, line 277: Perhaps we make `METADATA_FETCH_TIMEOUT_MS` env var configurable so that we can set it to a low value here and avoid the test taking (10 + 1) seconds to complete

src/tests-tokens/strict-mode-tests.ts
```diff
@@ -257,6 +262,60 @@ describe('token metadata strict mode', () => {
...
+    const mockTokenUri = {
+      okay: true,
+      result: cvToHex(stringCV('http://indigo.com/nft.jpeg', 'ascii')),
+    };
+    nock('http://127.0.0.1:20443')
+      .post(
+        '/v2/contracts/call-read/SP176ZMV706NZGDDX8VSQRGMB7QN33BBDVZ6BMNHD/project-indigo-act1/get-token-uri'
+      )
+      .reply(200, mockTokenUri);
+    const bigAssBuffer = Buffer.alloc(METADATA_MAX_PAYLOAD_BYTE_SIZE + 100);
```
zone117x, line 303: 🤣

zone117x (approved):
Looks great, thanks @rafaelcr!

Issue #1164, opened 2022-05-06
Incorrect/duplicated tx output for microblock endpoint


wileyj:
https://discord.com/channels/621759717756370964/909867922971820102/972181139148111953

consider this microblock:
https://stacks-node-api.mainnet.stacks.co/extended/v1/microblock/0xf5c42c81eb2f1d41b6a8fefe3be554229dce203e35e109ea4d37f10c707d439f

with output: 
```
{
  "canonical": true,
  "microblock_canonical": true,
  "microblock_hash": "0xf5c42c81eb2f1d41b6a8fefe3be554229dce203e35e109ea4d37f10c707d439f",
  "microblock_sequence": 0,
  "microblock_parent_hash": "0x59525ae493fc95498287fa2ba0dea3568bc230fac9b7cdace53e9960b3c84f43",
  "block_height": 57149,
  "parent_block_height": 57148,
  "parent_block_hash": "0x59525ae493fc95498287fa2ba0dea3568bc230fac9b7cdace53e9960b3c84f43",
  "block_hash": "0xa72a0b8a09dedcb03760947afbc4aa597834dc9a515decdeca0c101fafc0c75e",
  "txs": [
    "0x9a197d4a2763a1e4589a6d9c31d2794a5be223e45aecc91d996b227522deb1da",
    "0x9a197d4a2763a1e4589a6d9c31d2794a5be223e45aecc91d996b227522deb1da"
  ],
  "parent_burn_block_height": 733078,
  "parent_burn_block_hash": "0x00000000000000000008b4182984bf6b0c3acb7d4eb79093267d9bc6da3af5e7",
  "parent_burn_block_time": 1650672550,
  "parent_burn_block_time_iso": "2022-04-23T00:09:10.000Z"
}
```

Notice the duplicate **txs** for `0x9a197d4a2763a1e4589a6d9c31d2794a5be223e45aecc91d996b227522deb1da`

**Expected behaviour**: tx `0x9a197d4a2763a1e4589a6d9c31d2794a5be223e45aecc91d996b227522deb1da` would only be listed once

PR #1167, opened 2022-05-12
Remove duplicate txs microblock

MHassanTariq:
## Description

This PR removes duplicate `tx_ids` returned for `/extended/v1/microblock/{hash}`. 

For details refer to issue #1164 

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


src/datastore/postgres-store.ts
```diff
@@ -1878,7 +1878,7 @@ export class PgDataStore
       }
       const txQuery = await client.query<{ tx_id: Buffer }>(
         `
-        SELECT tx_id
+        SELECT DISTINCT(tx_id)
```
zone117x, line 1881: Just a `distinct` here seems like a bandaid, shouldn't we try to prevent the query from returning the duplicate transactions with a WHERE clause? Most likely needs `where canonical = true and microblock_canonical = true`.

src/tests/api-tests.ts
```diff
@@ -10813,6 +10813,58 @@ describe('api tests', () => {
...
+    const microblock = new TestMicroblockStreamBuilder().addMicroblock({ microblock_hash }).build();
+    // adding a tx in a block with canonical and microblock_canonical true
+    const block1 = new TestBlockBuilder({ block_hash: '0x1234', block_height: 1 })
+      .addTx({
+        microblock_canonical: true,
+        // tx is in the same microblock_hash
+        microblock_hash,
+        canonical: true,
+        block_hash: '0x1234',
+      })
```
rafaelcr, line 10827: This tx should be added to the `TestMicroblockStreamBuilder` and then written to the db using `updateMicroblocks`

rafaelcr (changes requested):

PR #1168, opened 2022-05-12
refactor: move `PgNotifier` to use postgres lib

rafaelcr:


src/datastore/pg-notifier.ts
```diff
@@ -56,38 +54,28 @@ type PgNotificationCallback = (notification: PgNotification) => void;
...
-    await this.subscriber.connect();
-    await this.subscriber.listenTo(this.pgChannelName);
+    try {
+      this.listener = await this.sql.listen(
+        this.pgChannelName,
+        message => eventCallback(JSON.parse(message) as PgNotification),
+        () => logger.info(`PgNotifier connected, listening on channel: ${this.pgChannelName}`)
+      );
+    } catch (error) {
+      logError('PgNotifier fatal connection error', error);
```
zone117x, line 77: Should this re-throw the error here after logging (which I think should exit the program)?

zone117x (approved):
Surprisingly clean and minimal refactor 👍

src/datastore/pg-notifier.ts
rafaelcr: Great point

Issue #1084, opened 2022-02-26
Optimize `getMicroblocks` query


rafaelcr:
PgHero is reporting this slow query (~4 seconds):
<img width="718" alt="Screen Shot 2022-02-26 at 12 16 24" src="https://user-images.githubusercontent.com/216195/155856607-674bc094-b463-48a9-88a8-f562262ba134.png">

It's used in `getMicroblocks` here: https://github.com/hirosystems/stacks-blockchain-api/blob/develop/src/datastore/postgres-store.ts#L1820

We should optimize it to try to avoid the `LEFT JOIN` to `txs` which is the most expensive operation here.

PR #1179, opened 2022-05-19
Optimize `getMicroblocks` query

MHassanTariq:
## Description

This PR optimizes the `getMicroblocks` query. Replaced `LEFT JOIN` with two different queries. One query is used to get `microblocks`. Other is used to get their respective `txs`. 
Time taken with this approach is 1/4th the previous approach with 900 microblocks and 901 txs.
Test is added to check the functionality for this query too.

For details refer to issue #1084

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/tests/api-tests.ts
+++ b/src/tests/api-tests.ts
@@ -10846,6 +10846,7 @@ describe('api tests', () => {
     const result = await supertest(api.server).get(`/extended/v1/microblock/${microblock_hash}`);
     expect(result.body.txs).toHaveLength(1);
     expect(result.body.txs[0]).toEqual(tx_id);
+  });
   test('/microblock', async () => {
     const microblock_hash = '0x0fff';
     const block = new TestBlockBuilder({ block_hash: '0x1234', block_height: 1 }).build();


src/datastore/pg-store.ts
```diff
@@ -448,42 +448,39 @@ export class PgStore {
...
-          ORDER BY block_height DESC, microblock_sequence DESC
-          LIMIT ${args.limit}
-          OFFSET ${args.offset}
-        ) microblocks
-        LEFT JOIN (
-          SELECT tx_id, tx_index, microblock_hash
-          FROM txs
-          WHERE canonical = true AND microblock_canonical = true
-          ORDER BY tx_index DESC
-        ) txs
```
rafaelcr, line 467: What's inefficient about this join here is that it is first creating a *huge* table of all canonical txs sorted by `tx_index` and then filters it on `microblock_hash`.

You shouldn't need to separate this into several queries, instead, try to perform the JOIN *without* sub-queries:
```
SELECT ... FROM microblocks
LEFT JOIN txs USING (microblock_hash)
WHERE ...
```

src/tests/api-tests.ts
```diff
@@ -10813,6 +10813,32 @@ describe('api tests', () => {
...
+    const microblock = microblocks.microblocks[0];
+    const block = new TestBlockBuilder()
+      .addTx({
+        microblock_hash: microblock.microblock_hash,
+        tx_id: '0xffff',
+      })
+      .addTx({
+        microblock_hash: microblock.microblock_hash,
+        tx_id: '0x1234',
+      })
```
rafaelcr, line 10827: These txs should be added to the `TestMicroblockStreamBuilder`, not to the block

rafaelcr (changes requested):

src/datastore/pg-store.ts
```diff
@@ -451,22 +451,13 @@ export class PgStore {
...
-        ON microblocks.microblock_hash = txs.microblock_hash
-        ORDER BY microblocks.block_height DESC, microblocks.microblock_sequence DESC, txs.tx_index DESC
+      SELECT microblocks.*, txs.tx_id 
+      FROM microblocks LEFT JOIN txs USING(microblock_hash)
+      WHERE microblocks.canonical = true AND microblocks.microblock_canonical = true AND 
+        ((txs.canonical = true AND txs.microblock_canonical = true) OR (txs.canonical IS NULL OR txs.microblock_canonical IS NULL))
```
rafaelcr, line 457: `canonical` and `microblock_canonical` can never be `null`. You should only check for both of them = TRUE

rafaelcr (changes requested):

Issue #1173, opened 2022-05-16
blocks endpoint is pretty slow consistently 


aulneau:
I think there might be a regression, or some room for optimization when it comes to the `/blocks` extended endpoint. As I've been using it, it's always one of the slowest endpoints compared to others from the api. Typically it's always over 1 seconds for me. I've been playing around with alternative ways of exposing the underlying data from the hiro api database, and when comparing the same kind of view from something like hasura, the performance is much, much better (100ms compared to ~1.96s):

**Hiro API:**
<img width="1218" alt="Screen Shot 2022-05-16 at 9 00 45 AM" src="https://user-images.githubusercontent.com/11803153/168611008-532eaed2-6d5f-4798-9933-b3e925222571.png">

**Same data, Hasura graphql:**
<img width="1568" alt="Screen Shot 2022-05-16 at 9 03 19 AM" src="https://user-images.githubusercontent.com/11803153/168611038-6474f444-8a3f-4797-8352-8a3bc5eedfa8.png">

PR #1190, opened 2022-05-31
Optimized block endpoint

MHassanTariq:
## Description

This PR optimizes `/block` endpoint. It changes the previous approach of individual queries for each block to 3 queries for `blocks`, `txs`, and `microblocks`. 
For details refer to issue #1173 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -413,7 +413,7 @@ export async function getMicroblocksFromDataStore(args: {
 }
 
 export async function getBlocksMetadata(args: { limit: number; offset: number; db: PgStore }) {
-  const blocks = await args.db.getMetadataBlocks({ limit: args.limit, offset: args.offset });
+  const blocks = await args.db.getBlocksWithMetadata({ limit: args.limit, offset: args.offset });
   const results = blocks.results.map(block =>
     parseDbBlock(
       block.block,
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -412,6 +412,20 @@ export async function getMicroblocksFromDataStore(args: {
   };
 }
 
+export async function getBlocksMetadata(args: { limit: number; offset: number; db: PgStore }) {
+  const blocks = await args.db.getMetadataBlocks({ limit: args.limit, offset: args.offset });
+  const results = blocks.results.map(block =>
+    parseDbBlock(
+      block.block,
+      block.txs,
+      block.microblocks_accepted,
+      block.microblocks_streamed,
+      block.microblock_tx_count
+    )
+  );
+  return { results, total: blocks.total };
+}
+
 export async function getBlockFromDataStore({
   blockIdentifer,
   db,
--- a/src/api/controllers/db-controller.ts
+++ b/src/api/controllers/db-controller.ts
@@ -412,7 +412,7 @@ export async function getMicroblocksFromDataStore(args: {
   };
 }
 
-export async function getBlocksMetadata(args: { limit: number; offset: number; db: PgStore }) {
+export async function getBlocksWithMetadata(args: { limit: number; offset: number; db: PgStore }) {
   const blocks = await args.db.getBlocksWithMetadata({ limit: args.limit, offset: args.offset });
   const results = blocks.results.map(block =>
     parseDbBlock(
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -7,11 +7,12 @@ import { ChainID, ClarityAbi } from '@stacks/transactions';
 import { getTxTypeId } from '../api/controllers/db-controller';
 import {
   assertNotNullish,
+  FoundOrNot,
+  hexToBuffer,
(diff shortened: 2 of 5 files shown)


src/datastore/pg-store.ts
```diff
@@ -396,6 +396,80 @@ export class PgStore {
     });
   }
 
+  async getMetadataBlocks({ limit, offset }: { limit: number; offset: number }) {
```
zone117x, line 399: This function name doesn't make sense. Should just make a plural version of the existing `getBlockWithMetadata`, i.e. `getBlocksWithMetadata`.

src/datastore/pg-store.ts
```diff
@@ -396,6 +396,80 @@ export class PgStore {
...
+    return await this.sql.begin(async sql => {
+      // get block list
+      const { results: blocks, total: block_count } = await this.getBlocks({ limit, offset });
+      const blockHashValues = blocks
+        .map(block => `('\\x${block.index_block_hash.slice(2)}'::bytea)`)
```
zone117x, line 404: Parameter encoding for bytea types should already be handled automatically for 0x-prefix strings, see https://github.com/hirosystems/stacks-blockchain-api/blob/78d7d9c9f72db8ce6c59e0ea59a2579aceec014e/src/datastore/connection.ts#L24-L30

zone117x (changes requested):

src/datastore/pg-store.ts
```diff
@@ -396,6 +396,80 @@ export class PgStore {
...
+        .join(', ');
+      const microblocksQuery = await sql<(MicroblockQueryResult & { transaction_count: number })[]>`
+          SELECT ${sql(MICROBLOCK_COLUMNS)}, (
+            SELECT COUNT(tx_id)::integer as transaction_count
+            FROM txs
+            WHERE txs.microblock_hash = microblocks.microblock_hash
+            AND canonical = true AND microblock_canonical = true
+          )
+          FROM microblocks
+          WHERE parent_index_block_hash = ANY(VALUES ${sql.unsafe(indexBlockHashValues)})
```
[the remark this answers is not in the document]
MHassanTariq: I tested the query on 1000+ `microblocks` and `txs`. `IN` was more efficient, so I have used that instead.

src/datastore/pg-store.ts
```diff
@@ -396,6 +402,76 @@ export class PgStore {
...
   }
 
+  async getBlocksWithMetadata({ limit, offset }: { limit: number; offset: number }) {
+    return await this.sql.begin(async sql => {
+      // get block list
+      const { results: blocks, total: block_count } = await this.getBlocks({ limit, offset });
+      const blockHashValues: Buffer[] = [];
+      const indexBlockHashValues: Buffer[] = [];
+      blocks.forEach(block => {
+        const indexBytea = hexToBuffer(block.index_block_hash);
```
zone117x, line 412: Are these conversions necessary? Doesn't the new pg lib handle both hex strings and buffers automatically now?

MHassanTariq: It works with `IN` but not with `ANY` clause. I have removed the conversions since we're using `IN` clause anyways.

zone117x (approved):
Needs a rebase, otherwise lgtm. 
@rafaelcr could you also give this a review?

src/datastore/pg-store.ts
```diff
@@ -396,6 +402,76 @@ export class PgStore {
     });
   }
 
+  async getBlocksWithMetadata({ limit, offset }: { limit: number; offset: number }) {
```
rafaelcr, line 405: Add explicit return type and a brief JSdoc comment

src/datastore/pg-store.ts
```diff
@@ -396,6 +402,76 @@ export class PgStore {
...
+      blocks.forEach(block => {
+        const indexBytea = block.index_block_hash;
+        const parentBytea = block.parent_index_block_hash;
+        indexBlockHashValues.push(indexBytea, parentBytea);
+        blockHashValues.push(indexBytea);
+      });
+
+      // get txs in those blocks
+      const txQuery = await sql<ContractTxQueryResult[]>`
+        SELECT ${unsafeCols(sql, [...TX_COLUMNS, abiColumn()])}
```
rafaelcr, line 420: You're pulling a bunch of columns here (including the `abi` JOIN) but it seems you only need `tx_id`s in the response (and maybe `index_block_hash` for some comparisons).

You should update this `SELECT` to make this query much faster.

src/api/controllers/db-controller.ts
```diff
@@ -427,6 +427,20 @@ export async function getMicroblocksFromDataStore(args: {
   };
 }
 
+export async function getBlocksMetadata(args: { limit: number; offset: number; db: PgStore }) {
```
rafaelcr, line 430: Change name to `getBlocksWithMetadata`

rafaelcr (approved):
LGTM

Issue #1187, opened 2022-05-26
change token metadata inserts to upserts


rafaelcr:
currently, when inserting data into `ft_metadata` or `nft_metadata`, we use an `INSERT`. If we ever need to manually re-process a token by re-adding it to the queue in the database, we would be duplicating metadata entries which may cause response errors.

1. we should add unique constraints to both tables using the `contract_id`
2. we should add an `ON CONFLICT` clause based on these constraints to the `INSERT` query to turn it into an upsert

PR #1193, opened 2022-06-01
Upsert nft and ft metadata

MHassanTariq:
## Description

This PR updates the insert query for `nft` and `ft` metadata. New approach upserts on constraint for `contract_id`.

For details refer to issue #1187 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/migrations/1621511823381_nft_metadata.ts
+++ b/src/migrations/1621511823381_nft_metadata.ts
@@ -32,6 +32,7 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
     contract_id: {
       type: 'string', 
       notNull: true, 
+      unique: true,
     },
     tx_id: {
       type: 'bytea',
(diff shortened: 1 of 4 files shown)


src/migrations/1621511823381_nft_metadata.ts
```diff
@@ -32,6 +32,7 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
     contract_id: {
       type: 'string', 
       notNull: true, 
+      unique: true,
```
rafaelcr, line 35: Only keeping `contract_id` unique does not work for NFTs because NFTs have multiple metadata entries (one per individual token).
The unique constraint should be with both `contract_id` and the token id (not yet stored in the table, you'll need to add that)

rafaelcr (changes requested):

rafaelcr (approved):

rafaelcr:
Can you rebase so I can merge? @MHassanTariq

Issue #1152, opened 2022-04-28
Websocket `client.subscribeTxUpdates` always receives messages with `"tx_status":"pending"`


mcintyre94:
**Describe the bug**

When subscribing to transaction events using `client.subscribeTxUpdates` all messages received look like:

```
{
  "tx_id":"0x3d6e86ce5a3dc65da18f0007248ef26437dcd5ec7506278a695e815e779f0a51",
  "tx_status":"pending",
  "tx_type":"contract_call"
}
```

I can see in explorer that the first message received corresponds with the transaction being completed and in a microblock, and the second is when it's in an anchor block. But the status is always show as `pending`

Sometimes I don't receive a message at all when the status changes, but if I do it always has tx_status: pending

**To Reproduce**

- Go to https://tx-listen.mcintyre94.repl.co/
- Connect (to testnet) + submit the transaction
- In the console you'll see a message like "Subscribed to transaction 3d6e86ce5a3dc65da18f0007248ef26437dcd5ec7506278a695e815e779f0a51"
- Wait ~a few mins for the transaction to be in a microblock and then an anchor block. You'll see the 2 messages received on the websocket displayed in the UI. Both will have `"tx_status":"pending",`
- Confirm using explorer that the transaction is not in fact pending

Replit source code: https://replit.com/@mcintyre94/tx-listen#pages/index.jsx 

**Expected behavior**

- Receive websocket messages with a correct `tx_status` field

**Screenshots**

![Screenshot 2022-04-28 at 14 26 10](https://user-images.githubusercontent.com/1711350/165762729-3c8b4992-ac13-484a-947f-6602ae5a9cea.png)

https://user-images.githubusercontent.com/1711350/165763004-7822af60-2364-4571-b570-c8cdad9ae0b8.mov

**Console log**

```
Set stacks client!
{address: 'STMVVMSZ66AR57354N583KPP8ZBKQFS1R45S0WSE'}
tx created {txId: '3d6e86ce5a3dc65da18f0007248ef26437dcd5ec7506278a695e815e779f0a51', txRaw: '0x8080000000040029bdd33f3195829c65254a81ced647d73b…f4240051af9e265622609cada3555d7c83ba9c6973c17f69d', stacksTransaction: nS}
Subscribed to transaction 3d6e86ce5a3dc65da18f0007248ef26437dcd5ec7506278a695e815e779f0a51
[...]

Issue #1153, opened 2022-04-28
Socket IO: `subscribeTransaction` is never receiving a message


mcintyre94:
**Describe the bug**

I have a simple app that submits a tutorial and then subscribes to the transaction ID to track progress. When using the socket IO API this app never receives a message for the transaction after subscribing. 

**To Reproduce**

- Go to https://tx-listen-socketio.mcintyre94.repl.co/
- Connect a wallet (testnet) + submit the transaction
- Notice the mempool messages on the page, verifying that the socket IO is working
- Optional: click the unsubscribe from mempool button to stop more of those happening
- Check the transaction ID (logged in console) on explorer, wait for it to complete and be in a microblock/anchor block
- Notice that no transaction messages ever get logged

Replit source code: https://replit.com/@mcintyre94/tx-listen-socketio#pages/index.jsx

**Expected behavior**

The socket has a catch-all listener:

```
  socket.onAny((event, ...args) => {
    console.log('received event!', event)
    setMessages(m => [...m, JSON.stringify({event, args})])
  })
```

So I'd expect that transaction update messages would show up

I also see that on the websocket () the messages never show up:

```
0{"sid":"iOFsA8tyv3IfSybWAAHJ","upgrades":[],"pingInterval":25000,"pingTimeout":20000} 
40{"sid":"yAlmBx-gulqCSKR7AAHK"}
420["subscribe","mempool"]
430[null] |
[...]

PR #1197, opened 2022-06-06
fix: incorrect websocket/socket.io transaction updates

### LINKED ISSUE #957: return full transaction objects through websockets
The current websocket updates for transactions and balances include very little information about the actual event. For example: An address update only gives users the `tx_id`, `tx_status` and `tx_type`, while socket.io returns a full `AddressTransactionWithTransfers` object with all the details saving one extra API call from the user if they want more info (related: #830)

Since this would be a breaking change for current websocket users (and because #840 is now merged) we should look at metrics for current usage to see the possible impact.

rafaelcr:
This PR fixes issues #1152 and #1153 , but it also refactors our WebSockets pipeline to the following structure:

* Instead of maintaining independent socket.io and ws-rpc controllers (bug prone), `WebSocketTransmitter` was created to receive all blockchain events, query for all relevant data, and finally just relay info to clients via a standardized channel interface
* A `WebSocketChannel` interface was created so we can create as many channels as we need (currently only socket.io and ws-rpc)
* `WsRpcChannel` and `SocketIOChannel` were refactored from the old controllers to manage each channel's subscriptions and relay data from the transmitter to clients

This PR also:
* [x] fixes some mempool events that are not currently transmitted (like garbage collection drops)
* [x] standardizes ws-rpc responses to close #957


src/datastore/pg-write-store.ts
```diff
@@ -1772,6 +1772,9 @@ export class PgWriteStore extends PgStore {
     `;
     await this.refreshMaterializedView(sql, 'mempool_digest');
     const deletedTxs = deletedTxResults.map(r => r.tx_id);
+    for (const txId of deletedTxs) {
+      await this.notifier?.sendTx({ txId: txId });
```
zone117x, line 1776: I think this is being called while in a sql transaction from the calling function, right? If so, seems like the deleted txs should be placed in a variable outside of the sql transaction function, then call the notifier after sql tx finishes?

rafaelcr: Good point, done

zone117x (approved):
Nice abstraction 💯

PR #1214, opened 2022-06-16
feat: add average fee to block response

### LINKED ISSUE #1209: Add average tx fee to blocks
Hello team!
I'm working on a new feature to add historical blocks information to tx page, we need to display the previous average fee for these blocks:
image
Currently, this is only possible through fetching txs for each block and iterating over them on the client, which requires multiple requests. A solution would be to add a new attribute `average_fee` to block response.

He1DAr:
Adding a new attribute average_fee to block response, closes #1209


rafaelcr (changes requested):
Thank you @He1DAr this looks awesome... just a few comments

rafaelcr (commented):
Can you please rebase from `develop` @He1DAr ? Otherwise LGTM 🔥

rafaelcr:
@He1DAr there still seem to be some issues with lint and some broken tests, let me know if I can help you with anything here

src/api/controllers/db-controller.ts
```diff
@@ -1063,7 +1068,9 @@ export async function searchHashWithMetadata(
       blockQuery.result.txs.map(tx => tx.tx_id),
       blockQuery.result.microblocks.accepted.map(mb => mb.microblock_hash),
       blockQuery.result.microblocks.streamed.map(mb => mb.microblock_hash),
-      blockQuery.result.microblock_tx_count
+      blockQuery.result.microblock_tx_count,
+      blockQuery.result.txs.reduce((acc, tx) => acc + tx.fee_rate, BigInt(0)) /
```
zone117x, line 1072: I'm curious if a simple fee average like this is very useful. You could see a block of mostly low-cost stx transfers, then a block of mostly high-cost contract-call txs. A simple average could indicate wildly fluctuating values even if the network fee market didn't change.

A possible more useful alternative would be to include the total block `execution_cost_*` values as opposed to just tx count. Client's may want to calculate/display those costs in different ways, so just returning a `total_fees` for a block would be adequate.

kantai: Yep — using an average weighted by the execution cost vector would be the way to go. The default cost metric used by the mining implementation is a dot product of the execution vector with a scaling vector.

The implementation of this is here:

https://github.com/stacks-network/stacks-blockchain/blob/master/src/cost_estimates/metrics.rs#71

And here

https://github.com/stacks-network/stacks-blockchain/blob/master/clarity/src/vm/costs/mod.rs#L1065

Issue #1186, opened 2022-05-26
add API version string to every response through a header


rafaelcr:
for easier debugging when looking at responses (especially when we have canary enabled), we should return the API's version string in every response through a header like `Server` (currently returns `cloudflare`) or a custom one like `X-Stacks-API-Version`:
```
Server: stacks-blockchain-api v3.0.3 (master:cd0c8aef)
```

PR #1216, opened 2022-06-21
Stacks API version in response header

MHassanTariq:
## Description

This PR adds stacks API version in the response header as, 
`X-Stacks-API-Version : stacks-blockchain-api v3.0.3 (master:cd0c8aef)`

For details refer to issue #1186 

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


src/api/init.ts
```diff
@@ -128,6 +128,16 @@ export async function startApiServer(opts: {
     });
     app.use(promMiddleware);
   }
+  // adding stacks-api-version in header
+  app.use((_, res, next) => {
+    try {
+      const [branch, commit, tag] = fs.readFileSync('.git-info', 'utf-8').split('\n');
```
zone117x, line 134: Should definitely avoid reading this file on _every_ request. These variables can be read and stored in a variable outside of this function.

src/api/init.ts
```diff
@@ -128,6 +128,16 @@ export async function startApiServer(opts: {
...
+      res.setHeader('X-Stacks-API-Version', `stacks-blockchain-api ${tag} (${branch}:${commit})`);
```
zone117x, line 135: This header needs added to CORs `Access-Control-Expose-Headers` allowlist if we want web browser js to be able to access it.

src/api/init.ts
```diff
@@ -128,6 +137,14 @@ export async function startApiServer(opts: {
     });
     app.use(promMiddleware);
   }
+  // adding stacks-api-version in header
+  app.use((_, res, next) => {
+    if (stacksVersion) {
+      res.setHeader('X-Stacks-API-Version', stacksVersion);
```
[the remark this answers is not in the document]
CharlieC3: Internally, we're restricting the content added to the `server` header for additional security. If we apply the version to that header, I'm not sure if it'll make it back to the origin, or if it'll be filtered out. We'd have to test it to be sure.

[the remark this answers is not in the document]
rafaelcr: Ok thanks, no problem... in that case, using `X-Stacks-API-Version` sounds good to me

src/api/init.ts
```diff
@@ -128,6 +137,14 @@ export async function startApiServer(opts: {
...
+    }
+    res.setHeader('Access-Control-Expose-Headers', 'X-Stacks-API-Version');
```
rafaelcr, line 145: @MHassanTariq this would overwrite previous values for this header. Can you modify it so that it appends this new value?

zone117x (approved):
LGTM!

Issue #1217, opened 2022-06-22
unable to run arm64 docker image with latest master


rswol:
**Describe the bug**
While I can build ARM docker image using buildx:
```
docker buildx build -f Dockerfile --platform linux/arm64 -t stacks-blockchain-api --load .
```
while trying to run it fails:
```
docker run --rm -t phx.ocir.io/ax3gejcudsku/stacks-blockchain-api:deploy-prod         
WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module './native/linux-arm64-musl.node'
Require stack:
- /app/node_modules/stacks-encoding-native-js/loader.js
- /app/node_modules/stacks-encoding-native-js/index.js
- /app/lib/helpers.js
- /app/lib/index.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/app/node_modules/stacks-encoding-native-js/loader.js:6:18)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/app/node_modules/stacks-encoding-native-js/loader.js',
    '/app/node_modules/stacks-encoding-native-js/index.js',
    '/app/lib/helpers.js',
    '/app/lib/index.js'
  ]
}
```

Most likely the issue is in stacks-encoding-native-js package which incorrectly detects the platform (should be aarch64 ?)

**Expected behavior**

Build and run arm docker image

**Desktop (please complete the following information):**
 - OS: ubuntu/x86

PR #1220, opened 2022-06-23
fix: arm64-alpine image, bump stacks-encoding-native-js #1217

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1217

Bumps `stacks-encoding-native-js` to a version that includes a binary for the linux-musl-arm64 target. 

See https://github.com/hirosystems/stacks-encoding-native-js/issues/4 for more info.


rafaelcr (approved):
Nice

PR #1228, opened 2022-07-06
OpenAPI.yaml:  Added examples and description

LakshmiLavanyaKasturi:
API's are tested to add examples and description. 

User Story: https://github.com/hirosystems/docs/issues/101

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -578,7 +578,7 @@ paths:
           schema:
             type: integer
             default: 20
-            maximum: 200
+            maximum: 30
         - name: offset
           in: query
           description: index of first block to fetch
@@ -604,10 +604,12 @@ paths:
           type: string
     get:
       summary: Get block by hash
-      description: Retrieves block details of a specific block for a given chain height
+      description: Retrieves block details of a specific block for a given chain height.
       tags:
         - Blocks
       operationId: get_block_by_hash
+      example: hash=0xbf06705d0b8c2389eb4b3a19ce72096b7158b158e2cd4ceb27d8de9db19bb3be
+      description: Use the hash from your latest block. get_block_list API gets you the latest blocks and their hash values.
       responses:
         200:
           description: Block
@@ -1228,7 +1230,8 @@ paths:
       parameters:
         - name: principal
           in: path
-          description: Stacks address or a Contract identifier (e.g. `SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0.get-info`)
+          description: Stacks address or a Contract identifier.
+          example: principal=SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0
           required: true
           schema:
             type: string
@@ -1242,7 +1245,7 @@ paths:
             default: false
         - name: until_block
           in: query
-          description: returned data representing the state up until that point in time, rather than the current block. Note - Use one parameter at a time.
+          description: returned data representing the state up until that point in time, rather than the current block. Note - Use either of the parameters but no both.
           example: https://stacks-node-api.testnet.stacks.co/extended/v1/address/SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0/stx?until_block=string
           required: false
           schema:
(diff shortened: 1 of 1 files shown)


docs/openapi.yaml
```diff
@@ -1191,8 +1193,8 @@ paths:
       parameters:
         - name: principal
           in: path
-          description: Stacks address or a Contract identifier (e.g. `SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0.get-info`)
-          required: true
```
rafaelcr, line 1195: This should be kept because the path parameter is required, though, right?

docs/openapi.yaml
```diff
@@ -1228,20 +1230,23 @@ paths:
...
+          description: Stacks address or a Contract identifier.
+          example: principal=SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0
           required: true
           schema:
             type: string
         - name: unanchored
           in: query
-          description: Include transaction data from unanchored (i.e. unconfirmed) microblocks
+          description: Include transaction data from unanchored (i.e. unconfirmed) microblocks.
+          example: https://stacks-node-api.testnet.stacks.co/extended/v1/address/SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0/stx?unanchored
```
rafaelcr, line 1241: I think this example should just be `unanchored=true` to avoid including full paths that may change in the future (testnet vs. mainnet, domain changes, etc.)

docs/openapi.yaml
```diff
@@ -1268,7 +1273,8 @@ paths:
       parameters:
         - name: principal
           in: path
-          description: Stacks address or a Contract identifier (e.g. `SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0.get-info`)
+          description: Stacks address or a Contract identifier 
+          example: https://stacks-node-api.testnet.stacks.co/extended/v1/address/SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0/transactions
```
rafaelcr, line 1277: `principal=SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0`

docs/openapi.yaml
```diff
@@ -1299,7 +1308,7 @@ paths:
             default: false
         - name: until_block
           in: query
-          description: returned data representing the state up until that point in time, rather than the current block.
+          description: returned data representing the state up until that point in time, rather than the current block. Use either of unachored or until_block parameters.
```
rafaelcr, line 1311:
```suggestion
          description: returned data representing the state up until that point in time, rather than the current block. Use either of `unachored` or `until_block` parameters.
```

docs/openapi.yaml
```diff
@@ -604,10 +604,12 @@ paths:
           type: string
     get:
       summary: Get block by hash
-      description: Retrieves block details of a specific block for a given chain height
+      description: Retrieves block details of a specific block for a given chain height.
       tags:
         - Blocks
       operationId: get_block_by_hash
+      example: hash=0xbf06705d0b8c2389eb4b3a19ce72096b7158b158e2cd4ceb27d8de9db19bb3be
+      description: Use the hash from your latest block. get_block_list API gets you the latest blocks and their hash values.
```
rafaelcr, line 612:
```suggestion
      description: Use the hash from your latest block. `get_block_list` API gets you the latest blocks and their hash values.
```

rafaelcr (changes requested):
This is looking great, thanks @LakshmiLavanyaKasturi 
Just a few style comments

docs/openapi.yaml
LakshmiLavanyaKasturi: Yes, deleted by mistake. Added back.

docs/openapi.yaml
LakshmiLavanyaKasturi: Got it. I will change this.

docs/openapi.yaml
```diff
@@ -599,12 +603,13 @@ paths:
       - name: hash
         in: path
         description: Hash of the block
+        example: hash=0xbf06705d0b8c2389eb4b3a19ce72096b7158b158e2cd4ceb27d8de9db19bb3be
```
zone117x, line 606: IIUC, `example` should contain literals which are syntactically valid https://swagger.io/specification/#example-object, so in this case it would just be `0xbf06705d0b8c2389eb4b3a19ce72096b7158b158e2cd4ceb27d8de9db19bb3be`.

E.g. some tooling like OpenAPI front-ends that have "try me" functionality will often use the `example` as a default value. 

There's a few other places where it looks like this change is also needed.

LakshmiLavanyaKasturi:
Closing this PR as a duplicate of https://github.com/hirosystems/stacks-blockchain-api/pull/1232. 

Reason: Created this PR by cloning the fork of stacks-blockchain-API repo which doesn't work.

PR #1229, opened 2022-07-06
feat: add cache control to `/extended/v1/tx/:tx_id`

### LINKED ISSUE #1225: add cache headers to `/extended/v1/tx/:tx_id`
We currently don't add cache headers to `/extended/v1/tx/:tx_id` because its results may come from either the canonical chain or the mempool, and we don't have a cache handler that can deal with that variable scenario. This is causing that endpoint to produce spikes of ~10K uncached requests every ~10 minutes as some applications/marketplaces may be polling for some transactions continuously.

We should add an "adaptive" cache handler that would choose the correct ETag depending if the result came from the mempool or the canonical chain, and install it in that endpoint.

rafaelcr:
Adds `ETag` header to `/extended/v1/tx/:tx_id` and `/extended/v1/tx/:tx_id/raw`.

Since transactions can be in the mempool or in the chain and they can later get re-orged, a new `ETagType.transaction` was added which builds the ETag using the `tx_id`, `index_block_hash` (if any), `microblock_hash` (if any), and `status` of the relevant transaction. This way, we'll get the best caching behavior even when the tx may be moved from block to microblock to mempool and back.

Closes #1225

### DIFF
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -4069,7 +4069,7 @@ export class PgDataStore
       const mempoolResult = await client.query<{ status: number }>(
         `SELECT status
         FROM mempool_txs
-        WHERE tx_id = $1 AND canonical = TRUE AND microblock_canonical = TRUE`,
+        WHERE tx_id = $1`,
         [hexToBuffer(txId)]
       );
       if (mempoolResult.rowCount > 0) {
--- a/src/datastore/postgres-store.ts
+++ b/src/datastore/postgres-store.ts
@@ -103,6 +103,8 @@ import {
   TransactionType,
   AddressUnlockSchedule,
   Block,
+  MempoolTransactionStatus,
+  TransactionStatus,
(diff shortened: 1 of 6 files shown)


src/datastore/postgres-store.ts
```diff
@@ -4053,6 +4056,44 @@ export class PgDataStore
     });
   }
 
+  async getTxStatus(txId: string): Promise<FoundOrNot<DbTxGlobalStatus>> {
+    return this.queryTx(async client => {
+      const chainResult = await client.query<DbTxGlobalStatus>(
+        `SELECT status, index_block_hash, microblock_hash
+        FROM txs
+        WHERE tx_id = $1 AND canonical = TRUE AND microblock_canonical = TRUE`,
```
zone117x, line 4064: Is there anything to gain if we add a `LIMIT 1` here and in the the mempool query?

zone117x (approved):
Nice optimization!

PR #1253, opened 2022-08-02
fix: include `index_block_hash` in `/block`

### LINKED ISSUE #1238: Include `index_block_hash` in the /block endpoint
**Describe the bug**

When using the `tip` parameter on the read-only function call (https://hirosystems.github.io/stacks-blockchain-api/#tag/Smart-Contracts/operation/call_read_only_function), it is not clear which tip (hash) it expects. For example, the hash `ab61ba097c33eba57863249d82ce4e082e33d6b8c72f9279ae175536ed9da93c` on https://stacks-node-api.mainnet.stacks.co/extended/v1/block/0xab61ba097c33eba57863249d82ce4e082e33d6b8c72f9279ae175536ed9da93c returns `Chain tip not found`.

With some help on Discord, we figured out it expects `index_block_hash`, but this is not exposed anywhere through the API.

**To Reproduce**

```
curl "https://stacks-node-api.mainnet.stacks.co/v2/contracts/call-read/SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR/arkadiko-governance-v3-1/get-proposals?tip=ab61ba097c33eba57863249d82ce4e082e33d6b8c72f9279ae175536ed9da93c" \
  -H 'accept: */*' \
  -H 'accept-language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -H 'pragma: no-cache' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: cross-site' \
  -H 'sec-gpc: 1' \
  --data-raw '{"sender":"SPX8T06E8FJQ33CX8YVR9CC6D9DSTF6JE0Y8R7DS","arguments":[]}' \
  --compressed
```

The above returns `Chain tip not found`. Without the tip parameter the call works (but we need to query in the past). I suppose passing the `index_block_hash` would work, but it's not clear where this hash can be found.

Hope that helps!

rafaelcr:
Closes #1238


zone117x (approved):
Awesome, thanks @rafaelcr!

Issue #1200, opened 2022-06-07
Import events failed 


friedger:
**Describe the bug**
Import of events fails with internal error

**To Reproduce**
1. run api node on 3.+
2. run upgrade using stacks-blockchain-docker
3. see error as shown below

**Expected behavior**
api node should close gracefully and provide information how to fix the issue OR better continue

**Console log**
```
stacks-blockchain-api    | {"level":"warn","message":"Could not find tx index for subdomain entry: {\"name\":\"id.stx\",\"namespace_id\":\"stx\",\"fully_qualified_subdomain\":\"niko_nie.id.stx\",\"owner\":\"SP3J34ASN2SDA641KT9MD4Y84SJ0P2FBY6HM89RPG\",\"zonefile_hash\":\"27c87cc6331bc588d21a911238aad4a9\",\"zonefile\":\"$ORIGIN niko_nie.id.stx\\n$TTL 3600\\n_http._tcp\\tIN\\tURI\\t10\\t1\\t\\\"https://gaia.blockstack.org/hub/1MobAXTm9b5wkfoBWEVVSxa5hkhiJhmyEW/profile.json\\\"\\n\\n\",\"tx_id\":\"0x2cd0c1d8be0dbdd1f5af45c30c172b91eb1e3fd2fa226dfa3d3f9bbf7a4b8be2\",\"tx_index\":-1,\"canonical\":true,\"parent_zonefile_hash\":\"90c356d9444bd6ed21274df21798a9a910003035\",\"parent_zonefile_index\":0,\"block_height\":56291,\"zonefile_offset\":1,\"resolver\":\"https://registrar.stacks.co\"}","timestamp":"2022-06-07T14:21:33.013Z"}
[...]

PR #1263, opened 2022-08-10
Update docs to include a simple sql query

wileyj:
Adds an additional line to the markdown docs for deploying the API:
```
3. Verify data is being written to the database: `select * from blocks limit 1;`
```

During the step to verify the DB is setup, running this query will confirm data is being written to the DB.

in the case where the blockchain was syncing from genesis *without* the event observer, then restarted with the event observer, the previous commands will make it appear as though the DB is fine, but not adding records. this single query will confirm if the API is running normally.


wileyj:
https://github.com/hirosystems/stacks-blockchain-api/issues/1200

rafaelcr (approved):

PR #1264, opened 2022-08-11
fix: event-replay log level configurable via env var

zone117x:
Added a new env var `STACKS_API_EVENT_REPLAY_LOG_LEVEL` which can be used to specify the log level used during event-replay. 

If unspecified, it defaults to the original log level of `warn`.


rafaelcr (approved):
LGTM, thanks

Issue #1204, opened 2022-06-13
Deadlocks in v4


CharlieC3:
**Describe the bug**
We're seeing a high amount of deadlocks being caused in Postgres since upgrading to v4.
![Screen_Shot_2022-05-27_at_11 15 45_AM](https://user-images.githubusercontent.com/2747302/173442108-4768c800-1037-4c79-8242-ac30253b82d1.png)

This is intermittently causing microblock processing errors to occur:
```
error: deadlock detected
    at Parser.parseErrorMessage (/app/node_modules/pg-protocol/src/parser.ts:369:69)
    at Parser.handlePacket (/app/node_modules/pg-protocol/src/parser.ts:188:21)
    at Parser.parse (/app/node_modules/pg-protocol/src/parser.ts:103:30)
    at Socket.<anonymous> (/app/node_modules/pg-protocol/src/index.ts:7:48)
    at Socket.emit (node:events:527:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23) 
```

```


 
--
Process 1774416 waits for AccessExclusiveLock on relation 1035814 of database 16384; blocked by process 1774556. Process 1774556 waits for AccessShareLock on relation 1035792 of database 16384; blocked by process 1774416.
```

and increasing response times during a deadlock.

CharlieC3:
@rafaelcr @zone117x 
Here are some logs I found from the PG DB which may help:
```
2022-06-14 17:45:33.026 GMT [2577263] ERROR:  deadlock detected at character 36
2022-06-14 17:45:33.026 GMT [2577263] DETAIL:  Process 2577263 waits for AccessShareLock on relation 1035784 of database 16384; blocked by process 2576913.
	Process 2576913 waits for AccessExclusiveLock on relation 1035814 of database 16384; blocked by process 2577263.
	Process 2577263:
	        SELECT value
	        FROM nft_custody
	        WHERE recipient <> $1 AND value = ANY($2)

	Process 2576913: REFRESH MATERIALIZED VIEW chain_tip
2022-06-14 17:45:33.026 GMT [2577263] HINT:  See server log for query details.
2022-06-14 17:45:33.026 GMT [2577263] STATEMENT:
	        SELECT value
	        FROM nft_custody
[...]

PR #1267, opened 2022-08-12
fix: use primary db for socket queries

rafaelcr:
temporary fix to try to avoid deadlocks #1204


zone117x (approved):
I don't think this by itself will fix the deadlocks for replica-related reasons we've discussed. But certainly willing to test it out.

PR #1280, opened 2022-08-17
fix: update several test suites to work with Stacks 2.1 docker image

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1281

Update the dev & test environment to use Stacks 2.1 (that is, the latest `next` branch of the stacks-node). 

Most test suites expect the stacks-node RPC endpoints to be responsive while event-broadcasts are in progress (and not retrying while an event-observer is unavailable). This behavior is no longer possible In the latest code for the stacks-node mocknet. So test startup logic and stacks-node RPC dependencies have to be cleaned up. 

Progress:

* `tests` (the default, regular unit test suite)
  * [x] Tests pass locally
  * [x] Tests pass in CI
  * [x] fix the RPC client tests which have been disabled (likely need to be moved into their own test suite)
  * [x] move tbtc faucet tests into own suite (only these tests require bitcoind container init)
* `test-bns`
  * [x] Tests pass locally
  * [x] Tests pass in CI
* `test-rosetta`
  * [x] Tests pass locally
  * [x] Tests pass in CI
  * [ ] TODO: potentially move Rosetta `network/*` endpoint tests to their own suite -- they need a stacks-node RPC connection, but aren't strictly related to the `construction` tests
* `test-rosetta-cli-data`
  * [ ] Tests pass locally
  * [x] Tests pass in CI
* `test-rosetta-cli-construction`
  * [ ] Tests pass locally
  * [x] Tests pass in CI
* `test-tokens`
  * [ ] Tests pass locally
  * [x] Tests pass in CI


.github/workflows/ci.yml
```diff
@@ -168,6 +168,65 @@ jobs:
...
+      - name: Setup env vars
+        run: echo "STACKS_CORE_EVENT_HOST=http://0.0.0.0" >> $GITHUB_ENV
+
+      - name: Setup integration environment
+        run: |
+          sudo ufw disable
+          npm run devenv:deploy -- -d
+          npm run devenv:logs -- --no-color &> docker-compose-logs.txt &
+
+      - name: Run bns tests
```
rafaelcr, line 207:
```suggestion
      - name: Run RPC tests
```

rafaelcr (commented):
Looks good so far

Since we're spinning off Rosetta into a microservice relatively soon, I think you shouldn't spend too much time on fixing those tests other than just making them work for now. IMO some of those suites need total refactoring (specially the `cli` ones that spin up the docker containers from within the test instead of from the outside) and we'll have to do that when making the spin-off anyway.

zone117x:
@rafaelcr this PR should be good to go now.

Issue #1273, opened 2022-08-16
zonefiles do not track canonical chain state correctly


rafaelcr:
The `zonefiles` table has been returning incorrect results for BNS endpoint queries given that it does not keep the canonical state of the related transaction correctly.

There are lots of instances when an `/attachments/new` message from the node comes which includes two (or more) updates for the same zonefile but only one of them is actually canonical and valid. As it stands today, the zonefiles table is just UPDATEd with the last received value rather than the correct value.

Another related issue is that no validation is being done to prevent inserting a zonefile with the same hash more than once in the table, making it huge and very difficult to navigate around in event replays.

Issue #1269, opened 2022-08-16
/v1/names/wowscience.id.stx is not returned by the API


pradel:
**Describe the bug**

When accessing `https://stacks-node-api.mainnet.stacks.co/v1/names/wowscience.id.stx`, the API is not returning the info regarding this username, instead, it's redirecting to `https://registrar.stacks.co/v1/names/wowscience.id.stx` to show the data included in the registrar.
This would be the expected behavior for a pending subdomain not submitted to the blockchain. However, the tx registering this username was included in an anchor block 10 days ago, see https://explorer.stacks.co/txid/2a0939c461c9caff0a91c5b66d0290dd0d16153f9b3c96edb3cc000e88e77cfc?chain=mainnet. The API should have the info and return it.
The result is that asking the names that [SP12YGP1C4V0XRSHDW9G2PSEZRK1T15M66K1GKMBR](https://stacks-node-api.mainnet.stacks.co/v1/addresses/stacks/SP12YGP1C4V0XRSHDW9G2PSEZRK1T15M66K1GKMBR) is holding return an empty array.
The user do not see the username in the Hiro wallet and can't connect to Sigle at the moment.

**To Reproduce**
1. Go to https://stacks-node-api.mainnet.stacks.co/v1/names/wowscience.id.stx
2. See that you are being redirected to https://registrar.stacks.co/v1/names/wowscience.id.stx
3. In the JSON response, get the txid and open it in the stacks explorer to see that the tx was mined

**Expected behavior**

The API should return the information directly as it's included in a block and not rely on the registrar.

pradel:
I just checked and the endpoint returning the names that this address is holding is returning an empty array https://stacks-node-api.mainnet.stacks.co/v1/addresses/stacks/SP12YGP1C4V0XRSHDW9G2PSEZRK1T15M66K1GKMBR

rafaelcr:
I originally thought this issue was only a bug related to canonical data, however, upon further inspection it looks like in some of our instances we are not receiving an `/attachments/new` message from the node which contains the updated zonefile we need to process in order to detect this new subdomain.

We're currently investigating whether this is a bug on the API side or the stacks node side.

rafaelcr:
Blocked by this issue on the Stacks blockchain: https://github.com/stacks-network/stacks-blockchain/issues/3270

pradel:
Hey there, any update? Our user is still blocked and can't use his account.

rafaelcr:
[...]

Issue #1176, opened 2022-05-17
BNS name renewal ->new expired_block value


sleepiJoe:
**Describe the bug**
Wrong data for `expire_block` of [this endpoint](https://docs.hiro.so/api#operation/get_name_info).

**To Reproduce**
1. Go to ['the endpoint'](https://stacks-node-api.mainnet.stacks.co/v1/names/friedger.id)
2. Review line `"expire_block": 52595,`
3. See error

**Expected behavior**
BNS name renewal/new expired_block value should be visible at this endpoint.

friedger:
The name was renewed at block 51951
https://explorer.stacks.co/txid/0xf037c8da8210e2a348bbecd3bc44901de875d3774c5fce49cb75d95f2dc2ca4d?chain=mainnet

rafaelcr:
looping in @lgalabru who might have some more context here...

Ludo, do you know if we get an event when a name renewal happens? Either an event emitted by the BNS contract or a new attachment from the Stacks node event emitter?

I looked directly at our postgres database to see which events we currently have in the `names` table for `friedger.id` and we only have the name registration, not the renewal. That's why it's returning stale data. Maybe the API is not listening for renewals correctly.

friedger:
For name-update txs at least the attachments are treated.

asimm241:
I have debugged the issue and here are my findings:

1) We haven't received the event for this particular name for 'name-renewal'. even though we should receive 'name-renewal' events. We have a test case for this.  The evidence for that is `last_txid` is not updated and `expire_block` is the same as It was in v1-import data. 

2) We don't recieve `expire_block` value from the node. We have to calculate it. It should calculated as following
`expire_block` = (block at which name `import`/`registered`/`renewed`) + `lifetime` 
`lifetime` is the property of corresponding namespace, in this case it's `id`.  So, we have to calculated after we recieve the event.

3) We don't have mechanism for calculating expire_block. We should make a decision if we shold  **store expire_block** in names table or **calculate expire_block runtime** when is queried for the name by joining with namespace table. 
1st option is easy to implment
2nd Is a little tricky because we don't recieve namespace infromation for name-import so probably have to make contract call to fetch namespace lifetime but not sure.

asimm241:
[...]

PR #1287, opened 2022-08-19
feat!: optimize BNS database tables

rafaelcr:
This PR deals with a few issues concerning BNS namespaces, names, subdomains and zonefiles:

* Tracks canonical chain state correctly for zonefiles, preventing duplicate entries and guaranteeing correct responses for endpoints #1273 #1269 
* Enforces chain uniqueness for all 4 BNS tables, reducing table sizes
* Adds chain tip ETag cache for all BNS API endpoints
* Bundles attachment updates into batched queries
* Expands BNS tests to validate v1 import completely

There is a caveat to this update, however. Since we're now using UNIQUE constraints to maintain table integrity, we cannot disable indexes for v1 imports anymore. This will impact v1 import times, but since the total number of indexes was reduced dramatically for tables, it should not be that much of a time increase (testing TBD).

TODO:
- [ ] Test full BNS v1 import
- [ ] Test archival event replay
- [ ] Fix expired_block values for names #1176


zone117x:
> Since we're now using UNIQUE constraints to maintain table integrity, we cannot disable indexes for v1 imports anymore. This will impact v1 import times

I _think_ the v1 names dataset is already unique, so it should still be possible to insert without indexes (although perhaps before the full tsv import), right?

src/datastore/postgres-store.ts
```diff
@@ -2159,6 +2148,94 @@ export class PgDataStore
...
+                resolver: zoneFileContents.uri ? parseResolver(zoneFileContents.uri) : '',
+              };
+              subdomains.push(subdomain);
+            }
+            subdomainData.push({ blockData, subdomains, attachment });
+          }
+        }
+        await this.updateBatchSubdomainsAndZonefiles(client, subdomainData);
+        // Update main zonefile as well.
+        await this.updateBatchSubdomainsAndZonefiles(client, [{ attachment }]);
```
zone117x, line 2231: I'm a bit confused at what's going on here. It looks like the `updateBatchSubdomainsAndZonefiles` function is a bit overloaded and maybe performing too many different things?

src/datastore/postgres-store.ts
```diff
@@ -6816,7 +6905,20 @@ export class PgDataStore
           tx_index, tx_id, status, canonical,
           index_block_hash, parent_index_block_hash, microblock_hash, microblock_sequence, microblock_canonical
         ) values($1, $2, $3, $4, $5, $6, $7, $8,$9, $10, $11, $12, $13, $14, $15)
-        `,
+        ON CONFLICT ON CONSTRAINT unique_name_tx_id_index_block_hash_microblock_hash DO
```
zone117x, line 6908: Could you provide some context around the condition here? Much of the BNS code is just simple data transforms and I'm having a hard time following the changes to the actual logic, but it seems like these unique constraints are the main ones?

So in this case, what are the conditions around a new name event containing a duplicate `(txid, index_block_hash, microblock_hash)`?

src/migrations/1608030374841_namespaces.ts
```diff
@@ -91,7 +91,14 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
     },
   });
 
-  pgm.createIndex('namespaces', 'index_block_hash', { method: 'hash' });
-  pgm.createIndex('namespaces', 'microblock_hash', { method: 'hash' });
-  pgm.createIndex('namespaces', [{ name: 'ready_block', sort: 'DESC' }]);
+  pgm.createIndex('namespaces', [
```
zone117x, line 94: This PR is intended to be a v4.x release right? If so, need to move these changes into new migration files.

But also, if this is for v4.x, will the changes fix the already-ingested BNS names, or only fix new BNS names? If only new ones, then is there a good reason to try to get this into v4 instead of develop/v5?

src/datastore/postgres-store.ts
zone117x: It would be super helpful to have some tests that replicate the behavior that this is solving

zone117x (commented):
PR looking great, and high impact issues to fix 👍 

I'm having a hard time following some of the logic (not really the fault of this PR, the BNS code is generally difficult to follow). Unit tests showing the kind of event payload patterns that this PR fixes would be super helpful.

Also left a comment regarding the viability of this being a v4.x change.

src/datastore/postgres-store.ts
rafaelcr: Good point, I was carrying the pattern from the old code but I can definitely split it up

src/datastore/postgres-store.ts
rafaelcr: In practice, there are two main cases in which we get duplicate names/subdomains/zonefiles:

1. **When we get a `name-register` followed by a `name-update` BNS event**
    The registration sends an attachment with an empty zonefile, and the update sends an attachment with the real zonefile usually a few blocks after. In the old code, each of these events inserted duplicate values because the attachments referenced the same tx_id/index_block_hash/microblock_hash instead of just updating the temporary registration rows (as it should be). This translates to literally thousands of duplicate and/or garbage rows in all three tables. When the old code DID update, though, it incorrectly updated based on the zonefile_hash which will *obviously* change when the name-update event gets sent (the zonefile changed, so the hash changed).
2. **When we get duplicate attachments**
    In the last replays we've done, I've seen attachment messages that contain duplicate entries for name updates (possibly from re-orgs or other node network issues). Today we process those as-is, inserting duplicates of everything.

> It would be super helpful to have some tests that replicate the behavior that this is solving

I'll definitely add some tests for this

src/migrations/1608030374841_namespaces.ts
rafaelcr: The idea for this PR is for it to become v5 by itself... a completely new replay is required. Develop would become v6. Wdyt?

zone117x: Sounds good to me!

src/tests-bns/event-server-tests.ts
```diff
@@ -128,6 +129,104 @@ describe('BNS event server tests', () => {
     expect(namespace.result?.ready_block).toBe(2);
   });
 
+  test('/attachments/new with re-orged zonefiles', async () => {
```
zone117x, line 132: Ty for the reorg test 👍

zone117x (approved):
Excellent PR 💯 

Some tests around the duplicate name behaviors you described would be really helpful. Otherwise, LGTM!

Issue #1294, opened 2022-08-24
[Stacks 2.1] Update stacks.js to v4


zone117x:
The API currently uses `@stacks/transactions` v2.0.1.

An upgrade to v4.4.x is necessary In order to construct `versioned-smart-contract` txs, and likely required to perform PoX-2 related operations in the Rosetta layer.

Specifically [4.4.0-stacks2.1-alpha.0](https://www.npmjs.com/package/@stacks/transactions/v/4.4.0-stacks2.1-alpha.0) which contains https://github.com/hirosystems/stacks.js/pull/1341.

zone117x:
Closed by https://github.com/hirosystems/stacks-blockchain-api/pull/1295

PR #1295, opened 2022-08-25
chore: upgrade to stacks.js v4.x

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1294

Couple changes of note:

We were able to get rid of the `bn.js` dependency 🎉

The debug/playground endpoint `/extended/v1/debug/broadcast/contract-deploy` was updated to support Stacks 2.1 `versioned-smart-contract` txs so that new Clarity-2 features can be used.

There were several instances, mostly in the Rosetta tests, where the fee was not specified and was left up to stacks.js to calculate it. This worked with the older version of stacks.js, because the original fee calculation was essentially `fee = someConstant * tx.byteLength`. The newer versions of stacks.js use the new `/v2/fees/transaction` fee estimator endpoint, which can return a `NoEstimateAvailable` response and cause the `await makeTx(...)` tx builder calls to throw. This happens often in test environments because the mocknet node doesn't usually have tx fee history available to come up with a response. 

This PR doesn't change the current fee behavior of the production Rosetta construction code, which currently still uses the legacy fee calculation method. This should probably be updated to use the new fee estimator endpoint (and fallback to the legacy one when `NoEstimateAvailable`). But that isn't a change related to Stacks 2.1, so I opened a new issue for tracking that change: https://github.com/hirosystems/stacks-blockchain-api/issues/1298


src/tests-tokens/tokens-metadata-tests.ts
```diff
@@ -98,16 +97,17 @@ describe('api tests', () => {
...
-    const feeRate = new BN(feeRateResult);
-    const fee = feeRate.mul(txBytes);
-    contractDeployTx.setFee(fee);
+    // const feeRateReq = await fetch(stacksNetwork.getTransferFeeEstimateApiUrl());
+    // const feeRateResult = await feeRateReq.text();
+    // const txBytes = BigInt(contractDeployTx.serialize().byteLength);
+    // const feeRate = BigInt(feeRateResult);
+    // const fee = feeRate * txBytes;
+    // contractDeployTx.setFee(fee);
     const { txId } = await sendCoreTx(contractDeployTx.serialize());
```
zone117x, line 111: The commented out code above never seemed to work as intended (including in v4). It attempted to set the fee _after_ the tx was already signed, which would result in the stacks-node rejecting the transaction because of signature mismatch.

Which means either A) the `.setFee(fee)` function wasn't actually changing the tx data, or B) it happened to be setting the fee to the same value already calculated by stacks.js during the `await makeContractDeploy(...)` call.

rafaelcr (approved):
LGTM 👍

PR #1301, opened 2022-08-30
fix: import BNS v1 data during event replay

rafaelcr:
* Import BNS v1 data during replay: first names and namespaces, then TSV data, then subdomains
* Obtain genesis block information from TSV to be used in BNS import
* Modify BNS insertion data to mark all imported names at block 1 instead of block 0

### DIFF
--- a/src/event-replay/helpers.ts
+++ b/src/event-replay/helpers.ts
@@ -1,5 +1,14 @@
+import * as fs from 'fs';
+import * as readline from 'readline';
+import { decodeTransaction, TxPayloadTypeID } from 'stacks-encoding-native-js';
+import { DataStoreBnsBlockData } from '../datastore/common';
 import { PgDataStore } from '../datastore/postgres-store';
 import { ReverseFileStream } from './reverse-file-stream';
+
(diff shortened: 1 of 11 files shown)


src/event-replay/helpers.ts
```diff
@@ -26,6 +35,56 @@ export async function findTsvBlockHeight(filePath: string): Promise<number> {
...
+ * @returns Genesis block data
+ */
+export async function findBnsGenesisBlockData(filePath: string): Promise<BnsGenesisBlock> {
+  const rl = readline.createInterface({
+    input: fs.createReadStream(filePath),
+    crlfDelay: Infinity,
+  });
+  let genesisBlock: BnsGenesisBlock | undefined;
+  return new Promise<BnsGenesisBlock>((resolve, reject) => {
+    rl.on('line', line => {
```
zone117x, line 51: How's memory usage with this? It's probably alright since there's no async code in the callback, if not, safer to handle stream backpressure, e.g. with an async iterator `for await (const line of rl) { ... }`

zone117x (approved):
Left a question, otherwise LGTM!

PR #1303, opened 2022-08-31
fix: detect name transfers and renewals in special circumstances

### LINKED ISSUE #1236: `/v1/names/{name}` does not reflect buying usernames from bids on marketplaces
**Describe the bug**

I am not sure if this is a bug or intended behaviour. Creating the issue to open discussion regarding the behavior of the `/v1/names/{name}` endpoint.

Problem: When buying a new .btc domain via a bid from a marketplace, the `/v1/names/{name}` seems to return outdated data.

The https://stacks-node-api.mainnet.stacks.co/v1/names/dayslikewater.btc endpoint, return the `last_txid` field with the transaction that registered the name. In the payload, we can see that `SP1TY00PDWJVNVEX7H7KJGS2K2YXHTQMY8C0G1NVP` is the current owner of this name (which is correct), but the `last_txid` [explorer link](https://explorer.stacks.co/txid/0x1316b570d85db81792b7fc18e7d7ec502ec886008e9b9be1f4bf891b4ab8380a?chain=mainnet) point to a transaction made by `SPP117ENNNDQVQ1G3E0N1AP178GXBTC2YNQ3H7J`. `SPP117ENNNDQVQ1G3E0N1AP178GXBTC2YNQ3H7J` is the account that bought the username at the beginning and then sold it to `SP1TY00PDWJVNVEX7H7KJGS2K2YXHTQMY8C0G1NVP`.

Shouldn't this endpoint reference the bid transaction in the `last_txid` result as the owner changed in this process?

**Expected behavior**

The https://stacks-node-api.mainnet.stacks.co/v1/names/dayslikewater.btc endpoint should return the last `last_txid` when the name was transferred to the new account.

rafaelcr:
* Allow name transfers triggered by contracts other than BNS
* Parse `name-renewal` contract calls that contained no zonefile hash (so no associated print event), related to #1176 
* Handle name expiration blocks and zonefiles correctly on transfers
* On transfers, calculate and save the new owner principal based on the name NFT event contained in the same transaction
* Add more tests for these specific cases

Fixes #1236

### DIFF
--- a/src/event-stream/event-server.ts
+++ b/src/event-stream/event-server.ts
@@ -63,7 +63,11 @@ import {
 } from 'stacks-encoding-native-js';
 import { ChainID } from '@stacks/transactions';
 import { BnsContractIdentifier } from './bns/bns-constants';
-import { parseNameFromContractEvent, parseNamespaceFromContractEvent } from './bns/bns-helpers';
+import {
+  parseNameFromContractEvent,
+  parseNameRenewalWithNoZonefileHashFromContractCall,
+  parseNamespaceFromContractEvent,
+} from './bns/bns-helpers';
 
 async function handleRawEventRequest(
   eventPath: string,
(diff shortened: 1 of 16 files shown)


src/event-stream/event-server.ts
```diff
@@ -325,16 +335,29 @@ function parseDataStoreTxEventData(
...
+          source_code: tx.parsed_tx.payload.code_body,
+          abi: JSON.stringify(tx.core_tx.contract_abi),
+          canonical: true,
+        });
+        break;
+      case TxPayloadTypeID.ContractCall:
+        // Name renewals can happen without a zonefile_hash. In that case, the BNS contract does NOT
+        // emit a `name-renewal` contract log, causing us to miss this event. This function catches
+        // those cases.
+        const name = parseNameRenewalWithNoZonefileHashFromContractCall(tx, chainId);
```
zone117x, line 354: To be clear, this is a "better than nothing" work-around right? As in, if a name-renewal was called from another contract then we wouldn't get here right?

rafaelcr: That's right, unless the other contract also did not include the zonefile hash when calling `name-renewal`.

PR #1318, opened 2022-09-16
feat: transactions count endpoint

He1DAr:
## Description

Adding a new endpoint `extended/v1/tx/count?from={FROM_TS}&to={TO_TS}` to get transactions count within a date range grouped by day. This allows us to display a line chart for transactions count.

Example:
GET `{API_URL}/extended/v1/tx/count?from=0&to=1663364033`
response:
```
{
  results: [
    {
      day: "2022-09-15T00:00:00.000Z",
      count: 122
    },
    {
      day: "2022-09-16T00:00:00.000Z",
      count: 613
    }
  ]
}
```

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other


src/datastore/postgres-store.ts
```diff
@@ -4230,6 +4231,28 @@ export class PgDataStore
     });
   }
 
+  async getTxsCount({ from, to }: { from: number; to: number }) {
+    let resultQuery: QueryResult<TransactionsCount>;
+    return this.queryTx(async client => {
+      const maxHeight = await this.getMaxBlockHeight(client, { includeUnanchored: true });
+      resultQuery = await client.query<TransactionsCount>(
+        `
+          SELECT DATE_TRUNC('day', to_timestamp(parent_burn_block_time)) as "day", count(*)::int as count
```
rafaelcr, line 4240: I think we'll need to run some performance tests on this query given that `COUNT(*)` is a very expensive operation and the `txs` table is the largest in the DB (currently at ~2.5 million rows IIRC). Maybe keeping a materialized view with these counts would make sense.

src/api/routes/tx.ts
```diff
@@ -286,6 +287,27 @@ export function createTxRouter(db: DataStore): express.Router {
     })
   );
 
+  router.get(
+    '/count',
+    cacheHandler,
+    asyncHandler(async (req, res) => {
+      const from = parseTimestampQueryInput(req.query.from);
+      const to = parseTimestampQueryInput(req.query.to);
```
rafaelcr, line 295: Would it make better sense to ask for YYYY-MM-DD dates rather than timestamps? With these args, you'd get incomplete day counts on the boundaries that you requested.

src/datastore/postgres-store.ts
```diff
@@ -4230,6 +4231,28 @@ export class PgDataStore
...
+          FROM txs
+          WHERE canonical = true AND 
+          microblock_canonical = true AND 
+          block_height <= $1 AND 
+          parent_burn_block_time >= $2 AND 
+          parent_burn_block_time <= $3
```
rafaelcr, line 4246: These dates are all stored in UTC. Will you need to translate dates to user's local times for explorer display?

rafaelcr (commented):
@He1DAr thanks for the PR! Could you give us some more detail on what you're trying to build?

Could you switch the target branch from `master` to `develop`? Unfortunately, you'll need to rewrite some of the SQL code given that develop uses a different pg library

saralab:
- This will be an expensive query to serve this view. We are discussing more architectural changes to serve this better - move toward a Lambda architecture that build a time series view

src/datastore/postgres-store.ts
jackromo888: I concure, now how do we go about doing that

src/api/routes/tx.ts
jackromo888: Starting to sound like a bit of a, well a bit of a dick, however im kinda a lead buy example type of guy, so im game now how do you recommend going about that

PR #1279, opened 2022-08-17
Implement support for Stacks 2.1 `versioned-smart-contract` transactions

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1274

The new `versioned-smart-contract` transactions are now being parsed, stored in pg, and the new `clarity_version` property is returned on applicable requests. 

For regular `smart-contract` transactions, the `clarity_version` property is null. 

Note: internally, we are storing this in pg as a new tx type (with the new type ID), but when generating http responses, we normalize the type as `"tx_type": "smart_contract"` so as not to cause unnecessary breaking changes for clients.

####  `GET /extended/v1/tx/<txid>`:
For a `versioned-smart-contract` tx, with a `clarity_version` of `2`:
```json
{
  "tx_id": "0x33c573f5ed06f1feecaa4a9df0225e109416dbba9792abb0cd94869bbad4a88a",
  "nonce": 0,
  "fee_rate": "0",
  "sender_address": "ST000000000000000000002AMW42H",
  "sponsored": false,
  "post_condition_mode": "deny",
  "post_conditions": [],
  "anchor_mode": "any",
  "is_unanchored": false,
  "block_hash": "0xd99feeda9df6cd2019cd48614693501d378d212f57a45649e343539085f15f1b",
  "parent_block_hash": "0x57ee3ea78ff90944bd17696b4037cdf59be69208247bba210b6cc2a61f700727",
  "block_height": 2,
  "burn_block_time": 1660732231,
  "burn_block_time_iso": "2022-08-17T10:30:31.000Z",
  "parent_burn_block_time": 1660732220,
  "parent_burn_block_time_iso": "2022-08-17T10:30:20.000Z",
  "canonical": true,
  "tx_index": 0,
  "tx_status": "success",
  "tx_result": {
    "hex": "0x0703",
    "repr": "(ok true)"
  },
  "microblock_hash": "0x",
  "microblock_sequence": 2147483647,
  "microblock_canonical": true,
  "event_count": 0,
  "events": [],
  "execution_cost_read_count": 0,
  "execution_cost_read_length": 0,
  "execution_cost_runtime": 0,
  "execution_cost_write_count": 0,
  "execution_cost_write_length": 0,
  "tx_type": "smart_contract", <-- Note: this property stays the same to avoid unnecessary breaking changes to clients
  "smart_contract": {
    "clarity_version": 2, <-- NEW
    "contract_id": "ST000000000000000000002AMW42H.pox-2",
[...]

PR #1328, opened 2022-09-21
fix: upgrade stacks node versions to 2.05.0.3.0

### LINKED ISSUE #1317: upgrade node references to 2.05.0.3.0
We should upgrade our Dockerfiles to reflect the new node version. This upgrade changes how the node boots so we will need to refactor our unit test setup in all of our suites.

rafaelcr:
Fixes #1317


src/tests-rosetta/api.ts
```diff
@@ -129,6 +130,30 @@ describe('Rosetta API', () => {
...
+        parent_network_id: 1,
+        peer_version: 1,
+        server_version: 1,
+        stable_burn_block_height: block.burn_block_height,
+        stable_burn_consensus: block.burn_block_hash,
+        stacks_tip: block.block_hash,
+        stacks_tip_burn_block: block.burn_block_height,
+        stacks_tip_height: block.block_height,
+        unanchored_tip: ''
+      });
```
zone117x, line 156: This and a few other test fixes look familiar -- similar changes were required in the `stacks-2.1` branch:
* https://github.com/hirosystems/stacks-blockchain-api/pull/1280/files#diff-352542c23d3fe38ede85cc2611da70e48ef527e1b09babb12bc6bb4ad263f2e0
* https://github.com/hirosystems/stacks-blockchain-api/pull/1279/files

It looks like some of the issues were caused by the bump from only `stacks-blockchain:2.05.0.0.0` (as opposed to `stacks-blockchain:next / 2.1`).

In the API's `stacks-2.1` branch, a larger refactor was done to fix these. Stubbing out these responses could potentially hide bugs if the stacks-node behavior diverges from these stubbed behaviors. However, we can probably address that later once we have to merge the `stacks-2.1` branch. In the meantime, seems okay to merge.

rafaelcr: Thanks for the note, I'll keep that in mind.

I did this mock only to skip the stacks node dependency on this test. Integration tests were already done in `construction.ts` so and adding a new event server here created problems given that the stacks node block height did not restart at 1 (it stayed at the previous height created by other tests).

CharlieC3 (approved):
:shipit:

PR #1329, opened 2022-09-21
chore!: remove support for `/nft_events` endpoint

### LINKED ISSUE #1325: remove old `/nft_events` endpoint
Now that the new NFT endpoints have been around for a long time, we should remove support for the old `/nft_events` endpoint completely. It uses a very old and inefficient query that is taking some toll in our postgres db instance.

This endpoint has been deprecated in our docs for months.

rafaelcr:
Fixes #1325


zone117x (approved):
👋

PR #1332, opened 2022-09-23
feat: add token_type metadata for rosetta ft operations

### LINKED ISSUE #1302: [Rosetta] add metadata to transfer amounts to identify token types
To avoid confusion between STX transfers and other fungible token transfers in block or transaction responses, we should add an entry to the amount metadata, like so:
```
"amount": {
  "value": "9076708051",
  "currency": {
    "decimals": 6,
    "symbol": "DIKO"
  },
  "metadata": {
    "token_type": "ft"
  }
},
```
This way, users can identify token types when processing JSON responses

rafaelcr:
Fixes #1302

### DIFF
--- a/src/rosetta-helpers.ts
+++ b/src/rosetta-helpers.ts
@@ -379,6 +379,9 @@ function makeFtBurnOperation(
         decimals: ftMetadata.decimals,
         symbol: ftMetadata.symbol,
       },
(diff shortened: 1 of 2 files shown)


zone117x (approved):
Nice, lgtm once CI finishes!

PR #1334, opened 2022-09-28
fix: event_observer_requests json writes

rafaelcr:
Fix an issue related to the new pg library usage where events to the raw requests table were written as invalid JSONs to the `payload` column e.g.
```
"{\"burn_amount\": 1, \"burn_block_hash\": ...
```
instead of 
```
{"burn_amount": 1, "burn_block_hash": ...
```


zone117x (approved):
:(
In the preorg mode PR, I found that `text` is faster to read and write than `jsonb`, without much of a storage penalty (especially since postgres disk space hasn't ever been a constraint). It seems like the `json`/`jsonb` pg types are primarily used for _querying_ json values with the postgres json query syntax.

So we can merge this as is (ensuring `jsonb` types are stored correctly) or could switch to `text` for this column. Both require an event replay I think. Up to you

PR #1335, opened 2022-09-28
release v6.0.0

### LINKED ISSUE #1310: release v6.0.0
Now that we've solved BNS and event replay issues in v5.0.0 we should release v6.0.0 with the new changes from the `develop` branch

rafaelcr:
Closes #1310


.gitpod.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM blockstack/stacks-blockchain:2.05.0.0.0-stretch as corenode
+FROM blockstack/stacks-blockchain:2.05.0.3.0-stretch as corenode
```
CharlieC3, line 1: Upgrading version once more. Is `stretch` still needed? This is becoming an outdated debian distro, and the alpine image is much lighter.

```suggestion
FROM blockstack/stacks-blockchain:2.05.0.4.0 as corenode
```

.env
```diff
@@ -35,9 +35,6 @@ PG_APPLICATION_NAME=stacks-blockchain-api
 # See https://node-postgres.com/api/pool
 # PG_CONNECTION_POOL_MAX=10
 
-# Enable to have stacks-node events streamed to a file while the application is running
-# STACKS_EXPORT_EVENTS_FILE=/tmp/stacks-events.tsv
```
CharlieC3, line 39: I think we should keep this ENV var here (or documented somewhere else) since it's still used during event export or event replay.

docker/stx-rosetta.Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG STACKS_API_VERSION=v0.71.2
-ARG STACKS_NODE_VERSION=2.05.0.0.0
+ARG STACKS_NODE_VERSION=2.05.0.3.0
```
CharlieC3, line 2: Same as above

```suggestion
ARG STACKS_NODE_VERSION=2.05.0.4.0
```

readme.md
```diff
@@ -166,14 +166,6 @@ could be ran:
      time during import, but sacrifices some historical data. You can use this mode if you're mostly
      interested in running an API that prioritizes real time information.
 
-Alternatively, instead of performing the `export-events` command in step 1, an environmental
-variable can be set which enables events to be streamed to a file as they are received, while the
-application is running normally. To enable this feature, set the `STACKS_EXPORT_EVENTS_FILE` env var
-to the file path where events should be appended. Example:
-```
-STACKS_EXPORT_EVENTS_FILE=/tmp/stacks-node-events.tsv
```
CharlieC3, line 174: Same as the other comment, we should probably keep this around, but change the description on when/how it's used.

src/event-replay/event-replay.ts
```diff
@@ -138,14 +144,12 @@ export async function importEventsFromTsv(
...
   const readStream = fs.createReadStream(resolvedFilePath);
-  const rawEventsIterator = PgDataStore.getRawEventRequests(readStream, status => {
+  const rawEventsIterator = getRawEventRequests(readStream, status => {
     console.log(status);
   });
   // Set logger to only output for warnings/errors, otherwise the event replay will result
   // in the equivalent of months/years of API log output.
   logger.level = 'warn';
-  // Disable this feature so a redundant export file isn't created while importing from an existing one.
-  delete process.env['STACKS_EXPORT_EVENTS_FILE'];
```
CharlieC3, line 148: Is this assuming the `STACKS_EXPORT_EVENTS_FILE` ENV var won't be present in the environment when running an event replay? What happens if it still is? Would it cause problems?

stacks-blockchain/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM blockstack/stacks-blockchain:2.05.0.0.0-stretch as build
+FROM blockstack/stacks-blockchain:2.05.0.3.0-stretch as build
```
CharlieC3, line 1: Same as other comment

```suggestion
FROM blockstack/stacks-blockchain:2.05.0.4.0 as build
```

CharlieC3 (commented):
LGTM! Left a few comments

.gitpod.Dockerfile
rafaelcr: If it's OK with you, I'll open this one as an issue for v7... I'm not sure how this could affect our test suites

.env
rafaelcr: Actually, it's no longer used anywhere 👍 
That ENV var only controlled the file path of the live TSV, so feel free to remove it from the pod ENV settings

src/event-replay/event-replay.ts
rafaelcr: No worries... since it's not used anywhere in code, having it defined in ENV will not cause any side effects

CharlieC3 (approved):

PR #1337, opened 2022-09-30
fix: support multiple BNS name events in the same transaction

### LINKED ISSUE #1333: Wrong owner of bns name
**Describe the bug**
The endpoint https://stacks-node-api.mainnet.stacks.co/v1/names/ape.mega shows SPC0KWNBJ61BDZRPF3W2GHGK3G3GKS8WZ7ND33PS as owner. However, the nft is owned by SPV48Q8E5WP4TCQ63E9TV6KF9R4HP01Z8WS3FBTG through tx https://explorer.stacks.co/txid/0xf9f9144793f6d4da9aba92a54ab601eb23bfe7f44c1edb29c2920bf5e7d2ac16?chain=mainnet

**To Reproduce**
Compare the two links provided

**Expected behavior**
the endpoint should show SPV48Q8E5WP4TCQ63E9TV6KF9R4HP01Z8WS3FBTG as owner

rafaelcr:
We used to assume BNS names could only have 1 event (register, transfer, etc.) per transaction, however, newer contracts like [ryder-handles-controller-v1](https://explorer.stacks.co/txid/SPC0KWNBJ61BDZRPF3W2GHGK3G3GKS8WZ7ND33PS.ryder-handles-controller-v1?chain=mainnet) are performing both `name-register` and `name-transfer` events at the same time which causes the API to return the wrong name owner. To fix this, we have to keep track of `event_index` in the names table and use that field to sort name changes on queries. 

This PR also fixes another bug that caused owner inconsistencies when multiple names were registered and transferred in the same block.

Fixes #1333

### DIFF
--- a/src/datastore/common.ts
+++ b/src/datastore/common.ts
@@ -517,11 +517,12 @@ export interface DbBnsName {
   expire_block: number;
   grace_period?: number;
   renewal_deadline?: number;
-  resolver?: string | undefined;
+  resolver?: string;
   zonefile: string;
   zonefile_hash: string;
   tx_id: string;
   tx_index: number;
+  event_index?: number;
   status?: string;
   canonical: boolean;
 }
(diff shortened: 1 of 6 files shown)


zone117x (approved):
Awesome, thanks for the quick fix 💯

Issue #1278, opened 2022-08-17
[Stacks 2.1] Support PoX-2 in Rosetta stacking operations


zone117x:
### Progress:
* [x] MVP support for PoX-2 / parity with current PoX-1 support
* [x] Support constructing & reading Stacking ops using the new segwit/taproot/bech32 address formats introduced in PoX-2
* [x] New integration test suite for Rosetta + PoX-2 against an Epoch 2.1 Stacks-node
* [x] Support in `/extended/v1/debug/broadcast/stack` (debug/testing webpage) to allow using the Rosetta implementation for performing Stacking operations
* [x] Fix locked STX balances and generate synthetic `stack-unlock` operations during the PoX-2 "period 2" transition (when all PoX-1 locked state is force unlocked at time PoX-2 is activated) -- see blocking issue https://github.com/stacks-network/stacks-blockchain/issues/3289
* [x] Issue `stack-stx` Rosetta operations to the correct pox contract version during the Epoch2.1 PoX transition periods

----
See related issue https://github.com/hirosystems/stacks-blockchain-api/issues/1277 which goes over similar details and questions around PoX-2.

The Rosetta implementation currently contains rosetta-operations for initiating a Stacking transaction. It also returns a synthetic "stx-unlocked" operation when querying for a block operations, where the block corresponds to an unlock height provided in the Stacks node `stx_lock_event` event. 

This was initially implemented in PR https://github.com/hirosystems/stacks-blockchain-api/pull/689

At a high level, here are the changes required to ensure Rosetta Stacking functionality works with Pox-2:

### PoX address format changes
PoX-2 functions will support segwit/taproot/bech32 addresses. See https://github.com/stacks-network/stacks-blockchain/issues/2586. This means the function signatures that used to use `{ (hashbytes (buff 20)) (version (buff 1)) }` will be switched to `{ (hashbytes (buff 32)) (version (buff 1)) }` (32 hashbytes instead of 20). The allowed `version` bytes will also be expanded to include new address formats. Various areas of code that assumed a 1-to-1 conversion between a STX and BTC address (C32 and B58) need to be refactored -- as the new BTC address formats cannot be represented as STX addresses.

### Constructing Stacking Operations
[...]

PR #1339, opened 2022-10-02
[Stacks 2.1] PoX-2 support in Rosetta

zone117x:
Issue https://github.com/hirosystems/stacks-blockchain-api/issues/1278

This PR adds support for the `stack-stx` operation in the Rosetta implementation. The primary intention is to ensure existing usage of `stack-stx` continues to work when PoX-2 is activated. Support for the _new_ PoX-2 operations (e.g. `stack-extend`, `stack-increase`) is not implemented, blocked on https://github.com/stacks-network/stacks-blockchain/issues/3289.

#### Added support for constructing PoX-2 `stack-stx` operations in the Rosetta construction endpoints:
  * __[KNOWN BUG/LIMITATION]__ The new Bitcoin address formats (e.g. segwit/taproot) are supported, although not tested end-to-end, currently blocked on https://github.com/stacks-network/stacks-blockchain/issues/2586 
  * cc @janniks I borrowed some of your WIP PoX-2 address encoding code from https://github.com/hirosystems/stacks.js/pull/1354 and added some fixes and tests, this code can be pulled back into `@stacks/stacking` at some point

#### Added support for parsing PoX-2 `stack-stx` operations in the Rosetta data querying endpoints:
  * __[KNOWN BUG/LIMITATION]__ This is still using the original approach of detecting only direct contract-calls to the `stack-stx` function in the PoX contracts. This means that it will work for clients that _only_ use the Rosetta implementation for creating _and_ reading stacking operations. If Rosetta is being used to monitor locked STX state for a given principal, and that principal performs a Stacking operation that is not currently supported by Rosetta (e.g. `stack-extend`, or using another contract to invoke `stack-stx`), then the client using Rosetta will end up with incorrect locked state data for that principal. Not much can be improved here until the Stacks node emits more events for the new operations, currently blocked on https://github.com/stacks-network/stacks-blockchain/issues/3289
  * __[KNOWN BUG/LIMITATION]__ This is also true for the end of "period 2" (when all PoX-1 locked state is force unlocked at time PoX-2 is activated) -- the Stacks node does not emit events for these unlocks. Systems using Rosetta may still operate as if the locks from PoX-1 are still in effect. This is difficult to fix in the Rosetta layer without new events, blocked on the same issue as above.

## Testing

The `mocknet` mode in the stacks-blockchain `next` branch cannot currently be configured to activate epoch 2.1 (or PoX-2). To work around this, a [new docker image](https://github.com/zone117x/stacks-regtest-env) has been created which operates like a drop-in replacement for `mocknet` mode. The image uses a 2.1-capable local testnet: `krypton` mode, a bundled bitcoind regtest instance, and pre-configured to bootstrap to epoche 2.1 and PoX-2.

For details about this new setup, see https://github.com/zone117x/stacks-regtest-env. Eventually, the code for this new docker image should probably be pulled into the API repo (or the repo should be moved to the Hiro org).

A new test suite has been added which uses the new Stacks 2.1 environment. It can be ran locally with `npm run test:integration:2.1`, and is also being ran in CI as a new job `test-2_1`.

Additionally, extensive manual testing is performed by launching a local API instance with the vscode `Launch: krypton 2.1-transition` task -- this spins up an API instance with the new image. Initially blocks are mined in epoch `2.0`, then after ~15 blocks epoch `2.1` is activated, then after an addition ~15 blocks `pox-2` is activated.

The debug webpage at http://localhost:3999/extended/v1/debug/broadcast/stack has been upgraded to allow using the Rosetta implementation for performing Stacking operations:

image


src/core-rpc/client.ts
```diff
@@ -104,8 +136,8 @@ export class StacksCoreRpcClient {
    * Throws an error if connection cannot be established.
    * @param retryTimeout - milliseconds
    */
-  async waitForConnection(retryTimeout = 30000): Promise<void> {
-    const retryInterval = 1000; // 1 second
+  async waitForConnection(retryTimeout = 60000): Promise<void> {
+    const retryInterval = 2500; // 1 second
```
rafaelcr, line 140:
```suggestion
    const retryInterval = 2500; // 2.5 seconds
```

src/rosetta-helpers.ts
```diff
@@ -798,25 +800,32 @@ function parseDelegateStxArgs(contract: ContractCallTransaction): RosettaDelegat
     throw new Error(`Could not find field function_args in contract call`);
   }
 
+  // (define-public (delegate-stx
+  //   (amount-ustx uint)
+  //   (delegate-to principal)
+  //   (until-burn-ht (optional uint))
+  //   (pox-addr (optional { version: (buff 1),
+  //                         hashbytes: (buff 20) })))
+
```
rafaelcr, line 809: This seems to have slipped in (also some comments in lines below)

src/tests-2.1/setup.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ChainID } from '@stacks/transactions';
+
+export interface TestEnvContext {
+  db: PgWriteStore;
+  eventServer: EventStreamServer;
+  api: ApiServer;
+  client: StacksCoreRpcClient;
+}
+
+async function standByForPox2ToBeReady(client: StacksCoreRpcClient): Promise<void> {
```
rafaelcr, line 16: Just FYI I think this will have to be removed once the 2.1 branch gets merged with the latest 2.05.0.3.0 code

rafaelcr (approved):
LGTM 👍 just a couple of comments

src/rosetta-helpers.ts
zone117x: It was put in somewhat intentionally to give context to the reader for what the below `function_args` correspond to (since they were changed to get by index rather than unreliable name)

src/tests-2.1/setup.ts
zone117x: You're referencing the bugged startup order right? Where the stacks-node RPC is not responsive until the API event server is responsive?

If so, this branch already has that fixed -- in this case by starting the event-server _then_ waiting for an RPC endpoint condition.

PR #1345, opened 2022-10-06
fix: guard against empty lists before querying postgres

rafaelcr:
this fix makes sure we don't query postgres with empty arrays in certain endpoints, which was causing 500 errors in some of our API pods. It also removes an unused `getSmartContractList` function.


CharlieC3 (approved):
:shipit:

PR #1348, opened 2022-10-11
fix: catch pg exceptions on queries outside of express

rafaelcr:
Express has an exception handler that resolves them gracefully, but we need to make sure we catch them on any query outside of it e.g. the token metadata handlers.

This PR:
* Adds exception handlers to all queries in the token metadata handlers
* Edits the error response to a 503 when a user calls an endpoint which results in a pg connection error
* Adds the new postgres lib's connection errors to the error handlers we have in place


CharlieC3 (approved):
LGTM! :shipit:

PR #1355, opened 2022-10-13
feat: configurable pg connection lifetime and idle timeouts

rafaelcr:


.env
```diff
@@ -27,12 +31,13 @@ PG_APPLICATION_NAME=stacks-blockchain-api
 # PG_PRIMARY_DATABASE=
 # PG_PRIMARY_SCHEMA=
 # PG_PRIMARY_SSL=
+# PG_PRIMARY_IDLE_TIMEOUT=
+# PG_PRIMARY_MAX_LIFETIME=
```
CharlieC3, line 35: Are these being used in code anywhere?

rafaelcr: Yep, they will be chosen if they're available: https://github.com/hirosystems/stacks-blockchain-api/blob/0eeee9558ae6ad4d0d233dca49c87430c0da57a6/src/datastore/connection.ts#L125-L144

CharlieC3: Interesting,  thanks!

CharlieC3 (approved):
:shipit:

Issue #1313, opened 2022-09-08
only save events to event_observer_requests table if message processing succeeded


rafaelcr:
we currently use a [catch-all handler](https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/event-stream/event-server.ts#L728-L743) in the stacks node event server that saves incoming events to the `event_observer_requests` table. And separately, we have individual route handlers for each type of event (new block, microblock, etc.).

This may create issues if the event got saved correctly to the DB but the individual handler had an error and returned a non-200 response to the stacks node, causing our events table to be out of sync with what actually happened inside the API.

We should make these two processes depend on each other.

PR #1359, opened 2022-10-17
fix(api): updates to event_observer_requests occur after specific eve…

BLuEScioN:

## Description

Describe the changes that where made in this pull request. When possible start with a user story - short, simple descriptions of a feature told from the perspective of the person who desires the new capability. Be sure to also include the following information:

1. Motivation for change

In situations where the route event handlers fail the db tables get out of sync because the event_observer_requests table records the event regardless of the route event handler failure. Thus, when the route event handler fails, the event request will be retried and the event_observer_request table will record duplicate events.

This issue is described here https://github.com/hirosystems/stacks-blockchain-api/issues/1313

3. What was changed

Removed event handler for the event_observer_request table from the wildcard route. Added it to all routes as a second event handler. All routes now call next to proceed to their second event handlers for recording the event in the event_observer_requests table

5. How does this impact application developers

Should be no change

7. Link to relevant issues and documentation
8. Provide examples of use cases with code samples and applicable acceptance criteria

Example:
As a Hiro Systems developer, I would like to encrypt files using the app private key. This is needed because storing unencrypted files is unacceptable. This pull request adds the `encryptContent` function which will take a string and encrypt it using the app private key.

```
encryptContent('my data')

// Running the above should result in the following encrypted data object
{"iv":"c91...","ephemeralPK":"031...","cipherText":"d61...","mac":"e73..."}
```

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

List the APIs or describe the functionality that this PR breaks.
Workarounds for or expected timeline for deprecation

This should not change any functionaliy. It is a bug fix

No 

- [ ] Link to documentation updates: 

## Testing information

Yes
Event replay can test if events are able to be handled correctly
Run test with bad mocknet tsv (mutated parent hash) to verify proper failure
Write unit tests. Standup db and event server. Test with malformed event. Mock event handler to throw exception.

3. If it’s a bug fix, list steps to reproduce the bug
4. Briefly mention affected code paths
5. List other affected projects if possible
6. Things to watch out for when testing

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated


zone117x (commented):
Thanks @BLuEScioN! Can you add a test for this? Something like sending a bad POST body to `/new_block` (ensuring that it errors), then querying the raw events table to verify that it did not record the event.

src/event-stream/event-server.ts
```diff
@@ -728,12 +729,15 @@ export async function startEventServer(opts: {
       .json({ status: 'ready', msg: 'API event server listening for core-node POST messages' });
   });
 
+  const handleRawEventRequest = asyncHandler(async req => {
+    await messageHandler.handleRawEventRequest(req.path, req.body, db);
+  });
+
   app.post(
     '*',
```
rafaelcr, line 737: I think this handler can be removed completely now since event write is being moved to each event handler below 👍

src/event-stream/event-server.ts
```diff
@@ -728,105 +729,104 @@ export async function startEventServer(opts: {
...
-    asyncHandler(async (req, res, next) => {
-      const eventPath = req.path;
-      let payload = JSON.stringify(req.body);
-      await messageHandler.handleRawEventRequest(eventPath, req.body, db);
-      if (logger.isDebugEnabled()) {
-        // Skip logging massive event payloads, this _should_ only exclude the genesis block payload which is ~80 MB.
-        if (payload.length > 10_000_000) {
-          payload = 'payload body too large for logging';
-        }
-        logger.debug(`[stacks-node event] ${eventPath} ${payload}`);
```
zone117x, line 742: Any reason for removing this log?

BLuEScioN: @rafaelcr I took your comment as you wanted this removed. Did I read your comment correctly?

rafaelcr: @BLuEScioN I think you'll just need to move it to the `handleRawEventRequest` function you created below

src/datastore/pg-store.ts
```diff
@@ -2846,6 +2846,20 @@ export class PgStore {
     return { found: true, result: queryResult };
   }
 
+  async getRawEventCount(): Promise<number> {
+    try {
+      const result = await this.sql<
+        Promise<number>[]
+      >`SELECT count(*) from event_observer_requests`;
```
rafaelcr, line 2853: Since you just need this for testing I think you could simply perform this query directly inside the test:
```typescript
const result = await db.sql<{ count: number }[]>`SELECT count(*) from event_observer_requests`;
```
The reason I say this is because COUNTs are extremely expensive so we should not facilitate their use in our regular codebase

src/event-stream/event-server.ts
```diff
@@ -728,105 +729,114 @@ export async function startEventServer(opts: {
...
+  const handleRawEventRequest = asyncHandler(async req => {
+    await messageHandler.handleRawEventRequest(req.path, req.body, db);
+
+    const eventPath = req.path;
+    let payload = JSON.stringify(req.body);
```
rafaelcr, line 736: To save some CPU cycles, you should move these two lines to be inside the `logger.isDebugEnabled()` check

src/datastore/pg-store.ts
BLuEScioN: got it

src/event-stream/event-server.ts
BLuEScioN: done

src/tests-bns/event-server-tests.ts
```diff
@@ -1027,9 +1036,55 @@ describe('BNS event server tests', () => {
     expect(namespaceList.results).toStrictEqual(['ape.mega']);
   });
 
-  afterEach(async () => {
-    await eventServer.closeAsync();
-    await db?.close();
-    await runMigrations(undefined, 'down');
+  test('If there is an event request error, then the event will not be recorded in the events_observer_request table', async () => {
```
rafaelcr, line 1039: This looks great. Can you just move it to the Event Replay test suite instead of BNS?

rafaelcr (commented):
Looks very good, just one comment.

src/tests-bns/event-server-tests.ts
BLuEScioN: Ok

src/tests-bns/event-server-tests.ts
```diff
@@ -1,3 +1,4 @@
+import * as fs from 'fs';
```
rafaelcr, line 1: I don't think this is required anymore (same with imports below).
You should revert this file to be like `develop`:
```
git checkout develop -- src/tests-bns/event-server-tests.ts
```

src/tests-event-replay/import-export-tests.ts
```diff
@@ -99,3 +106,81 @@ describe('import/export tests', () => {
     await db?.close();
   });
 });
+
+describe('Events table', () => {
```
rafaelcr, line 110: Test is great. Since you're declaring a new `describe` block, you should send this to a new file inside `src/tests-event-replay`. I'd suggest `src/tests-event-replay/raw-event-request-tests.ts`

rafaelcr (changes requested):
Small changes and then we're good to go

rafaelcr (approved):
LGTM, feel free to merge after comments are addressed

zone117x (approved):
Super useful change, thanks @BLuEScioN!

PR #1360, opened 2022-10-17
fix: remove duplicate tx socket updates inside db transactions

rafaelcr:
These events are already being sent outside the update transaction. Sending them here may be causing postgres transaction consistency errors in reader pods.


CharlieC3 (approved):
:shipit:

PR #1363, opened 2022-10-18
PoX-2 integration tests

zone117x:
Add integration tests for several new PoX-2 functions such as `stack-extend`, `stack-increase`, etc. 

Fixed a bug where account balance endpoints like `/extended/v1/address/{addr}/stx` would return a 500 after multiple `StxLockEvent` are received for an address during the same Stacking cycle (e.g. after `pox-2::stack-increase` or `pox-2::stack-extend`).

Also closes https://github.com/hirosystems/stacks-blockchain-api/issues/1364 -- in anticipation of testing/supporting segwit PoX addresses.

This PR updates the `stacks-2.1` test suite so that a single global event observer (and the pg connection it requires) is created before tests are ran. This is required because the test suite boots a single krypton Stacks-node instance and bitcoind-regtest instance which are alive for the duration of the test suite. The alternative to this is attempting to orchestrate docker containers programmatically from within the js test src, which has proven to be painful and unreliable in some of our other test suites.

Note to reviewers: the postgres migration files previously existed in `/src/migrations/*.ts` -- however, programmatic usage of the pg-migration library does not actually support loading ts files. It worked previously in production because the ts files were compiled to js, and worked in debugging because the `ts-node` and `ts-jest` libraries inject transparent ts file loading support to allow for `require('<*.ts>')`. However, when using certain jest configurations such as `globalSetup` and `setupFilesAfterEnv` config, the auto-ts-loading functionality would break and prevent `migration/*.ts` files from loading correctly (thus breaking jest debugging in certain cases). The fix for this was to move the migration files to js. This is actually a pretty minimal change (it looks like a large change in the diff because of the renames) -- there should only be a minor merge annoyance later on.


src/tests-2.1/pox-2.ts
```diff
@@ -0,0 +1,405 @@
...
+        const dbTxQuery = await api.datastore.getTx({ txId: txId, includeUnanchored: false });
+        if (!dbTxQuery.found) {
+          return;
+        }
+        api.datastore.eventEmitter.removeListener('txUpdate', listener);
+        resolve(dbTxQuery.result);
+      };
+      api.datastore.eventEmitter.addListener('txUpdate', listener);
+    });
+  }
```
rafaelcr, line 48: I think we should refactor these onto the test utils (in another PR when we have some time in the future), they're very useful and used in a bunch of test suites

rafaelcr (approved):
Looks great!

PR #1366, opened 2022-10-18
feat: add ENV configs for DB close and API shutdown timeouts

### LINKED ISSUE #1352: read-only API left running indefinitely after receiving unhandled exception
**Describe the bug**
After receiving an unhandled exception (`unhandledRejection read ECONNRESET read ECONNRESET`), the API started terminating, but wasn't able to do so completely.

**Console log**
```
API WebSocket channels closed.
API Server was force killed after taking longer than 5000ms to shutdown
Closing API http server...
Closing Profiler server...
API http server closed.
Profiler server closed
Closing DB...
PgNotifier closed channel: stacks-api-pg-notifier
DB is taking longer than expected to shutdown, possibly hanging indefinitely
DB is taking longer than expected to shutdown, possibly hanging indefinitely
DB is taking longer than expected to shutdown, possibly hanging indefinitely
DB is taking longer than expected to shutdown, possibly hanging indefinitely
DB is taking longer than expected to shutdown, possibly hanging indefinitely
...
```

version 6.0.3

rafaelcr:
* Add `PG_CLOSE_TIMEOUT` and `PG_PRIMARY_CLOSE_TIMEOUT` to control seconds before killing in-flight queries when closing `PgStore` and `PgWriteStore`. Fixes #1352 
* Add `STACKS_SHUTDOWN_FORCE_KILL_TIMEOUT` to control seconds before force-killing an API component when shutting down
* Add `READ ONLY` annotation to all `BEGIN` postgres transactions on `PgStore`. No real performance upgrades from this but it is an extra security measure to avoid any modifications to the DB during read-only API execution: https://www.postgresql.org/docs/current/sql-set-transaction.html
* Fix style for `async`/`await` pg transaction use

### DIFF
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -380,24 +380,6 @@ export class PgStore {
     return { found: true, result: block } as const;
   }
 
-  async getBlocks({ limit, offset }: { limit: number; offset: number }) {
-    return await this.sql.begin('READ ONLY', async sql => {
-      const total = await sql<{ count: number }[]>`
-        SELECT block_count AS count FROM chain_tip
-      `;
-      const results = await sql<BlockQueryResult[]>`
-        SELECT ${sql(BLOCK_COLUMNS)}
-        FROM blocks
-        WHERE canonical = true
-        ORDER BY block_height DESC
-        LIMIT ${limit}
-        OFFSET ${offset}
-      `;
-      const parsed = results.map(r => parseBlockQueryResult(r));
-      return { results: parsed, total: total[0].count } as const;
-    });
-  }
-
   /**
    * Returns Block information with metadata, including accepted and streamed microblocks hash
    * @returns `BlocksWithMetadata` object including list of Blocks with metadata and total count.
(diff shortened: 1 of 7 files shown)


CharlieC3 (approved):
LGTM :shipit: 

On a related note the number of env vars are stacking up! How many do you think is too many?

src/datastore/pg-write-store.ts
```diff
@@ -96,7 +101,9 @@ class MicroblockGapError extends Error {
  */
 export class PgWriteStore extends PgStore {
   readonly isEventReplay: boolean;
-  private cachedParameterizedInsertStrings = new Map<string, string>();
+  protected get closeTimeout(): number {
+    return parseInt(getPgConnectionEnvValue('CLOSE_TIMEOUT', PgServer.primary) ?? '5');
```
CharlieC3, line 105: This is in the `PgWriteStore` class, does that mean it's only being applied to API writers? Or is this used for both writers/readers?

rafaelcr: There's a similar statement in `pg-store.ts`, this is only an override so I can change the PgServer to primary (it's a subclass)

src/datastore/pg-store.ts
```diff
@@ -133,7 +136,7 @@ export class PgStore {
 
   async close(): Promise<void> {
     await this.notifier?.close();
-    await this.sql.end();
+    await this.sql.end({ timeout: this.closeTimeout });
```
zone117x, line 139: This line is the primary fix in this PR right?

src/shutdown-handler.ts
```diff
@@ -19,7 +19,7 @@ async function startShutdown() {
     return;
   }
   isShuttingDown = true;
-  const timeoutMs = 5000;
+  const timeoutMs = parseInt(process.env['STACKS_SHUTDOWN_FORCE_KILL_TIMEOUT'] ?? '60') * 1000;
   let errorEncountered = false;
   for (const config of shutdownConfigs) {
     try {
```
zone117x, line 25: Potentially related to shutdown problems -- I recently came across line https://github.com/hirosystems/stacks-blockchain-api/blob/ebd8fff7f383cdd6a1c354fb362b9f7b3bb2efc0/src/helpers.ts#L708-L711

AFAIK this can lead to unhandledPromiseRejection errors, see https://eslint.org/docs/latest/rules/no-async-promise-executor

I think it needs to catch and reject the promise, like:
```ts
    new Promise(async (resolve, reject) => {
      try {
        await promise;
      } catch (error) {
        reject(error);
      } finally {
        clearTimeout(timer);
        resolve(true);
      }
    }),
```

src/datastore/pg-store.ts
rafaelcr: Yes, that's correct

src/shutdown-handler.ts
rafaelcr: Great catch, I'll add the fix to this PR

rafaelcr: Done. Moved the executor to *not* be async following the docs you linked to.

zone117x (approved):

PR #1367, opened 2022-10-19
[Stacks 2.1] Upgrade to bitcoinjs-lib-v6.x, initial P2TR work

zone117x:
Preparation for PoX-2 segwit support / testing:
* Upgrade to bitcoinjs-lib-v6.x (and its new refactored dependencies) which has taproot/P2TR support
* Util code for constructing key-spend only taproot/P2TR bitcoin addresses


rafaelcr (approved):
LGTM

PR #1372, opened 2022-10-20
PoX-2 integration testing for segwit addresses

zone117x:
Integration tests against stacks-blockchain branch https://github.com/stacks-network/stacks-blockchain/pull/3283 (segwit support) for PoX-2 operations

### DIFF
--- a/docker/docker-compose.dev.stacks-krypton-2.1-transition.yml
+++ b/docker/docker-compose.dev.stacks-krypton-2.1-transition.yml
@@ -1,7 +1,7 @@
 version: '3.7'
 services:
   stacks-blockchain:
-    image: "zone117x/stacks-api-e2e:stacks2.1-transition-8fb8b77"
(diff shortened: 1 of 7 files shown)


rafaelcr (approved):
LGTM - Would it make sense to (eventually) migrate these EC functions into the Rust Stacks encoding/decoding lib you wrote? Seems like there could be a new CPU win here but what do you think?

zone117x:
> Would it make sense to (eventually) migrate these EC functions into the Rust Stacks encoding/decoding lib you wrote? Seems like there could be a new CPU win here but what do you think?

Possibly -- depends on how often these functions will be called. So far it's only used in integration tests for segwit support in PoX-2, but may be used in the Rosetta layer for creating PoX-2 segwit operations later on.

However, the `tiny-secp256k1` package is now a WASM-compiled version of the same secp256k1 C library used in bitcoin (and the rust wrapper lib), so it's already fairly performant compared to the old js implementation.

Issue #1316, opened 2022-09-15
API stays in IBD mode even after the block height specified in IBD_MODE_UNTIL_BLOCK environment variable


mutharasus:
**Is your feature request related to a problem? Please describe.**
Currently the API Node / Database does not have a way to go into a "Initial Block Download(IBD)" mode when the node is far behind its other peers. There is a [proposed feature](https://github.com/stacks-network/stacks-blockchain/pull/2768) that is a likely long term solution. This feature request proposes as a stop gap solution until a permanent fix is implemented on the blockchain itself.

**Describe the solution you'd like**
The solution being proposed is to introduce a new environment variable which when set to a particular block height informs the API Node to go into a "Initial Block Download(IBD)" mode similar to Bitcoin where in it can avoid any processing that could be considered redundant until the node is fully synced up to its peers. Some examples for processing that can be avoided are: REFRESH MATERIALIZED VIEW SQLs that are extremely CPU intensive on the PG instance, Mempool messages, etc.,

The name of the environment variable can be "IBD_MODE_UNTIL_BLOCK=12345" or similar.

**Describe alternatives you've considered**
Other alternatives considered beyond the long term solution mentioned in the problem statement above is, introduce a environment variable which is a boolean flag that can be manually set to indicate the node should operate in a "Initial Block Download(IBD)" mode. The down side with this approach is that the node operator will have to stop the node and restart it with the flag reset once the node is in sync with its peers. 

**Additional context**
None

UPDATE: This environment variable was added in API v7.0.0 as part of a different issue. This issue is left open to address the remaining work. Title has been adjusted to reflect this.

rafaelcr:
Related to #698

zone117x:
Testing is blocked on https://github.com/stacks-network/stacks-blockchain/issues/3456

wileyj:
> Testing is blocked on [stacks-network/stacks-blockchain#3456](https://github.com/stacks-network/stacks-blockchain/issues/3456)

Is this still blocked? it seems unrelated (only evidence we could find is that the burnchain node was just overwhelmed at the time of the errors). 

another idea i had when speaking to @mutharasus was rather than adding this as a one-off env var, it might be a cool idea to add something like an API key env var locally that can be used to POST admin changes to the API. 

something like this, where the sample endpoint is firewalled for localhost requests:
```bash
curl --request POST \
  'http://localhost:3999/extended/v1/admin/setvars' \
[...]

PR #1373, opened 2022-10-21
feat(ibd): added middleware for checking chainheight

BLuEScioN:

## Description

Describe the changes that where made in this pull request. When possible start with a user story - short, simple descriptions of a feature told from the perspective of the person who desires the new capability. Be sure to also include the following information:

Currently the API Node / Database does not have a way to go into a "Initial Block Download(IBD)" mode when the node is far behind its other peers. There is a permanent solution (https://github.com/stacks-network/stacks-blockchain/pull/2768) that is a likely long term solution. This feature request proposes as a stop gap solution until a permanent fix is implemented on the blockchain itself.

The solution being proposed is to introduce a new environment variable which when set to a particular block height informs the API Node to go into a "Initial Block Download(IBD)" mode similar to Bitcoin where in it can avoid any processing that could be considered redundant until the node is fully synced up to its peers. Some examples for processing that can be avoided are: REFRESH MATERIALIZED VIEW SQLs that are extremely CPU intensive on the PG instance, Mempool messages, etc.,

1. Motivation for change

Implement a stop gap solution to help improve syncing speeds when a node is far behind its peers.

3. What was changed

When the env var for IBD mode is set, we activate a middleware that queries for the chaintip if the chaintip is behind the env var value, then particular routes and their route handlers are not called.

The refreshed materialized view function is now doing the same thing and not running if the chaintip is behind the env var

5. How does this impact application developers

Shouldn't because they would have to set the env var

7. Link to relevant issues and documentation

https://github.com/hirosystems/stacks-blockchain-api/issues/1316

9. Provide examples of use cases with code samples and applicable acceptance criteria

Example:
As a Hiro Systems developer, I would like to encrypt files using the app private key. This is needed because storing unencrypted files is unacceptable. This pull request adds the `encryptContent` function which will take a string and encrypt it using the app private key.

```
encryptContent('my data')

// Running the above should result in the following encrypted data object
{"iv":"c91...","ephemeralPK":"031...","cipherText":"d61...","mac":"e73..."}
```

## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Other

List the APIs or describe the functionality that this PR breaks.
Workarounds for or expected timeline for deprecation

No, the code changes are behind a check for the presence of an env var

- [ ] Link to documentation updates: 

We should update the documentation to inform api providers that they can use this env var to improve their sync speeds.

## Testing information

Yes
Set env var
Set db to be behind env var
Test if routes run
Pre-existing tests of these routes should suffice for testing whether their route handlers run in the absence of this env var

Improvement:

image

3. If it’s a bug fix, list steps to reproduce the bug
4. Briefly mention affected code paths
5. List other affected projects if possible
6. Things to watch out for when testing

## Checklist
- [X] Code is commented where needed
- [ ] `npm run test` passes
- [ ] Changelog is updated


.env
```diff
@@ -140,3 +140,9 @@ MAINNET_SEND_MANY_CONTRACT_ID=SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.send-man
 # of openapi.yaml for test / development NODE_ENV. 
 # For production, /doc is not served if this env var is not provided.
 # API_DOCS_URL="https://docs.hiro.so/api" 
+
+# For use while syncing. Places the API into an "Initial Block Download(IBD)" mode, 
+# forcing it to stop any redundant processing until the node is fully synced up to its peers. 
+# Some examples of processing that are avoided are: 
+# REFRESH MATERIALIZED VIEW SQLs that are extremely CPU intensive on the PG instance, Mempool messages, etc.,
+# IBD_MODE_UNTIL_BLOCK=123
```
rafaelcr, line 148: Since the default is null, you should leave it as blank in the example (remove the 123)

src/datastore/pg-write-store.ts
```diff
@@ -2450,6 +2449,16 @@ export class PgWriteStore extends PgStore {
     if (this.isEventReplay && skipDuringEventReplay) {
       return;
     }
+    if (process.env.IBD_MODE_UNTIL_BLOCK) {
+      const db = new PgWriteStore(sql);
+      const chainTip = await db.getChainTip(sql);
```
rafaelcr, line 2454: You can use `this` instead of having to instantiate a new `PgWriteStore`

src/datastore/pg-write-store.ts
```diff
@@ -2450,6 +2449,16 @@ export class PgWriteStore extends PgStore {
...
+      if (
+        process.env.IBD_MODE_UNTIL_BLOCK &&
```
rafaelcr, line 2456: This check is already done in the outer `if`, you can remove it here

src/event-stream/event-server.ts
```diff
@@ -722,6 +723,25 @@ export async function startEventServer(opts: {
...
 
   app.use(bodyParser.json({ type: 'application/json', limit: '500MB' }));
+
+  if (process.env.IBD_MODE_UNTIL_BLOCK) {
+    app.use(['/new_mempool_tx', '/drop_mempool_tx', '/new_burn_block'], async (req, res, next) => {
+      try {
+        const chainTip = await db.getChainTip(db.sql, false);
+        if (
+          process.env.IBD_MODE_UNTIL_BLOCK &&
+          chainTip.blockHeight > Number.parseInt(process.env.IBD_MODE_UNTIL_BLOCK)
```
rafaelcr, line 733:
```suggestion
          chainTip.blockHeight >= Number.parseInt(process.env.IBD_MODE_UNTIL_BLOCK)
```

src/event-stream/event-server.ts
```diff
@@ -722,6 +723,25 @@ export async function startEventServer(opts: {
   );
 
   app.use(bodyParser.json({ type: 'application/json', limit: '500MB' }));
+
+  if (process.env.IBD_MODE_UNTIL_BLOCK) {
+    app.use(['/new_mempool_tx', '/drop_mempool_tx', '/new_burn_block'], async (req, res, next) => {
+      try {
+        const chainTip = await db.getChainTip(db.sql, false);
+        if (
+          process.env.IBD_MODE_UNTIL_BLOCK &&
```
rafaelcr, line 732: You can remove this check

src/event-stream/event-server.ts
```diff
@@ -722,6 +723,25 @@ export async function startEventServer(opts: {
...
+          chainTip.blockHeight > Number.parseInt(process.env.IBD_MODE_UNTIL_BLOCK)
+        ) {
+          next();
+        } else {
+          res.send(`${req.originalUrl} is not available while IBD mode is active.`);
```
rafaelcr, line 737: You should only send a blank `200` status response in this case

src/event-stream/event-server.ts
```diff
@@ -722,6 +723,25 @@ export async function startEventServer(opts: {
...
+        }
+      } catch (error) {
+        console.error(error);
```
rafaelcr, line 740: I'd suggest adding a message string here (including the error at the end) to make it more informative

src/tests-bns/event-server-tests.ts
```diff
@@ -1027,6 +1027,56 @@ describe('BNS event server tests', () => {
...
+
+    const routes = ['/new_mempool_tx', '/drop_mempool_tx', '/new_burn_block']
+
+    routes.forEach(async route => {
+      const response = await httpPostRequest({
+        host: '127.0.0.1',
+        port: eventServer.serverAddress.port,
+        path: route,
+        headers: { 'Content-Type': 'application/json' },
+        body: Buffer.from(JSON.stringify({}), 'utf8'),
```
rafaelcr, line 1063: I think you should test a real payload here, I can help you with one tomorrow so you can add it

src/tests-bns/event-server-tests.ts
```diff
@@ -1027,6 +1027,56 @@ describe('BNS event server tests', () => {
     expect(namespaceList.results).toStrictEqual(['ape.mega']);
   });
 
+  test('IBD mode blocks certain API routes', async () => {
```
rafaelcr, line 1030: These tests look great but they should go in the event replay test suite instead of BNS.

rafaelcr (commented):
Looks great so far! Left a few comments mostly on style and some optimizations

src/event-stream/event-server.ts
BLuEScioN: done

src/event-stream/event-server.ts
BLuEScioN: done

src/event-stream/event-server.ts
BLuEScioN: done

src/tests-bns/event-server-tests.ts
BLuEScioN: done

src/event-stream/event-server.ts
BLuEScioN: done

src/datastore/pg-write-store.ts
BLuEScioN: done

src/datastore/pg-write-store.ts
BLuEScioN: done

.env
BLuEScioN: done

src/tests-bns/event-server-tests.ts
BLuEScioN: done

rafaelcr (commented):
This looks very good, I think it only needs a test (you should add one to the Event Replay test suite).

Also, this ENV var should be used during `pruned` event replay as well. It's a very quick change, we can sync tomorrow to discuss this if you have some time.

src/event-replay/event-replay.ts
```diff
@@ -112,11 +113,13 @@ export async function importEventsFromTsv(
...
     process.env['STACKS_MEMPOOL_TX_GARBAGE_COLLECTION_THRESHOLD'] ?? '256'
   );
-  const prunedBlockHeight = Math.max(tsvBlockHeight - blockWindowSize, 0);
+  const prunedBlockHeight =
+    prunedBlockHeightOption ?? Math.max(tsvBlockHeight - blockWindowSize, 0);
   console.log(`Event file's block height: ${tsvBlockHeight}`);
   console.log(`Starting event import and playback in ${eventImportMode} mode`);
   if (eventImportMode === EventImportMode.pruned) {
     console.log(`Ignoring all prunable events before block height: ${prunedBlockHeight}`);
+    process.env.IBD_MODE_UNTIL_BLOCK = `${prunedBlockHeight}`;
```
zone117x, line 122: Can the new `IBD_MODE_UNTIL_BLOCK` var be "documented" in the `.env` file? (quotes because the .env file is probably not the best place to document these but it's what we have today 😅 )

Just something like
```
# Uncomment to enable "initial block download mode", while syncing this will ....
# IBD_MODE_UNTIL_BLOCK=100
```

zone117x: Oops nvm! I missed where you had already done this! Thanks!

.env
```diff
@@ -140,3 +140,9 @@ MAINNET_SEND_MANY_CONTRACT_ID=SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.send-man
 # of openapi.yaml for test / development NODE_ENV. 
 # For production, /doc is not served if this env var is not provided.
 # API_DOCS_URL="https://docs.hiro.so/api" 
+
+# For use while syncing. Places the API into an "Initial Block Download(IBD)" mode, 
+# forcing it to stop any redundant processing until the node is fully synced up to its peers. 
+# Some examples of processing that are avoided are: 
+# REFRESH MATERIALIZED VIEW SQLs that are extremely CPU intensive on the PG instance, Mempool messages, etc.,
+# IBD_MODE_UNTIL_BLOCK=
```
zone117x, line 148: Would be neat if this could be set to the URL of a stable API deployment, and the init code is to "if (idbIsUrl) idbHeight = fetch(idbUrl).blockHeight" ... but that has some complexity and weird implications. Probably best to wait until the stacks-node itself is able to start reporting this value.

zone117x:
Looks like there's some git history issues. @BLuEScioN I think your changes for this PR probably need to be rebased onto latest develop branch?

src/event-stream/event-server.ts
```diff
@@ -722,111 +738,126 @@ export async function startEventServer(opts: {
...
         await messageHandler.handleNewAttachment(msg, db);
         res.status(200).json({ result: 'ok' });
+        next();
       } catch (error) {
         logError(`error processing core-node /attachments/new: ${error}`, error);
         res.status(500).json({ error: error });
       }
-    })
+    }),
+    handleRawEventRequest
```
zone117x, line 844: It looks like branch https://github.com/hirosystems/stacks-blockchain-api/pull/1359 was merged into this branch, I'm assuming that was unintentional? It looks like a rebase or cherrypicking is necessary for this branch

zone117x (commented):
Do we have mainnet testing & benchmarks for these changes?

BLuEScioN:
> Do we have mainnet testing & benchmarks for these changes?



> Do we have mainnet testing & benchmarks for these changes?


Here is the results of the IBD mode playback NODE_ENV=production IBD_MODE_UNTIL_BLOCK=78320 node lib/index.js import-event  1155.07s user 284.84s system 29% cpu 1:21:07.23 total

The event replay command ran without any errors.

We were discussing past benchmarks and future benchmarks in this Slack thread https://hiropbc.slack.com/archives/C03UFUA5Q65/p1667598495593179

Summarizing, we don't currently have any stored data to compare this timing against, unless logs already exist that has this data. Benchmarking will vary widely based on reply environment and available resources, so that will have to be taken into account as well.

rafaelcr (approved):
LGTM, but this needs a rebase @BLuEScioN to make sure new tests pass

zone117x (approved):

PR #1374, opened 2022-10-21
fix: guarantee db is empty before performing a replay

rafaelcr:
* To determine if the DB is empty before a replay, use the `pg_class` table to look for any data structure that may be remaining (tables, indexes, views, constraints).
* When clearing the DB, use the same `pg_class` table to first delete materialized views and then all tables with `CASCADE`.
* When performing `import-events` and cycling migrations, make sure we did in fact remove *all* data from the DB (except migration tables) and throw an error if we didn't.

Also:
* Create an Event Replay test suite to test all of these cases
* Include `mainnet.tsv` and `mocknet.tsv` example TSVs to perform a full replay cycle during tests


src/datastore/migrations.ts
```diff
@@ -65,30 +71,31 @@ export async function dangerousDropAllTables(opts?: {
...
-  } catch (error) {
-    await client.query('ROLLBACK');
-    throw error;
+    await sql.begin(async sql => {
+      const relNamesQuery = async (kind: string) => sql<{ relname: string }[]>`
+        SELECT relname
+        FROM pg_class c
+        JOIN pg_namespace s ON s.oid = c.relnamespace
+        WHERE s.nspname = ${schema} AND c.relkind = ${kind}
+      `;
```
zone117x, line 86: Is there a way to constrain this query to the database specified in `clientConfig.database`? Otherwise, it seems like this could drop everything from the entire postgres instance rather than only the specific database.

The previous code was marked as "dangerous"/"unsafe" because it's conceivable that devs might _also_ use the same database for their app-specific data, e.g. something like a `stacks_blockchain_api` db with tables like:
```
transactions
blocks
...<the other tables configured by migration files>
my_app_users
my_app_posts
```

So the idea behind requiring that `--force` flag for wiping is that devs might think twice about this situation. However, it seems like the above code would drop data even if devs put their app-specific data in a separate database yet still in the same postgres instance.

rafaelcr: The `pg_class` system catalog is actually database-specific. There are some which are shared but `pg_class` is not:
> Most system catalogs are copied from the template database during database creation and are thereafter database-specific. A few catalogs are physically shared across all databases in a cluster; these are noted in the descriptions of the individual catalogs.

https://www.postgresql.org/docs/14/catalogs-overview.html

> it's conceivable that devs might also use the same database for their app-specific data

This is a great point, though, that I think we should definitely discourage somehow. The difficulty with this is that when we change schemas across API version we'd need to somehow keep track of historical migrations to validate that we're only dropping things the API created as opposed to the developer.

I could expand the warning message we display during replay to explain this situation (i.e. check for empty DB, if not, ask the user to use --wipe-db --force but also explain they shouldn't keep other data in the same DB)

zone117x: > The pg_class system catalog is actually database-specific. There are some which are shared but pg_class is not

Got it, I think in that case this PR is good to go. The extra log warning sounds helpful but doesn't have to be in the PR imo since there doesn't seem to be extra danger introduced with this change.

zone117x (approved):

Issue #1375, opened 2022-10-21
connections remain in `idle in transaction` when running v6 in production


rafaelcr:
In our API v6 production runs, we've seen sql connections that remain `idle in transaction` with a stuck `begin` query... this could be from failed transaction management or postgres dropping some queries due to replication race conditions when refreshing materialized views.

We should also investigate how we open and close transactions across the codebase to see if there is cleanup to be done in that regard.

Issue #1370, opened 2022-10-19
Rosetta error in /block when searching by block_height


rafaelcr:
After upgrading from 6.0.1 to 6.1.0, a user ran into this issue:
```
"message": "{\"respBody\":{\"error\":\"Error: unexpected tx not found -- could not get contract from data store\",\"stack\":\"Error: unexpected tx not found -- could not get contract from data store\\n    at makeCallContractOperation (/app/src/rosetta-helpers.ts:595:11)\\n    at runMicrotasks (<anonymous>)\\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\\n    at getOperations (/app/src/rosetta-helpers.ts:141:23)\\n    at parseRosettaTxDetail (/app/src/api/controllers/db-controller.ts:508:22)\\n    at getRosettaBlockTransactionsFromDataStore (/app/src/api/controllers/db-controller.ts:578:23)\\n    at getRosettaBlockFromDataStore (/app/src/api/controllers/db-controller.ts:317:16)\\n    at /app/src/api/routes/rosetta/block.ts:33:21\",\"errorTag\":\"9a140353-b782-4fe4-9f67-ba8269002525\"},\"status\":500,\"req\":{\"path\":\"/block\",\"body\":{\"block_identifier\":{\"index\":80103},\"network_identifier\":{\"blockchain\":\"stacks\",\"network\":\"mainnet\"}}}}",
```
Happens when they get the latest block from status and then try to query the /block endpoint immediately afterward.

saralab:
Awaiting v6 to confirm the fix works

PR #1376, opened 2022-10-21
fix: nest sql transactions correctly

rafaelcr:
This PR fixes the use of postgres `BEGIN`/`COMMIT` transactions across the entire API

### Background
For code reusability, we have functions within `PgStore` that each perform a series of queries against the DB within their own `BEGIN`/`COMMIT` transaction blocks for result consistency. We then group some of these function calls inside other utility functions (e.g. in `db-controller.ts`) and use them in different parts of the API like endpoints, tests, Rosetta, etc.

This creates several problems:

1. If SQL transactions are performed in series without an enclosing transaction, results may be inconsistent or wrong if a new block is being written to the DB at the same time (causing #1370)
2. If SQL transactions are nested (i.e. a function that opens a transaction and calls another function inside which opens another transaction) it can lead to an incorrect use of SQL connections and to `WARNING: there is already a transaction in progress` logs appearing in the DB (probably causing #1375)

### Solution
This PR creates a new utility function called `sqlTransaction` which allows us to use and nest transaction blocks wherever we want.

When used, this function checks the SQL connection to see if it was already scoped to a transaction. If so, it runs queries directly instead of calling `BEGIN`. This allows us to keep consistent transaction use and to return correct results in all of our endpoints.

To enforce correct transaction use, all `PgStore` functions now require a first `PgSqlClient` parameter which indicates the connection they'll use to begin transactions if needed. Also, `AsyncLocalStorage` is used to store transaction data in an `async`/`await` context so we can identify if a connection is being attempted that would break a postgres transaction.

---

### Note for reviewers
This PR appears very large but the core of it is in `src/datastore/connection.ts`. All other changes are just to accomodate the use of this function.

### DIFF
--- a/src/api/routes/bns/namespaces.ts
+++ b/src/api/routes/bns/namespaces.ts
@@ -9,6 +9,7 @@ import {
   getETagCacheHandler,
   setETagCacheHeaders,
 } from '../../../api/controllers/cache-controller';
+import { sqlTransaction } from 'src/datastore/connection';
 
 export function createBnsNamespacesRouter(db: PgStore): express.Router {
   const router = express.Router();
@@ -19,7 +20,7 @@ export function createBnsNamespacesRouter(db: PgStore): express.Router {
     cacheHandler,
     asyncHandler(async (req, res, next) => {
       const includeUnanchored = isUnanchoredRequest(req, res, next);
-      const { results } = await db.getNamespaceList({ includeUnanchored });
+      const { results } = await db.getNamespaceList(db.sql, { includeUnanchored });
       const response: BnsGetAllNamespacesResponse = {
         namespaces: results,
       };
@@ -36,22 +37,30 @@ export function createBnsNamespacesRouter(db: PgStore): express.Router {
       const { tld } = req.params;
       const page = parsePagingQueryInput(req.query.page ?? 0);
       const includeUnanchored = isUnanchoredRequest(req, res, next);
-      const response = await db.getNamespace({ namespace: tld, includeUnanchored });
-      if (!response.found) {
-        res.status(404).json(BnsErrors.NoSuchNamespace);
-      } else {
-        const { results } = await db.getNamespaceNamesList({
-          namespace: tld,
-          page,
-          includeUnanchored,
-        });
-        if (results.length === 0 && req.query.page) {
-          res.status(400).json(BnsErrors.InvalidPageNumber);
+      await sqlTransaction(db.sql, async sql => {
+        const response = await db.getNamespace(sql, { namespace: tld, includeUnanchored });
+        if (!response.found) {
+          throw BnsErrors.NoSuchNamespace;
         } else {
+          const { results } = await db.getNamespaceNamesList(sql, {
+            namespace: tld,
+            page,
+            includeUnanchored,
+          });
+          if (results.length === 0 && req.query.page) {
+            throw BnsErrors.InvalidPageNumber;
+          } else {
+            return results;
+          }
+        }
+      })
+        .then(results => {
           setETagCacheHeaders(res);
           res.json(results);
-        }
-      }
+        })
+        .catch(error => {
+          res.status(400).json(error);
+        });
     })
   );
 
--- a/src/api/routes/bns/namespaces.ts
+++ b/src/api/routes/bns/namespaces.ts
@@ -9,7 +9,7 @@ import {
   getETagCacheHandler,
   setETagCacheHeaders,
 } from '../../../api/controllers/cache-controller';
-import { sqlTransaction } from 'src/datastore/connection';
+import { sqlTransaction } from '../../../datastore/connection';
 
 export function createBnsNamespacesRouter(db: PgStore): express.Router {
   const router = express.Router();
--- a/src/datastore/connection.ts
+++ b/src/datastore/connection.ts
@@ -40,8 +40,21 @@ export async function sqlTransaction<T>(
   }
   const usageName = sql.options.connection.application_name ?? '';
   return sqlTransactionContext.run({ usageName }, () => {
-    return sql.begin(readOnly ? 'read only' : '', callback);
+    return sql.begin(readOnly ? 'read only' : 'read write', callback);
   });
+}
+
+/**
+ * Start a SQL write transaction. See `sqlTransaction`.
+ * @param sql - SQL client
+ * @param callback - Callback with a scoped SQL client
+ * @returns Transaction results
+ */
+export async function sqlWriteTransaction<T>(
+  sql: PgSqlClient,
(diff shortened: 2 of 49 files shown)


src/datastore/connection.ts
```diff
@@ -26,9 +26,11 @@ export async function sqlTransaction<T>(
   readOnly = true
 ): Promise<UnwrapPromiseArray<T>> {
   if ('savepoint' in sql) {
-    return sql.savepoint(callback);
+    // Only the `postgres.TransactionSql` type contains a `savepoint` method, which means we're
+    // already inside a SQL transaction.
+    return callback(sql) as UnwrapPromiseArray<T>;
   } else {
-    return sql.begin(readOnly ? 'READ ONLY' : '', callback);
+    return sql.begin(readOnly ? 'read only' : '', callback);
```
zone117x, line 33: IIUC, this helps prevent/fix situations like:
```ts
sqlTransaction(store.sql, scopedSqlA => {
  scopedSqlA.query(...);
  // some indirection and nested function calls later, an unintentional new sql transaction:
  sqlTransaction(store.sql, scopedSqlB => {
    scopedSqlB.query(...); // localSqlB will now be the same instance as scopedSqlA with this PR
  });
})
```
right?

If so, this seems like it would fix the nested transaction-in-transaction warnings 👍 


I'm curious if there's anything we could do to also prevent scenarios like:
```ts
sqlTransaction(store.sql, scopedSqlA => {
  scopedSqlA.query(...);
  // some indirection and nested function calls later, an unintentional out-of-transaction sql query:
  store.sql.query(...); // oops, using a different sql client instance from the pool inside of the transaction of another sql client instance.. deadlocks? socket data corruption? idk?
  });
})
```

rafaelcr: That's exactly right... I've been looking at options this morning for this and the only concrete idea I've had is to actually **get rid** of `this.sql` altogether.

It would take some refactoring, but by doing that we would open a `sql` object only when we first need to interface with `PgStore` and from there on all the functions inside in the call stack would only have access to their scoped `sql` object preventing these mis-uses.

What do you think?

zone117x: I think removing `this.sql` as an easy to use property makes sense. It's extremely easy to unintentionally reference it when you think you're using a transaction instance, or when you think you're not using one, or not knowing if you're already in one, etc..

What do you think? Something like requiring an explicit function with a callback in order to access to the `sql` client instance -- similar to the `sql.begin(...)` fn?

rafaelcr: Yeah exactly, that could work but we'd need to make sure it also works in a nested kind of way (so we don't run into the same problem i.e. a call to `getNewClient(sql ...)` inside one that already exists).

I'll try some options and let you know

rafaelcr: @zone117x I've pushed a sample of the code I'm currently working on to get your opinion.

In a nutshell, I've added a new `sqlSession` helper here: https://github.com/hirosystems/stacks-blockchain-api/pull/1376/commits/4ed98de8db081eaee672d14a653cd7e8a7ca96a6 that creates a new connection, runs a callback with that SQL object, and closes the connection. The helper also allows us to set a name per session which will make query errors much easier to debug, for example: https://github.com/hirosystems/stacks-blockchain-api/pull/1376/commits/4ed98de8db081eaee672d14a653cd7e8a7ca96a6. I've also removed `this.sql` from `PgStore`.

To absolutely prevent anyone from calling `sqlSession` in a nested fashion, we could add a custom eslint rule (maybe in another PR).

What do you think?

zone117x:
image

src/datastore/connection.ts
zone117x: How long do we expect the `sql` instance inside of the `sqlSession(...)` callback to last? It looks like it's intended usage to not last for very long (e.g. only for the duration of a given http query). If so, we'd be missing out on a lot of benefits provided by connection pooling, such as avoiding the renegotiating connections to pg constantly, and cached prepared statements (pretty sure this cache is associated with the connection lifetime).

zone117x: I'm thinking.. conceptually if this was another language and we used synchronous threads (where each session of sql queries would take place on the same thread), then it would be pretty simple to ensure a "thread local" value that could indicate if a sql transaction is open, and a helper function to open a sql transaction could check this thread local and throw.

I started looking around to see if there was anything like an "AsyncLocal" in js (it's conceptually similar but instead of variable persisted through a thread, it's a variable persisted through an asynchronous flow (a series of async/await calls). For example see C#'s [`AsyncLocal`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=net-6.0).

Then I came across the nodejs API `async_hooks` which I haven't really looked at before. It appears to provide this "AsyncLocal" capability, and their docs even have example code referencing db usage with the API: https://nodejs.org/docs/latest-v16.x/api/async_context.html#usage-with-asyncawait

So it kinda looks like we could use (and potentially even without touching a ton of code) to enforce a given async flow never opens more than one sql transaction at once. What do you think?

src/datastore/connection.ts
```diff
@@ -1,8 +1,61 @@
...
+ * @param readOnly - If a `BEGIN` transaction should be marked as `READ ONLY`
+ * @returns Transaction results
+ */
+export async function sqlTransaction<T>(
+  sql: PgSqlClient,
+  callback: (sql: PgSqlClient) => T | Promise<T>,
+  readOnly = true
+): Promise<UnwrapPromiseArray<T>> {
+  if (sqlTransactionContext.getStore()) {
+    return callback(sql) as UnwrapPromiseArray<T>;
```
zone117x, line 39: I'm a bit uncertain about the typing here -- the return value of `callback` can be `T | Promise<T>`, but `sqlTransaction(...)` returns a Promise, so I think that makes certain usages of this function dependent on the js behavior where technically you can await something that isn't a promise, e.g.
```js
const test = 'abc';
console.log(await test);
// > 'abc'
```

That's probably why the typecast to `UnwrapPromiseArray` is required? Are there usages of the callback that are synchronous and actually need that behavior?

src/api/routes/bns/namespaces.ts
```diff
@@ -19,7 +20,7 @@ export function createBnsNamespacesRouter(db: PgStore): express.Router {
     cacheHandler,
     asyncHandler(async (req, res, next) => {
       const includeUnanchored = isUnanchoredRequest(req, res, next);
-      const { results } = await db.getNamespaceList({ includeUnanchored });
+      const { results } = await db.getNamespaceList(db.sql, { includeUnanchored });
```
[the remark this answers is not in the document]
rafaelcr: Good question, I think we can leverage the async local storage to get rid of all these

PR #1390, opened 2022-10-26
fix: v5 backport BNS ownership for multiple name operations in the same transaction

rafaelcr:
This PR is a v5 backport fix of the issues solved in #1337 for API v6. Since we're still working on deploying v6 in our environment, this patch will allow us to serve correct ownership data in the meantime.

**NOTE:** This only fixes the *owner* of the name, not its *status* e.g. a name's status could still appear as `name-register` when it should be `name-transfer`


zone117x (approved):
LGTM, thanks for the fix!

Issue #1330, opened 2022-09-22
reduce max allowed page size for requests


rafaelcr:
To avoid and/or reduce OOM errors, we should consider reducing the maximum page size we allow in all of our requests.

We don't have a standardized page size across all of our endpoints (some allow 200, others 50, etc.) so we should first make a quick inventory on what we have and then provide one single number we can enforce across all endpoints. We could even make it configurable via an ENV var in case users want to increase these limits in their API nodes.

BLuEScioN:
https://github.com/hirosystems/stacks-blockchain-api/pull/1401/files

PR #1401, opened 2022-10-28
feat(agg-paging-limits): aggregated all paging query limits

BLuEScioN:
Aggregating query limit logic. 

Motivation: https://github.com/hirosystems/stacks-blockchain-api/issues/1330

Eventually the goal is to use data to inform max query limits. This work needs to be planned because it's a bigger bit of work that needs undertaking.

Here is a good breakdown of what will be needed for data analysis https://hiropbc.slack.com/archives/C03UFUA5Q65/p1666816957436129

This change sets query limits based on the resource type being queried, which mostly gets determined from what tables are getting queried. Categorizing queryies in this way meant modyifing some of the existing query limits for some endpoints. The categorization of endpoints and their new  query limits are summarized below:

const pagingQueryLimits: Record<ResourceType, { defaultLimit: number; maxLimit: number }> = {
  [ResourceType.Block]: {
    defaultLimit: 20,
    maxLimit: 30,
  },
  // '/block' 
  [ResourceType.Tx]: {
    defaultLimit: 20,
    maxLimit: 50,
  },
  // '/address/:principal/transactions'
  // '/address/:stx_address/transactions_with_transfers'
  // '/address/:address/mempool'
  // '/address/:address/mempool'
  // '/address/:stx_address/stx_inbound' Changed max from 500 to 50
  // '/tx' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/mempool/dropped' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/mempool' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/multiple' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/events' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/:tx_id' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/block/:block_hash' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/block_height/:height' Changed max from 200 to 50, Changed default from 96 to 20
  // '/tx/mempool' Changed max from 200 to 50, Changed default from 96 to 20
  [ResourceType.Event]: {
    defaultLimit: 20,
    maxLimit: 50,
  },
  // '/address/:stx_address/assets'
  // '/address/:stx_address/nft_events' Deprecated?
  [ResourceType.Burnchain]: {
    defaultLimit: 96,
    maxLimit: 250,
  },
  // '/burnchain/reward_slot_holders'
  // '/burnchain/reward_slot_holders/:address'
  // '/burnchain/rewards'
  // '/burnchain/rewards/:address'
  [ResourceType.Contract]: {
    defaultLimit: 20,
    maxLimit: 50,
  },
  // '/contract/by_trait'
  // '/contract/:contract_id/events'
  [ResourceType.Microblock]: {
    defaultLimit: 20,
    maxLimit: 200,
  },
  //'/microblock': {
  [ResourceType.Token]: {
    defaultLimit: 50,
    maxLimit: 200,
  },
  // '/tokens/nft/holdings'
  // '/tokens/nft/history'
  // '/tokens/nft/mints'
  // '/tokens/ft/metadata' Changed default from 96 to 50
  // '/tokens/nft/metadata' Changed default from 96 to 50
};


src/api/routes/address.ts
```diff
@@ -246,7 +226,7 @@ export function createAddressRouter(db: PgStore, chainId: ChainID): express.Rout
       } else {
         blockHeight = await getBlockHeight(untilBlock, req, res, next, db);
       }
-      const limit = parseTxQueryLimit(req.query.limit ?? 20);
+      const limit = getPagingQueryLimit('/address/:principal/transactions');
```
zone117x, line 229: Nit: would be a bit cleaner to use an `enum` or `const enum` for `pagedApiRoutes`, so that all the other references can avoid duplicating long strings and instead use something like:
```ts
getPagingQueryLimit(PagedApiRoutes.AddressPrincipalTransactions);
```

BLuEScioN: I considered this. Utlimately my feeling is because routes don't translate well to sensible names, ie AddressPrincipalTransactions (You can imagine another being something like AddressPrincipalTransactionsID), they don't really add much value and can even add an element of misdrection when trying to understand the code. Here you can easily see you need to enter the route and what route limit you are fetching. I typed the route parameter so that it's not possible to accidently misspell the route.

zone117x: That sounds reasonable to me. Thanks @BLuEScioN.

src/api/pagination.ts
```diff
@@ -25,16 +25,176 @@ export function parsePagingQueryInput(val: any) {
...
-}
+const MAX_BLOCKS_PER_REQUEST = 30;
+const MAX_TX_PER_REQUEST = 50;
+const MAX_ASSETS_PER_REQUEST = 50;
+const MAX_STX_INBOUND_PER_REQUEST = 500;
+const CONTRACT_MAX_EVENTS_PER_REQUEST = 50;
+const MAX_MICROBLOCKS_PER_REQUEST = 200;
+const MAX_TXS_PER_REQUEST = 200;
+const MAX_MEMPOOL_TXS_PER_REQUEST = 200;
+const TX_MAX_EVENTS_PER_REQUEST = 200;
```
rafaelcr, line 36:
```suggestion
const MAX_TX_EVENTS_PER_REQUEST = 200;
```

src/api/pagination.ts
```diff
@@ -25,16 +25,176 @@ export function parsePagingQueryInput(val: any) {
...
 
-interface ParseLimitQueryParams {
-  maxItems: number;
-  errorMsg: string;
-}
+const MAX_BLOCKS_PER_REQUEST = 30;
+const MAX_TX_PER_REQUEST = 50;
+const MAX_ASSETS_PER_REQUEST = 50;
+const MAX_STX_INBOUND_PER_REQUEST = 500;
+const CONTRACT_MAX_EVENTS_PER_REQUEST = 50;
```
rafaelcr, line 32:
```suggestion
const MAX_CONTRACT_EVENTS_PER_REQUEST = 50;
```

src/api/pagination.ts
```diff
@@ -25,16 +25,176 @@ export function parsePagingQueryInput(val: any) {
...
+type PagedApiRoutes = typeof pagedApiRoutes[number];
 
-export function parseLimitQuery({ maxItems, errorMsg }: ParseLimitQueryParams) {
-  return (val: any) => {
-    const limit = parsePagingQueryInput(val);
-    if (limit > maxItems)
-      throw new InvalidRequestError(errorMsg, InvalidRequestErrorType.invalid_query);
-    return limit;
-  };
+const pagingQueryLimits: Record<PagedApiRoutes, { defaultLimit: number; maxLimit: number }> = {
```
rafaelcr, line 72: I think instead of keeping limits per endpoint, they should be per "return type" e.g.:
* If the endpoint is returning blocks, default limit should be 10
* If it returns transactions, default should be 20
* etc.

BLuEScioN: Hey I have a branch here that implements your feedback (though I haven't changed the calls to get the limits yet. That will be simple and the calls will look pretty much identical). Could you take a look? A few endpoints collapsed nicely into resource categories, but a lot didn't. If one where to add a new endpoint it's not obvious for the resource type approach what default and max limit should be used. Additionally, using an endpoint approach, there's more flexibility to use whatever default and limit we want. I think we should stick with the endpoint approach.

src/api/pagination.ts
```diff
@@ -25,16 +25,58 @@ export function parsePagingQueryInput(val: any) {
...
+  },
+  [ResourceType.Microblock]: {
+    defaultLimit: 20,
+    maxLimit: 200,
+  },
+  [ResourceType.Token]: {
+    defaultLimit: 50,
+    maxLimit: 200,
+  },
+};
```
rafaelcr, line 67: @CharlieC3 what do you think about these limits? They seem reasonable to me but we could probably make the defaults even smaller to increase performance

rafaelcr (approved):
LGTM, great work @BLuEScioN 
I just added Charlie to get his opinion on default page limits

CharlieC3:
@BLuEScioN Were there any query limits that are being _raised_ in this PR? If I understand the description correctly, each query limit type either didn't change or was decreased from its previous setting. Is that correct?

BLuEScioN:
> @BLuEScioN Were there any query limits that are being _raised_ in this PR? If I understand the description correctly, each query limit type either didn't change or was decreased from its previous setting. Is that correct?

"Each query limit type either didn't change or was decreased from its previous setting". Correct.

src/api/pagination.ts
CharlieC3: Considering they're lower than what we currently have and are most familiar with, they look great. I wouldn't want to decrease them drastically in one commit, but I agree it's something we should continue to revisit and adjust.

Issue #1277, opened 2022-08-17
[Stacks 2.1] Support PoX-2 for account STX balance `locked` state


zone117x:
The API currently uses the Stacks 2.0 `stx_lock_event` event to store and retrieve "locked" and "unlock_height" data when responding to account balance requests. This was implemented in https://github.com/hirosystems/stacks-blockchain-api/pull/324 -- the account balance response is currently:
```ts
interface StxBalance {
  balance: string;
  total_sent: string;
  total_received: string;
  total_fees_sent: string;
  total_miner_rewards_received: string;
  /* [PoX] The transaction where the lock event occurred. Empty if no tokens are locked. */
  lock_tx_id: string;
  /* [PoX] The amount of locked STX, as string quoted micro-STX. Zero if no tokens are locked. */
  locked: string;
  /* [PoX] The STX chain block height of when the lock event occurred. Zero if no tokens are locked. */
  lock_height: number;
  /* [PoX] The burnchain block height of when the lock event occurred. Zero if no tokens are locked. */
  burnchain_lock_height: number;
  /* [PoX] The burnchain block height of when the tokens unlock. Zero if no tokens are locked. */
  burnchain_unlock_height: number;
}
```

PoX v2 introduces several new features, some already merged:
* https://github.com/stacks-network/stacks-blockchain/pull/2755

Some WIP:
* https://github.com/stacks-network/stacks-blockchain/issues/2534
* https://github.com/stacks-network/stacks-blockchain/issues/2533
* https://github.com/stacks-network/stacks-blockchain/issues/2556

It doesn't look like there are any new events (similar to `stx_lock_event`) for these new PoX operations -- can/will the various new operations just emit another `stx_lock_event` event. If so, does it make sense for the API use the same event to query for the latest `locked` state when returning stx balance data?

kantai:
Yes, these new operations can emit events.

Would the most useful thing be to emit something like a `StxBalanceUpdate` event that just emitted the new state of the balance. Something like:

```js
{
  /* The current available balance, as string quoted micro-STX */
  balance: string,
  /* [PoX] The amount of locked STX, as string quoted micro-STX. Zero if no tokens are locked. */
  locked: string,
  /* [PoX] The burnchain block height of when the tokens unlock. Zero if no tokens are locked. */
  burnchain_unlock_height: number,
}
```

[...]

PR #1403, opened 2022-10-29
Ingestion and querying for new PoX-2 events

zone117x:
* Implemented parsing for the new PoX-2 print events in the event-stream handler.
* Pg schema, writing, and reading of the new events.
* Integration tests for all PoX-2 operations required to trigger each event type.
* Account balance locked state validation tests, ensures API and stacks-node have matching balance & locked state after each of the various Stacking operations and through each of the pox transition periods as defined in SIP-15.
* Update API account balance/locked queries to use the new events -- closes https://github.com/hirosystems/stacks-blockchain-api/issues/1277
* Tests Stacking with a segwit address via Rosetta -- completes one of the remaining Rosetta issues described at https://github.com/hirosystems/stacks-blockchain-api/issues/1278#issuecomment-1265768156
* Synthetic Rosetta stx-unlock operations now generated for both the `pox_v1_unlock_height` transition, and the new PoX-v2 auto-unlocks, completes another remaining Rosetta issue described at https://github.com/hirosystems/stacks-blockchain-api/issues/1278#issuecomment-1265768156

TODO:
* [x] Some new endpoints were implemented to expose the events for testing purposes -- these should probably be removed and the tests can query the db directly
  * EDIT: These endpoints are left in for now, they will likely be useful to allow stacking pool operators to monitor for various new pox-2 operations (e.g. delegate stack increases). They still need openapi/schema documentation but that can be done in a later PR.


src/pox-helpers.ts
```diff
@@ -221,31 +247,360 @@ function networkToHrp(network: 'mainnet' | 'testnet' | 'regtest'): string {
...
+    return `${principal.address}.${principal.contract_name}`;
+  }
+  throw new Error(
+    `Unexpected Clarity value type for principal: ${(principal as ClarityValue).type_id}`
+  );
+}
+
+// TODO: this and the logic referencing it can be removed once the "stale" data issue is fixed in
+// https://github.com/stacks-network/stacks-blockchain/pull/3318
+const PATCH_EVENT_BALANCES = true;
```
zone117x, line 411: Note for reviewers:
The pox2 events emitted are currently bugged with an issue where the locked state (balance, locked amount, unlock height) are all "stale" and report the state from _before_ the transaction. The code checking this value is performing a hacky workaround where operation-specific args are used to "patch" the stale values into the correct state.

zone117x: It was decided to keep this behavior, at least for now.

src/api/routes/pox2.ts
```diff
@@ -0,0 +1,84 @@
...
+      const response = {
+        limit,
+        offset,
+        results: parsedResult,
+      };
+      res.json(response);
+    })
+  );
+
+  // TODO: this should probably be a tx route e.g. /extended/v1/tx/:tx_id/pox2_events
```
rafaelcr, line 45: I agree, could we make this change? Maybe in a later PR if it's too much work

src/api/routes/pox2.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+      const parsedResult = queryResults.result.map(r => parsePox2Event(r));
+      const response = {
+        results: parsedResult,
+      };
+      res.json(response);
+    })
+  );
+
+  // TODO: this should probably be an account route e.g. /extended/v1/address/:stx_address/pox2_events
```
rafaelcr, line 64: Same here

src/api/routes/status.ts
```diff
@@ -15,6 +15,12 @@ export function createStatusRouter(db: PgStore): express.Router {
         server_version: `stacks-blockchain-api ${API_VERSION.tag} (${API_VERSION.branch}:${API_VERSION.commit})`,
         status: 'ready',
       };
+      const pox1UnlockHeight = await db.getPox1UnlockHeight();
+      if (pox1UnlockHeight.found) {
+        response.pox_v1_unlock_height = pox1UnlockHeight.result;
+      } else {
+        response.pox_v1_unlock_height = null as any;
+      }
```
rafaelcr, line 23: Is this change temporary? i.e. will we need to remove this once it's no longer relevant? Or will it disappear once this height is reached?

migrations/1666703991492_pox_events.js
```diff
@@ -0,0 +1,190 @@
...
+        reward_cycle IS NOT NULL AND 
+        amount_ustx IS NOT NULL
+      WHEN 'stack-aggregation-increase' THEN
+        reward_cycle IS NOT NULL AND 
+        amount_ustx IS NOT NULL
+      ELSE false
+    END
+  )`);
+
+  pgm.createIndex('pox2_events', 'block_height');
```
rafaelcr, line 174: This would be much more performant as 
```js
pgm.createIndex('txs', [
  { name: 'microblock_sequence', sort: 'DESC' },
  { name: 'event_index', sort: 'DESC' },
]);
  ```

File: src/datastore/pg-store.ts [hunk 6]
```diff
@@ -1909,6 +1948,62 @@ export class PgStore {
...
+        LIMIT ${limit}
+        OFFSET ${offset}
+      `;
+      const result = queryResults.map(result => parseDbPox2Event(result));
+      return result;
+    });
+  }
+
+  async getPox2EventsForTx({ txId }: { txId: string }): Promise<FoundOrNot<DbPox2Event[]>> {
+    // TODO: this query should be performed in the same sql tx as the next
```
rafaelcr, line 1973: Once you merge from `develop` this will be simple with `sqlTransaction`

rafaelcr (commented):
Initial review. Looks very good so far but I still need to go over the tests.

src/api/routes/pox2.ts
zone117x: Yeah, it's not super clear yet what the interface for exposing these should look like just yet. I think it makes sense to figure that out in a later PR.

migrations/1666703991492_pox_events.js
zone117x: Good call, done!

rafaelcr (approved):
LGTM, great work 🔥 

The only thing that concerns me a bit is that we're mixing Rosetta tests with regular tests and it will make things a bit harder for us when separating Rosetta to a microservice, but we can worry about that later 👍

PR #1410, opened 2022-11-03
fix: sql transaction consistency

rafaelcr:
This PR fixes the use of postgres `BEGIN`/`COMMIT` transactions across the entire API

### Background
For code reusability, we have functions within `PgStore` that each perform a series of queries against the DB within their own `BEGIN`/`COMMIT` transaction blocks for result consistency. We then group some of these function calls inside other utility functions (e.g. in `db-controller.ts`) and use them in different parts of the API like endpoints, tests, Rosetta, etc.

This creates several problems:

1. If SQL transactions are performed in series without an enclosing transaction, results may be inconsistent or wrong if a new block is being written to the DB at the same time (causing #1370)
2. If SQL transactions are nested (i.e. a function that opens a transaction and calls another function inside which opens another transaction) it can lead to an incorrect use of SQL connections and to `WARNING: there is already a transaction in progress` logs appearing in the DB (probably causing #1375)

### Solution
This PR creates a new utility function called `sqlTransaction` which allows us to use and nest transaction blocks wherever we want.

When used, this function checks the SQL connection to see if it was already scoped to a transaction. If so, it runs queries directly instead of calling `BEGIN`. This allows us to keep consistent transaction use and to return correct results in all of our endpoints.

This PR also uses `AsyncLocalStorage` to share a single SQL connection inside a transaction, regardless if the code calls to `this.sql` inside.

### DIFF
--- a/.env
+++ b/.env
@@ -9,6 +9,8 @@ PG_SSL=false
 # PG_IDLE_TIMEOUT=30
 # Max connection lifetime in seconds, defaults to 60
 # PG_MAX_LIFETIME=60
+# Seconds before force-ending running queries on connection close, defaults to 5
+# PG_CLOSE_TIMEOUT=5
 
 # Can be any string, use to specify a use case specific to a deployment
 PG_APPLICATION_NAME=stacks-blockchain-api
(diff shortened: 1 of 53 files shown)


rafaelcr:
@zone117x FYI the `sqlTransaction` function was moved to `pg-store.ts` to make it easier to use

zone117x (approved):
LGTM, thanks @rafaelcr!

Issue #1275, opened 2022-08-17
[Stacks 2.1] Support `coinbase-pay-to-alt-recipient` transactions


zone117x:
Support `coinbase-pay-to-alt-recipient` transactions, see https://github.com/stacks-network/stacks-blockchain/pull/3164

**Blocked** by bug in the stacks-node, see https://github.com/stacks-network/stacks-blockchain/issues/3258

zone117x:
Support implemented in https://github.com/hirosystems/stacks-blockchain-api/pull/1283

However, STX balances will be reported incorrectly for miners who use the new `coinbase-pay-to-alt` due to https://github.com/stacks-network/stacks-blockchain/issues/3258

Once that bug is fixed on the Stacks-node then it should also be fixed in the API without additional changes.

zone117x:
Closed with https://github.com/hirosystems/stacks-blockchain-api/pull/1413 + https://github.com/stacks-network/stacks-blockchain/pull/3373

PR #1413, opened 2022-11-04
feat: store new Stacks block miner address value from rewards event

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1275

PR https://github.com/stacks-network/stacks-blockchain/pull/3373 fixed issue [Incorrect matured_miner_rewards for CoinbasePayToAltRecipient ](https://github.com/stacks-network/stacks-blockchain/issues/3258), and also introduced the new property `miner_address` to the `miner_rewards` event payload to indicate the principal that mined the block.

This PR stores and queries that value (in a <2.1 backwards compatible way) so that later sql queries are able to use it to still easily query for things like unique miners. The `miner_rewards` sql column will be null until stacks-node 2.1 is deployed, and the insert/query code fills in the null value with the `recipient` principal which is accurate behavior prior to 2.1.


rafaelcr (approved):
LGTM 👍

PR #1427, opened 2022-11-09
fix: handle websocket messages with a priority queue

rafaelcr:
We currently process and send all web socket messages synchronously and in series, which may be causing OOM errors in node.js when a huge anchor block comes that tries to send potentially hundreds of messages to dozens of subscribers.

This PR:
* Adds a priority queue to handle incoming web socket messages from API read-only nodes, with a concurrency of `1` and a max job timeout of 5 seconds (e.g. 5 seconds to send a single `block` update to all block subscribers).
* Adds configurable 5 second timeouts to:
  * Client ping/pong
  * Each message sent to each client
  * Per update enqueued to all clients
* Changes the ws-channel message send operations to be processed in parallel, so we can dispatch one message to all channels immediately and we can better use the 5 seconds we have available.

cc @CharlieC3

### DIFF
--- a/src/api/routes/ws/channels/socket-io-channel.ts
+++ b/src/api/routes/ws/channels/socket-io-channel.ts
@@ -22,7 +22,7 @@ import {
  * SocketIO channel for sending real time API updates.
  */
 export class SocketIOChannel extends WebSocketChannel {
-  private io?: SocketIOServer<ClientToServerMessages, ServerToClientMessages>;
+  private io?: SocketIOServer<ClientToServerMessages, ServerToClientMessages<true>>;
   private adapter?: Adapter;
 
   constructor(server: http.Server) {
@@ -33,11 +33,14 @@ export class SocketIOChannel extends WebSocketChannel {
   }
 
   connect(): void {
-    const io = new SocketIOServer<ClientToServerMessages, ServerToClientMessages>(this.server, {
-      cors: { origin: '*' },
-      pingInterval: 5_000,
-      pingTimeout: 5_000,
-    });
+    const io = new SocketIOServer<ClientToServerMessages, ServerToClientMessages<true>>(
+      this.server,
+      {
+        cors: { origin: '*' },
+        pingInterval: 5_000,
+        pingTimeout: 5_000,
+      }
+    );
     this.io = io;
 
     io.on('connection', async socket => {
@@ -155,6 +158,21 @@ export class SocketIOChannel extends WebSocketChannel {
     return false;
   }
 
+  private async getTopicSockets(room: Topic) {
+    if (!this.io) {
+      return;
+    }
+    const sockets = [];
+    const socketIds = await this.io.to(room).allSockets();
+    for (const id of socketIds) {
+      const socket = this.io.sockets.sockets.get(id);
+      if (socket) {
+        sockets.push(socket);
+      }
+    }
+    return sockets;
+  }
+
   send<P extends keyof WebSocketPayload>(
     payload: P,
     ...args: ListenerType<WebSocketPayload[P]>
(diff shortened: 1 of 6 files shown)


CharlieC3:
> Adds a priority queue to handle incoming web socket messages from API write-only nodes

@rafaelcr Do you mean to say read-only here? We're only seeing this issue occur in pods exposed to high load (read-only)

rafaelcr:
Whoops typo @CharlieC3

src/api/routes/ws/web-socket-transmitter.ts
```diff
@@ -23,23 +24,45 @@ export class WebSocketTransmitter {
...
+  private queue: PQueue;
 
   constructor(db: PgStore, server: http.Server) {
     this.db = db;
     this.server = server;
+    // This queue will send all messages through web socket channels, one at a time.
+    this.queue = new PQueue({
+      autoStart: true,
+      concurrency: 1,
+      timeout: 10_000, // 10 seconds
```
CharlieC3, line 36: How does this timeout function? Does it mean 10 seconds in total to send updates to all registered sockets? Or 10 seconds per sockets? etc

CharlieC3 (approved):
LGTM!

src/api/routes/ws/web-socket-transmitter.ts
```diff
@@ -23,23 +24,45 @@ export class WebSocketTransmitter {
...
+      throwOnTimeout: true,
```
zone117x, line 37: I'm thinking through the implications of this, and it seems like a client (either malicious or just really slow network) could make a bunch of subscriptions that take a long time, say the full 10 second timeout, and effectively DDoS?

CharlieC3: Hmm that's a good point. The queue is a good way to protect itself against limited DDoS's in this way, but could still delay legitimate subscriptions from getting their updates in a major way. Perhaps the timeout should be shorter, like 1 second? Is there any update that would reasonably take longer than that?

Another change that would safeguard against this is to set a max number of subscriptions an API should logically have. However, without a shared state of socket subscriptions across all APIs, this would result in any new subscriptions over the max being denied. With a shared state of socket subscriptions across all APIs, they could redistribute the load and we could actually autoscale based on this metric. Food for thought in future designs perhaps.

rafaelcr: Good idea. I'm also going to double check the current server timeout behavior on WS and Socket IO i.e. there should be a timeout that automatically drops a client if a ping is not responded with a pong within a few seconds.

zone117x: It still sounds like I could craft an annoying client that responds to websocket pings in a timely fashion, but take a long time to receive each subscription message, then clog up all API deployments with a ton of subscriptions and result in no one getting subscription message. 

Maybe, in combination with your above changes, also limit the number of ws connections/subscriptions per client/IP?

rafaelcr: @zone117x check it out, this commit adds a per-emit timeout for socket.io... if 5 seconds are reached, the socket will be disconnected https://github.com/hirosystems/stacks-blockchain-api/pull/1427/commits/e76b2b1c608283ffa53cfba3ddb7cf09446a2e30

rafaelcr:
@CharlieC3 @zone117x I added a couple of commits to this PR, in summary:

* Add single priority queue for all messages.
* Set 5 second timeouts for:
    * Client ping/pong
    * Each message sent to each client
    * Per update enqueued to all clients

If any timeout is reached, the offending client is dropped.

CharlieC3:
Do we happen to have unit tests for these additions? If not, how easy would it be to add them to this PR?

rafaelcr:
@CharlieC3 ping timeout and message disconnect tests added in this commit: https://github.com/hirosystems/stacks-blockchain-api/pull/1427/commits/4900984996643ef2b9742625e08067217ee7f660

CharlieC3 (approved):
LGTM! Thanks for adding in the tests. Not needed for the merge into `beta`, but can we document the three new env vars in the `.env` or README before merging into `main`?
`STACKS_API_WS_PING_INTERVAL`, `STACKS_API_WS_PING_TIMEOUT`, `STACKS_API_WS_MESSAGE_TIMEOUT`

rafaelcr:
For sure, good catch! Done @CharlieC3

src/api/routes/ws/channels/socket-io-channel.ts
```diff
@@ -153,74 +163,126 @@ export class SocketIOChannel extends WebSocketChannel {
...
+    // If a client takes more than this number of ms to respond to an event `emit`, it will be
+    // disconnected.
+    const timeout = getWsMessageTimeoutMs();
     switch (payload) {
       case 'block': {
         const [block] = args as ListenerType<WebSocketPayload['block']>;
         this.prometheus?.sendEvent('block');
-        this.io.to('block').emit('block', block);
+        void this.getTopicSockets('block').then(sockets =>
+          sockets?.map(s => s.timeout(timeout).emit('block', block, _ => s.disconnect(true)))
```
zone117x, line 196: Is `map` used here for a reason? It seems like the usage is more inline with a `forEach`?

zone117x (commented):
PR is looking great, thanks @rafaelcr! Left a couple more questions, apologies for the nits!

src/api/routes/ws/channels/socket-io-channel.ts
rafaelcr: Nope, nice catch too, done!

zone117x (approved):

Issue #1327, opened 2022-09-21
Not possible to import BNS data without first syncing


wileyj:
From conversation with @rafaelcr and related issue: https://github.com/stacks-network/stacks-blockchain-docker/issues/90

With the release of v5.0.0, `BNS_IMPORT_DIR` is no longer used during starting the API normally. 
In previous versions of the API, with this env var defined - the API would process the BNS data and add DB records. 

In v5.0.0, i understand this work is being offloaded to the event-replay function. In my testing, i noticed that starting the API normally alongside postgres and a stacks-blockchain node (with `STACKS_EXPORT_EVENTS_FILE` defined) works as expected and the tsv file is populated with blocks. 
Stopping the API after a few blocks, then running the event-replay allows the BNS data to be imported normally. 

What doesn't work (and i think this would be the use case for most people) is importing BNS data **before** running the API normally (i.e. syncing from genesis). 
1. if  you haven't started the API, no tsv file exists and the `import-events` function will fail on file not found (expected)
2. using an empty tsv file gets past the previous error, and i was expecting the BNS import to proceed but it does not: 
```
{12:07}~/Git/blockstack/public/stacks-blockchain-api:master ✗ ➭ touch file.tsv
node ./lib/index.js import-events \
    --file file.tsv --wipe-db --force
{"level":"info","message":"Shutting down... received beforeExit.","timestamp":"2022-09-21T16:07:44.961Z"}
{"level":"info","message":"App shutdown successful.","timestamp":"2022-09-21T16:07:44.963Z"}
```

tl;dr - it's not possible to import BNS data until you've started the API normally and synced a few blocks. I believe it should be possible via cmd-line switch or natively to import this BNS data before you start syncing the API.

zone117x:
Makes sense. 

One of the reasons for this change was to get the genesis block info required for correctly importing BNS, which we can't know if importing from scratch. 

We still have the `config_state` table with some global singleton values in order to track state like this.

Perhaps we could add a new field like `bns_import_needs_genesis: boolean`, which we could use to first import BNS data with empty genesis block info (setting `bns_import_needs_genesis=true`), then on genesis block receipt, we can check `bns_import_needs_genesis` to determine if the "partially" inserted BNS data needs a giant UPDATE.

@rafaelcr thoughts?

rafaelcr:
That's a good idea @zone117x ... we should aim for a solution that works both for event replay and genesis sync, and this sounds like it could fit the bill.

[...]

PR #1441, opened 2022-11-15
Nickb/fix bns import

BLuEScioN:
Addressing the issue here https://github.com/hirosystems/stacks-blockchain-api/issues/1327

As far as I can understand, the user wants to be able to sync bns data not just from running event-replay but also from a genesis sync.

Solution:

Add logic to the /new_block handler that checks if the block is the genesis block, which we need to do a correct bns import. Once we have that block, we can check if we should import bns data based on the config_state. If we haven't, we run the import bns and subdomain bns data

Tests:

Post a genesis block and then check if the config_state confirms that the bns data has been imported correctly.

### DIFF
--- a/.env
+++ b/.env
@@ -140,3 +140,5 @@ MAINNET_SEND_MANY_CONTRACT_ID=SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.send-man
 # of openapi.yaml for test / development NODE_ENV. 
 # For production, /doc is not served if this env var is not provided.
 # API_DOCS_URL="https://docs.hiro.so/api" 
+
+TSV_IMPORT_IN_PROCESS=false
--- a/.env
+++ b/.env
@@ -141,4 +141,4 @@ MAINNET_SEND_MANY_CONTRACT_ID=SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.send-man
 # For production, /doc is not served if this env var is not provided.
 # API_DOCS_URL="https://docs.hiro.so/api" 
 
-TSV_IMPORT_IN_PROCESS=false
+# TSV_IMPORT_IN_PROCESS=false
--- a/.env
+++ b/.env
@@ -9,6 +9,8 @@ PG_SSL=false
 # PG_IDLE_TIMEOUT=30
 # Max connection lifetime in seconds, defaults to 60
 # PG_MAX_LIFETIME=60
+# Seconds before force-ending running queries on connection close, defaults to 5
+# PG_CLOSE_TIMEOUT=5
 
 # Can be any string, use to specify a use case specific to a deployment
 PG_APPLICATION_NAME=stacks-blockchain-api
@@ -33,6 +35,7 @@ PG_APPLICATION_NAME=stacks-blockchain-api
 # PG_PRIMARY_SSL=
 # PG_PRIMARY_IDLE_TIMEOUT=
 # PG_PRIMARY_MAX_LIFETIME=
+# PG_PRIMARY_CLOSE_TIMEOUT=
 # The connection URI below can be used in place of the PG variables above,
 # but if enabled it must be defined without others or omitted.
 # PG_PRIMARY_CONNECTION_URI=
@@ -84,6 +87,9 @@ STACKS_CORE_RPC_PORT=20443
 ## configure the chainID/networkID; testnet: 0x80000000, mainnet: 0x00000001
 STACKS_CHAIN_ID=0x00000001
 
+# Seconds to allow API components to shut down gracefully before force-killing them, defaults to 60
+# STACKS_SHUTDOWN_FORCE_KILL_TIMEOUT=60
+
 BTC_RPC_HOST=http://127.0.0.1
 BTC_RPC_PORT=18443
 BTC_RPC_USER=btc
@@ -91,7 +97,7 @@ BTC_RPC_PW=btc
 BTC_FAUCET_PK=29c028009a8331358adcc61bb6397377c995d327ac0343ed8e8f1d4d3ef85c27
 
 # The contracts used to query for inbound transactions
-TESTNET_SEND_MANY_CONTRACT_ID=STR8P3RD1EHA8AA37ERSSSZSWKS9T2GYQFGXNA4C.send-many-memo
+TESTNET_SEND_MANY_CONTRACT_ID=ST3F1X4QGV2SM8XD96X45M6RTQXKA1PZJZZCQAB4B.send-many-memo
 MAINNET_SEND_MANY_CONTRACT_ID=SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.send-many-memo
 
 # Enable debug logging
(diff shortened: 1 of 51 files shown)


src/event-replay/helpers.ts
```diff
@@ -42,36 +44,57 @@ export async function findTsvBlockHeight(filePath: string): Promise<number> {
  * @returns Genesis block data
  */
 export async function findBnsGenesisBlockData(filePath: string): Promise<BnsGenesisBlock> {
+  const genesisBlockMessage = await getGenesisBlockData(filePath);
+  const bnsGenesisBlock = getBnsRegistrationEvent(genesisBlockMessage);
+  return bnsGenesisBlock;
+}
+
+export async function getGenesisBlockData(filePath: string): Promise<CoreNodeBlockMessage> {
```
rafaelcr, line 52: A more explicit name would be best
```suggestion
export async function getGenesisBlockDataFromTsv(filePath: string): Promise<CoreNodeBlockMessage> {
```

src/event-replay/helpers.ts
```diff
@@ -42,36 +44,57 @@ export async function findTsvBlockHeight(filePath: string): Promise<number> {
...
         }
       }
     }
+  } finally {
+    rl.close();
+  }
+  throw new Error('Genesis block data not found');
+}
+
+export function getBnsRegistrationEvent(
```
rafaelcr, line 74:
```suggestion
export function getBnsGenesisBlockFromBlockMessage(
```

src/event-stream/event-server.ts
```diff
@@ -743,12 +745,25 @@ export async function startEventServer(opts: {
...
   });
 
+  const handleBnsImport = async (blockMessage: CoreNodeBlockMessage) => {
+    const bnsDir = process.env.BNS_IMPORT_DIR;
+    if (blockMessage.block_height === 1 && bnsDir) {
+      const configState = await db.getConfigState();
+      if (!configState.bns_names_onchain_imported && !configState.bns_subdomains_imported) {
+        const bnsRegistrationEvent = getBnsRegistrationEvent(blockMessage);
+        await importV1BnsNames(db, bnsDir, bnsRegistrationEvent);
+        await importV1BnsSubdomains(db, bnsDir, bnsRegistrationEvent);
```
rafaelcr, line 755: A log message before these two imports would be useful for debugging

rafaelcr (commented):
Looking good so far... I wonder if we could standardize this new process with the event replay so that they happen in the same manner i.e. import BNS names after block 1 is processed either via TSV or genesis sync.

Can you see if that change would fit in this PR?

src/event-replay/helpers.ts
BLuEScioN: Yes, I believe this does cover both because on event replay, each new block event will run handleBnsImport in the /new_block handler. I'll right another test for this specifically

src/event-replay/helpers.ts
```diff
@@ -4,6 +4,8 @@ import * as readline from 'readline';
 import { decodeTransaction, TxPayloadTypeID } from 'stacks-encoding-native-js';
 import { DataStoreBnsBlockData } from '../datastore/common';
 import { ReverseFileStream } from './reverse-file-stream';
+import { CoreNodeBlockMessage } from '../event-stream/core-node-message';
+import { rawListeners } from 'process';
```
rafaelcr, line 8: I think this import is unused

.env
```diff
@@ -162,3 +162,5 @@ MAINNET_SEND_MANY_CONTRACT_ID=SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.send-man
 # of openapi.yaml for test / development NODE_ENV. 
 # For production, /doc is not served if this env var is not provided.
 # API_DOCS_URL="https://docs.hiro.so/api" 
+
+# TSV_IMPORT_IN_PROCESS=false
```
rafaelcr, line 166: Since this ENV is for internal use only we shouldn't expose it here (maybe users can get confused and set it to true can cause some issues)

src/event-stream/event-server.ts
```diff
@@ -743,12 +745,27 @@ export async function startEventServer(opts: {
...
+
   app.post(
     '/new_block',
     asyncHandler(async (req, res, next) => {
       try {
-        const msg: CoreNodeBlockMessage = req.body;
-        await messageHandler.handleBlockMessage(opts.chainId, msg, db);
+        const blockMessage: CoreNodeBlockMessage = req.body;
+        await messageHandler.handleBlockMessage(opts.chainId, blockMessage, db);
+        await handleBnsImport(blockMessage);
```
rafaelcr, line 768: Have you tested locally that this works as expected? The code looks correct but I worry that because this BNS import process will take a long time (~5 minutes), the Stacks node will be left waiting for a response from the API acknowledging the block message (because you're `await`ing here) and decide to declare a timeout, dropping the message or worse trying again (which would re-trigger the same functionality).

@zone117x maybe has more context here on the core node behavior.

Maybe a better idea would be **not** to `await` here and just let the promise run in parallel, but we'd need to verify if this doesn't cause any problems.

BLuEScioN: That sounds right to me. We can make bnsimport middleware that runs after the /new_block handler. That way the req will be handled the same as before this change, sending the client a response as soon as the handleblockmessage completes.

rafaelcr (approved):
LGTM!

src/event-stream/event-server.ts
```diff
@@ -678,6 +680,34 @@ export type EventStreamServer = net.Server & {
...
 };
 
+export const bnsImportMiddleware = (db: PgWriteStore) => {
+  let bns_names_onchain_imported: boolean = false;
+  let bns_subdomains_imported: boolean = false;
+  return asyncHandler(async (req, res, next) => {
+    const blockMessage: CoreNodeBlockMessage = req.body;
+    const bnsDir = process.env.BNS_IMPORT_DIR;
+    if (!bns_names_onchain_imported || !bns_subdomains_imported) {
+      const configState = await db.getConfigState();
```
zone117x, line 690: It looks like this function can be simplified. On first review, it appeared to be performing a sql query on every new block, and had to re-read the function to see a bit of indirection used for those boolean flags to skip the check. 

Is there a reason why this can't just check `blockMessage.block_height === 1` first (an inexpensive check that will only be hit once), then perform the other checks like this sql query?

src/event-stream/event-server.ts
zone117x: Hmm, I'm not sure what's more error prone: A) blocking the event-observer response for longer than usual, or B) running the BNS insertion sql at the same time as later new blocks are being ingested with its sensitive sql code.

With either approach, we should test that the BNS import completes and the first few blocks are ingested while using the mainnet dataset. This could be tested locally fairly quickly. @BLuEScioN is this something you can test to confirm nothing breaks?

src/event-stream/event-server.ts
```diff
@@ -52,7 +52,7 @@ import {
   ClarityValueTuple,
   TxPayloadTypeID,
 } from 'stacks-encoding-native-js';
-import { ChainID } from '@stacks/transactions';
+import { ChainID, nextSignature } from '@stacks/transactions';
```
zone117x, line 55: Used import?

src/tests-event-replay/import-export-tests.ts
```diff
@@ -95,7 +99,11 @@ describe('import/export tests', () => {
...
   });
 
-  afterEach(async () => {
-    await db?.close();
+  test('Bns import occurs', async () => {
+    process.env.BNS_IMPORT_DIR = 'src/tests-bns/import-test-files';
+    await importEventsFromTsv('src/tests-event-replay/tsv/mocknet.tsv', 'archival', true, true);
+    const configState = await db.getConfigState();
+    expect(configState.bns_names_onchain_imported).toBe(true);
+    expect(configState.bns_subdomains_imported).toBe(true);
```
zone117x, line 107: Do any of these tests validate that the imported names are correct?

src/event-stream/event-server.ts
BLuEScioN: I have a PR to add import sorting and removal of unused imports on save. @rafaelcr @zone117x would you be interested in that?

src/tests-event-replay/import-export-tests.ts
BLuEScioN: No, I can look into that, but I figured I didn't make any changes to the import process and that there would be other tests in place for this

src/event-stream/event-server.ts
BLuEScioN: Yes that should be first

src/event-stream/event-server.ts
zone117x: Sure, that should be a separate PR without any other changes, since it will cause merge conflicts for all other pending branches.

src/event-stream/event-server.ts
BLuEScioN: Going to skip the closure. First check if height = 1 and we have bnsDir. We should only encounter this once. Then query for import states. If we havent imported one or the other, run import functions.

src/tests-event-replay/import-export-tests.ts
BLuEScioN: We have the v1 import test and additionally the config wouldnt update if any part of the import process failed so we should be good!

zone117x (approved):
LGTM, thanks @BLuEScioN!

PR #1453, opened 2022-11-23
fix: avoid selecting `raw_tx` column on read queries

rafaelcr:
This PR removes the `raw_tx` field from all read queries in the API but keeps it for write queries so that this data is still written to the DB.

It is not a breaking change since the API never returns the `raw_tx` column in any endpoint except for `/extended/v1/tx/{tx_id}/raw` which will stay the same.


zone117x (approved):
LGTM! But are we intending for this to go into `master`? Doesn't seem like an urgent hotfix release is warranted for this?

rafaelcr:
@zone117x I think I'll better switch to the `beta` branch for this so we can follow that route instead

PR #1454, opened 2022-11-25
fix: add automatic pox switching to rosetta

janniks:
## More PoX Transition Tests

- adds @stacks/stacking library from Stacks.js
- refactors test helpers

* adds rosetta auto-pox-contract switching
* adds rosetta btc address tests

- adds some additional general pox-transitions tests

---

**Node issues found:** 
- The reported `/v2/pox` contract_id is wrong. The switch to `.pox-2` as "active" (`.contract_id`) happens 1 block early. According to SIP-015 and related discussions, it should happen on `v1_unlock_height + 1` and NOT `v1_unlock_height` itself. https://github.com/stacks-network/stacks-blockchain/issues/3422
- Stacking to both pox-1 and pox-2 from same account in Period 2a triggers a node panic https://github.com/stacks-network/stacks-blockchain/issues/3423


src/helpers.ts
```diff
@@ -508,11 +509,11 @@ export function getCurrentGitTag(): string {
  * Encodes a buffer as a `0x` prefixed lower-case hex string.
  * Returns an empty string if the buffer is zero length.
  */
-export function bufferToHexPrefixString(buff: Buffer): string {
+export function bufferToHexPrefixString(buff: Uint8Array): string {
   if (buff.length === 0) {
     return '';
   }
-  return '0x' + buff.toString('hex');
+  return '0x' + bytesToHex(buff);
```
zone117x, line 516: This function is called in a few hot paths, we should keep the node.js `Buffer` usage unless comprehensive benchmarking can show that this `byteToHex` function doesn't result in any regressions to CPU or memory usage.

src/helpers.ts
```diff
@@ -1022,7 +1023,7 @@ export function bnsHexValueToName(hex: string): string {
   const tuple = hexToCV(hex) as TupleCV;
   const name = tuple.data.name as BufferCV;
   const namespace = tuple.data.namespace as BufferCV;
-  return `${name.buffer.toString('utf8')}.${namespace.buffer.toString('utf8')}`;
+  return `${bytesToUtf8(name.buffer)}.${bytesToUtf8(namespace.buffer)}`;
```
zone117x, line 1026: Same here

src/test-utils/test-helpers.ts
```diff
@@ -1,34 +1,68 @@
...
+  TransactionSigner,
   TransactionVersion,
-  tupleCV,
   TupleCV,
+  tupleCV,
 } from '@stacks/transactions';
-import { AddressStxBalanceResponse, ServerStatusResponse } from 'docs/generated';
-import { testnetKeys } from '../api/routes/debug';
+import { RPCClient } from 'rpc-bitcoin';
+import { getRosettaNetworkName, RosettaConstants } from 'src/api/rosetta-constants';
```
zone117x, line 36: Import path here to should relative

src/helpers.ts
janniks: Got it. Is it ok to keep `Uint8Array` here, or should the underlying data also stay in `Buffer`?

zone117x: Needs to be `Buffer` since that is what has the `.toString('hex')` capability

janniks: I mean, would wrapping the input Uint8Array in a `Buffer.from(buff)` work? Is the concern general regressions (from switching types) or specifically `bytesToHex`?

I think I'll revert the change for now and maybe keep two versions of stacks.js — otherwise some vars will have to be wrapped with either `Uint8Array.from` or `Buffer.from`.

zone117x: > would wrapping the input Uint8Array in a Buffer.from(buff) work?

Can you do that in the calling code? Most of the API codebase is using `Buffer` instances, and we don't want to incur any unnecessary memory or CPU penalties from constructing new `Buffer` instances and/or using `bytesToHex` rather than `Buffer.toString('hex')`.

I don't think you'd need `Uint8Array.from` anywhere, because `Buffer` is an instance of `Uint8Array`.

janniks: Yep, sounds good.

src/test-utils/test-helpers.ts
janniks: Fixed

zone117x (approved):
Great work @janniks! Nice job uncovering those issues 💯 

Thanks again for the help.

PR #1462, opened 2022-11-30
feat: add `smartContractUpdate` and `smartContractLogUpdate` to `PgWriteStore` event emitter

### LINKED ISSUE #1460: add event emitter events for `smartContractUpdate` and `smartContractLogUpdate`
the Token Metadata Service requires these events to be emitted so it can listen for new token contract deployments and support SIP-019's metadata update notifications

rafaelcr:
Adds internal events (event emitter and postgres `NOTIFY`) for when:
1. A smart contract is deployed
2. A smart contract log event is produced

These events will be emitted in blocks and microblocks.

Also, tweak some event interfaces so we get more type safety when emitting events.

Fixes #1460


zone117x (approved):
LGTM!

PR #1463, opened 2022-11-30
test: improve flaky tests

janniks:
### Changes

`src/tests-2.1/pox-2-rosetta-btc-addr-types.ts`:
- trims skippable wait blocks (reducing test time a bit)
- split into separate test files. (could also be added as a separate matrix step in ci.yml parameterized with ENV variables), but this works well for now

`src/tests-2.1/pox-2-rosetta-cycle-phases.ts`:
- ~add `jest.setTimeout` for long-running test~
- ~remove one block unnecessary wait, to speed up test~
- Splits the `.each` matrix into separate files, (ugly but works, and is a lot faster, if we don't need to wait for full cycles for the next parameter). The only way to DRY this up, would be CI matrix / parameterization — would require it's own CI step/job; so, I opted for this approach for now.
- This test could be removed, I'm not sure it adds value (compared to pre-2.1). But I also think it's important, in the sense that it highlights assumptions baked into some parts (e.g. rosetta using the `current_block_height` to construct txs) these assumptions might be fine to not test, but in the weird edge-cases (e.g. 1st/last block of a cycle) something unexpected might happen. So... 
- This helps with timeouts, but not with the off-by-one expects (for testing stacking on last block of cycle)
- The `stackStxWithRosetta` can take >250ms (this might cause problems, where the next block is mined before the Rosetta tx is broadcast)
  
`src/tests-2.1-transition/pox-transition.ts`:
- makes sure rosetta test doesn't stack on last cycle of block, in case the tx doesn't make it into the next block

`general`:
- ~Change `standByForPoxCycle` helper to not use the fickle `...cycle.id`. I only updated this method, since the others don't currently use the `.id` attribute, but there may be some overlap — I'm not sure what the different methods try to achieve.~ EDIT: Kept as is, because many tests expect the 2nd block of a cycle (ie after unlock) rather than the first block of a cycle.
- I'm trying to increase to mine_interval, which I'm not sure helps yet... (total CI time is not really increasing)

### Observations:

`general`:
- Some tests have seen failures from event-processing not being fast enough in testing. I.e. the `expect` checking before the data is processed. I haven't seen this in a while, so I'm ignore it for now. Potential solutions could be: hacky short `sleep(25)` to give event-processing time (seems to work well in some cases); or use socket listens to `expect` after we are certain the event is processed. Although, we should check that the timing fixes would not overlap into the next mined block, which might render some tests ineffective or fail them altogether.
- Sometimes tests fail on codecov-upload — _not on our end_

`test-rosetta-cli-data`:
- Sometimes fails on setup, when failing to download the rosetta cli from GitHub — _not on our end_


src/tests-2.1-transition/pox-transition.ts
```diff
@@ -513,6 +513,34 @@ describe('PoX transition tests', () => {
...
+        expect(BigInt(apiBalance.balance)).toBeLessThan(testAccountBalance);
+        expect(BigInt(apiBalance.locked)).toBe(ustxAmount);
+
+        // test Rosetta address endpoint balance
+        const rosettaBalance = await getRosettaAccountBalance(account.stxAddr);
+        expect(BigInt(rosettaBalance.account.balances[0].value)).toBeLessThan(testAccountBalance);
+        expect(BigInt(rosettaBalance.locked.balances[0].value)).toBe(ustxAmount);
+      });
+
+      test('REPEAT - Verify expected amount of STX are locked', async () => {
```
janniks, line 535: If we want, we can keep this "repeat" in the test, to debug any future flaky failures. The flake hasn't reappeared in a while though.

zone117x (approved):
Thanks again for all the help with these @janniks!

zone117x:
I think we still want to improve these flaky 2.1 related tests, but we're more likely to end up removing some of them in the future, since a lot of these were first implemented to test and develop against under-specified behavior in pox-2.

As it stands, this PR would need a lot of merge conflict resolution, so going to close for now.

PR #1476, opened 2022-12-06
[Stacks 2.1] Support new "block 0" boot events

zone117x:
Implements support for ingesting the `block 0` boot event:

* PR https://github.com/stacks-network/stacks-blockchain/pull/3409 solves the bug with [missing genesis block events](https://github.com/stacks-network/stacks-blockchain/issues/3367)
  * It introduces a `block 0` event payload, which the stacks-node now emits immediately on startup, before syncing any burnchain blocks or mining a Stacks block. The boot contracts and genesis accounts are now contained in this "block 0" payload instead of the first mined Stacks block.

* In order to avoid breaking changes for API consumers and the API itself, the event-ingestion changes preserve the original genesis block behavior, and the `block 0` concept is treated as an internal implementation detail and is not publicly exposed:
  * The `block 0` data is first stored as non-canonical in the db during event ingestion
  * Then, once the first mined Stacks block is received, the `block 0` data is adjusted so that it is associated with the first mined Stacks block.

* For a reference to the existing behavior, see first block data from mainnet, e.g. `https://stacks-node-api.mainnet.stacks.co/extended/v1/block/by_height/1`

* A new `block zero handling` test suite is implemented to validate the event ingestion changes.

* This PR uses the bitcoin regtest + stacks-node image from https://github.com/zone117x/stacks-regtest-env/pull/18 which has been updated to handle the new boot event behavior.

Re-enable double-stacking tests:

* Stacking to both pox-1 and pox-2 during Period 2a no longer triggers a stacks-node panic as described at https://github.com/stacks-network/stacks-blockchain/issues/3423

Standalone regtest image:

* Initial implementation of a new standalone regtest dockerfile which is epoch 2.1 capable. This is intended to be used in CI/local testing as an alternative to the existing standalone docker image (which uses a mocknet stacks-node that is non-functional for epoch 2.1 testing). 
  * The new dockerfile is not yet hooked up to CI -- this will likely be done in a follow-up PR.


janniks (approved):
Looks good!

src/datastore/pg-write-store.ts
```diff
@@ -715,6 +725,45 @@ export class PgWriteStore extends PgStore {
...
   }
 
+  async fixBlockZeroData(sql: PgSqlClient, blockOne: DbBlock): Promise<void> {
+    const tablesUpdates: Record<string, number> = {};
+    const txsResult = await sql<TxQueryResult[]>`
+      UPDATE txs
+      SET 
+        canonical = true,
+        block_height = 1,
+        tx_index = tx_index + 1,
```
rafaelcr, line 735: Shouldn't this be `last_block_1_tx_index + 1`? IIRC block 1 has like 2 or 3 transactions within it

zone117x: In theory it seems like there could have been an alternative history where block 1 also included some arbitrary additional txs, but that didn't happen in current mainnet so we don't have to worry about it.

rafaelcr (approved):
LGTM 👍🏻

PR #1484, opened 2022-12-12
Fix bns test

BLuEScioN:
Fix bns test that is causing a timeout on build. 

https://github.com/hirosystems/stacks-blockchain-api/actions/runs/3678727169

Test is passing CI, but failing build. Test still confirms that the post request returns before bns import has finished.

### DIFF
--- a/src/tests-bns/event-server-tests.ts
+++ b/src/tests-bns/event-server-tests.ts
@@ -1067,12 +1067,5 @@ describe('BNS event server tests', () => {
     const configState = await db.getConfigState();
     expect(configState.bns_names_onchain_imported).toBe(false)
     expect(configState.bns_subdomains_imported).toBe(false)
-
-    await new Promise(resolve => setTimeout(async() => {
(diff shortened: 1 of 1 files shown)


src/tests-bns/event-server-tests.ts
```diff
@@ -1067,12 +1067,5 @@ describe('BNS event server tests', () => {
     const configState = await db.getConfigState();
     expect(configState.bns_names_onchain_imported).toBe(false)
     expect(configState.bns_subdomains_imported).toBe(false)
-
-    await new Promise(resolve => setTimeout(async() => {
```
rafaelcr, line 1071: I think this test is valuable, but what you're missing is that you're not invalidating the timeout once your test is complete.

You should keep a reference to the timeout so you can stop and invalidate it once the promise resolves.

zone117x (approved):

zone117x:
Reminder to use [conventional commits](https://www.conventionalcommits.org) for the squash and merge commit message

PR #1491, opened 2022-12-15
fix: add bnsImportUpdate to event emitter to fix BNS import test

BLuEScioN:
Fixing bns test 

I verified again that the bns middleware is working asynchronously via breakpoints. When making two consecutive requests, I saw the runtime switch between running the asyn tasks in the bns middleware and working on a second request.

The issue I saw was that the processing of the 1st genesis block, specifically the update db step, is taking a long time, a couple minutes sometimes. I observed the actual bns import taking only a few seconds (considering the small test bns data) Because bns tests have a timeout of 60 seconds, the test times out. 

Increasing the general timeout time for bns tests could leave to much longer build times.

In this PR, I fixed the test by making bns import an event that could be listened to, thereby eliminating the need to use an arbitrary timeout using settimeout

### DIFF
--- a/src/api/routes/ws/web-socket-transmitter.ts
+++ b/src/api/routes/ws/web-socket-transmitter.ts
@@ -80,10 +80,12 @@ export class WebSocketTransmitter {
         .add(() => this.addressUpdate(address, blockHeight))
         .catch(error => logger.error(`WebSocketTransmitter addressUpdate error: ${error}`))
     );
-    this.db.eventEmitter.addListener('bnsImportUpdate', () =>
-      this.queue
-        .add(() => this.bnsImportUpdate())
-        .catch(error => logger.error(`WebSocketTransmitter bnsImportUpdate error: ${error}`))
+    this.db.eventEmitter.addListener(
+      'bnsImportUpdate',
+      (bnsNamesOnchainImported, bnsSubdomainsImported) =>
+        this.queue
+          .add(() => this.bnsImportUpdate(bnsNamesOnchainImported, bnsSubdomainsImported))
+          .catch(error => logger.error(`WebSocketTransmitter bnsImportUpdate error: ${error}`))
     );
 
     this.channels.push(new SocketIOChannel(this.server));
@@ -135,7 +137,7 @@ export class WebSocketTransmitter {
     }
   }
 
-  private async bnsImportUpdate() {
+  private async bnsImportUpdate(bnsNamesOnchainImported: boolean, bnsSubdomainsImported: boolean) {
     if (this.channels.find(c => c.hasListeners('bnsImport'))) {
       try {
         const {
--- a/src/api/routes/ws/web-socket-transmitter.ts
+++ b/src/api/routes/ws/web-socket-transmitter.ts
@@ -80,6 +80,11 @@ export class WebSocketTransmitter {
         .add(() => this.addressUpdate(address, blockHeight))
         .catch(error => logger.error(`WebSocketTransmitter addressUpdate error: ${error}`))
     );
+    this.db.eventEmitter.addListener('bnsImportUpdate', () =>
+      this.queue
+        .add(() => this.bnsImportUpdate())
+        .catch(error => logger.error(`WebSocketTransmitter bnsImportUpdate error: ${error}`))
+    );
 
     this.channels.push(new SocketIOChannel(this.server));
     this.channels.push(new WsRpcChannel(this.server));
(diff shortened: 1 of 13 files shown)


src/api/routes/ws/web-socket-transmitter.ts
```diff
@@ -130,6 +137,12 @@ export class WebSocketTransmitter {
     }
   }
 
+  private async bnsImportUpdate(bnsNamesOnchainImported: boolean, bnsSubdomainsImported: boolean) {
+    if (this.channels.find(c => c.hasListeners('bnsImport'))) {
+      await this.send('bnsImport', bnsNamesOnchainImported, bnsSubdomainsImported);
+    }
+  }
+
```
rafaelcr, line 145: `bnsImport` is an event that is only relevant internally to the API, so there's no need to add it as a WebSocket event for external users to consume. You should revert this file (and web-socket-channel.ts)

src/datastore/pg-store.ts
```diff
@@ -3851,4 +3860,9 @@ export class PgStore {
       return { results: parsed, total: totalQuery[0].count };
     });
   }
+
+  async getConfigState(): Promise<DbConfigState> {
+    const queryResult = await this.sql<DbConfigState[]>`SELECT * FROM config_state`;
+    return queryResult[0];
+  }
```
rafaelcr, line 3867: This doesn't seem to be needed in `PgStore`, can you move it back to `PgWriteStore`?

src/tests-bns/event-server-tests.ts
```diff
@@ -1069,15 +1068,14 @@ describe('BNS event server tests', () => {
...
 
-    const timeoutId: NodeJS.Timeout = await new Promise(resolve => {
-      const timeoutId = setTimeout(async() => {
-        const configState = await db.getConfigState();
-        expect(configState.bns_names_onchain_imported).toBe(true)
-        expect(configState.bns_subdomains_imported).toBe(true)
-        resolve(timeoutId)
-      }, 2000)
+    await new Promise(resolve => {
+      db.eventEmitter.on('bnsImportUpdate', (bnsNamesOnchainImported, bnsSubdomainsImported) => {
```
rafaelcr, line 1072: You should also remove this listener once this promise is awaited and done

rafaelcr (changes requested):
Can you change this PR's name to something like
"fix: add bnsImportUpdate to event emitter to fix BNS import test"

src/datastore/pg-write-store.ts
```diff
@@ -1683,6 +1683,10 @@ export class PgWriteStore extends PgStore {
         bns_subdomains_imported = ${configState.bns_subdomains_imported},
         token_offering_imported = ${configState.token_offering_imported}
     `;
+    await this.notifier?.sendBnsImport({
+      bnsNamesOnchainImported: configState.bns_names_onchain_imported,
+      bnsSubdomainsImported: configState.bns_subdomains_imported,
+    });
```
rafaelcr, line 1689: Now that I think about it, since you're calling this on `updateConfigState` I think you should rename your update to `configStateUpdate` and add the `tokenOfferingImported` property as well.

That way we'll have a more generalized update when any config has changed.

src/tests-bns/event-server-tests.ts
```diff
@@ -1069,14 +1069,14 @@ describe('BNS event server tests', () => {
...
-      db.eventEmitter.on('bnsImportUpdate', (bnsNamesOnchainImported, bnsSubdomainsImported) => {
-        if (bnsNamesOnchainImported && bnsSubdomainsImported) {
-          expect(bnsNamesOnchainImported).toBe(true)
-          expect(bnsSubdomainsImported).toBe(true);
+      db.eventEmitter.on('configStateUpdate', (configState) => {
+        if (configState.bns_names_onchain_imported && configState.bns_subdomains_imported) {
+          expect(configState.bns_names_onchain_imported).toBe(true)
+          expect(configState.bns_subdomains_imported).toBe(true);
           resolve(undefined);
         }
```
rafaelcr, line 1077: You'll just need an `else` with a reject so the test actually fails if this is the case

rafaelcr (approved):
LGTM, feel free to merge after addressing the latest feedback

src/tests-bns/event-server-tests.ts
BLuEScioN: This event can be emitted when one or both of these conditions are false. In fact, this is expected because the names will be imported first, and then the subdomains. So, the first time that this check runs, subdomains will be false. We don't want to reject because we are exepcting subdomains to be imported next. This test will fail with a timeout if we never hit resolve.

PR #1498, opened 2022-12-22
API v7 RC (2.1-capable release)

zone117x:
PR to pull all the 2.1 related capabilities into the develop branch, from which we can then cut a v7 RC. 

So far this branch has been able to fully boot-up and run against current 2.0 mainnet via event-replay, and has been running against the 2.1-testnet for a while.

Once this is in the develop branch, we may want to then merge to the `beta` branch -- @CharlieC3 thoughts?


.github/workflows/regtest-image.yml
```diff
@@ -0,0 +1,141 @@
...
+        description: 'stacks-blockchain git commit'
+        required: true
+        type: string
+
+env:
+  STACKS_API_COMMIT: ${{ github.sha }}
+  STACKS_BLOCKCHAIN_COMMIT: ${{ inputs.stacks_blockchain_commit }}
+
+jobs:
+  build-stacks-node:
```
CharlieC3, line 16: Curious why the stacks-node is being built in this workflow? Can't we just download a pre-built binary from one of the github releases/tags?

zone117x: See https://github.com/hirosystems/stacks-blockchain-api/pull/1481

The docker image to create a working regtest setup is much more complicated than one in, say, the blockchain repo.

zone117x: Ohh if you're asking specifically about just the stacks-node binary -- yes in the future it should just grab an existing one. That requires the following to be done:
1. Stable release/tag for the version we need (as opposed to just the `next` branch or on the PRs off of that one).
2. The other images published need to support arm64 (and ideally also glibc linux distros as opposed to just alpine).

CharlieC3 (approved):
LGTM! :shipit:

PR #1514, opened 2023-01-09
fix: prevent token metadata processor from blocking api launch

### LINKED ISSUE #1482: [testnet] port 3999 fails to open when STACKS_API_ENABLE_FT_METADATA is set to 1
As the title states, the testnet api fails to open port 3999 when the `STACKS_API_ENABLE_FT_METADATA` env var is set to `1`.  This is occurring in `testnet` running api v6.2.2.  

some additional information for thought:
- the `token_metadata_queue` is empty inside the database, so the issue is not with the replay
- v5.x.x has an empty `token_metadata_queue` table and the host proxy serving 3999 starts up fine, this is not the case with v6.2.2
- workaround is to remove the `STACKS_API_ENABLE_FT_METADATA` env var from the deployment, this will also cause metadata to be unavailable for users

rafaelcr:
The current call to `drainDbQueue` on the token processor during boot sequence waits for the `queue.onEmpty` promise to resolve before continuing to launch the API server.

This promise is supposed to wait until every token is processed, so if the token queue has a lot of work (and especially if it has `STACKS_API_TOKEN_METADATA_STRICT_MODE` enabled) this could lead to a very long (or infinite) wait time for the token metadata work to conclude, preventing the API server from ever launching.

This PR only changes the call so the first few tokens are enqueued and the API launch can proceed as expected. The rest of the token queue will be processed automatically afterwards in a non-blocking way.

Fixes #1482

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -163,8 +163,8 @@ async function init(): Promise<void> {
         handler: () => tokenMetadataProcessor.close(),
         forceKillable: true,
       });
-      // check if db has any non-processed token queues and await them all here
-      await tokenMetadataProcessor.drainDbQueue();
+      // Enqueue a batch of pending token metadata processors, if any.
+      await tokenMetadataProcessor.checkDbQueue();
     }
   }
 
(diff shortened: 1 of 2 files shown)


zone117x (approved):
LGTM!

PR #1523, opened 2023-01-16
fix: add contract_identifier index on contract_logs table

rafaelcr:
For SIP-013 SFT support, the Token Metadata Service needs to scan the `contract_logs` table to look for `sft_mint` events emitted by a SIP-013 contract so it can determine the number of tokens declared in the contract. https://github.com/hirosystems/token-metadata-service/issues/56

This index optimizes queries that use `contract_identifier` as the main filter column for this table.


zone117x (approved):
LGTM

Issue #1522, opened 2023-01-13
disable BTC faucet endpoint if BTC_RPC_PORT & BTC_RPC_HOST are not defined


deantchi:
While it's not clear if providing an invalid `BTC_RPC_PORT` & `BTC_RPC_HOST` env var combination causes issues with the API as the API appeared to continued to function, is it possible to check for validity in case it does break something?

PR #1530, opened 2023-01-25
fix: disabled BTC faucet endpoint

sulemanqaiser97:
## Description

1 . Added middleware to BTC faucet endpoint to catch a specific exception if BTC_RPC_PORT & BTC_RPC_HOST are not defined and respond with 403 status code with the following message.  
{
    "error": "BTC Faucet not fully configured.",
    "success": false
}

2. Added a unit test to verify the above scenario. 
3. Added a new response type 403 in /extended/v1/faucets/btc endpoint in openapi.yaml file.  

For details refer to issue #1522

## Type of Change
- [ ] Bug fix

## Checklist
- [ ] `npm run test` passes

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -85,6 +85,7 @@ export async function startApiServer(opts: {
   const app = express();
   const apiHost = serverHost ?? process.env['STACKS_BLOCKCHAIN_API_HOST'];
   const apiPort = serverPort ?? parseInt(process.env['STACKS_BLOCKCHAIN_API_PORT'] ?? '');
+  const { BTC_RPC_PORT, BTC_RPC_HOST } = process.env;
 
   if (!apiHost) {
     throw new Error(
@@ -225,7 +226,7 @@ export async function startApiServer(opts: {
       router.use('/fee_rate', createFeeRateRouter(datastore));
       router.use('/tokens', createTokenRouter(datastore));
       router.use('/pox2_events', createPox2EventsRouter(datastore));
-      if (chainId !== ChainID.Mainnet && writeDatastore) {
+      if (chainId !== ChainID.Mainnet && writeDatastore && BTC_RPC_PORT && BTC_RPC_HOST) {
         router.use('/faucets', createFaucetRouter(writeDatastore));
       }
       return router;
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -85,7 +85,6 @@ export async function startApiServer(opts: {
   const app = express();
   const apiHost = serverHost ?? process.env['STACKS_BLOCKCHAIN_API_HOST'];
   const apiPort = serverPort ?? parseInt(process.env['STACKS_BLOCKCHAIN_API_PORT'] ?? '');
-  const { BTC_RPC_PORT, BTC_RPC_HOST } = process.env;
 
   if (!apiHost) {
     throw new Error(
@@ -226,7 +225,7 @@ export async function startApiServer(opts: {
       router.use('/fee_rate', createFeeRateRouter(datastore));
       router.use('/tokens', createTokenRouter(datastore));
       router.use('/pox2_events', createPox2EventsRouter(datastore));
-      if (chainId !== ChainID.Mainnet && writeDatastore && BTC_RPC_PORT && BTC_RPC_HOST) {
+      if (chainId !== ChainID.Mainnet && writeDatastore) {
         router.use('/faucets', createFaucetRouter(writeDatastore));
       }
       return router;
--- a/src/api/routes/faucets.ts
+++ b/src/api/routes/faucets.ts
@@ -69,59 +69,62 @@ export function createFaucetRouter(db: PgWriteStore): express.Router {
   const router = express.Router();
   router.use(express.urlencoded({ extended: true }));
   router.use(express.json());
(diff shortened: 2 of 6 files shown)


asimm241 (changes requested):

src/api/init.ts
```diff
@@ -225,7 +226,7 @@ export async function startApiServer(opts: {
       router.use('/fee_rate', createFeeRateRouter(datastore));
       router.use('/tokens', createTokenRouter(datastore));
       router.use('/pox2_events', createPox2EventsRouter(datastore));
-      if (chainId !== ChainID.Mainnet && writeDatastore) {
+      if (chainId !== ChainID.Mainnet && writeDatastore && BTC_RPC_PORT && BTC_RPC_HOST) {
```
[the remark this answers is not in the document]
sulemanqaiser97: Yeah. You are right this will disable all endpoints for /faucets. I have fixed it and the current change will only disable /faucets/btc and /faucet/btc/:address endpoints. Kindly look into it?

zone117x (commented):
My preference would actually be to still register the route, but return something like "not enabled". That would be more helpful to anyone trying to work with this route, in my opinion. @rafaelcr thoughts?

sulemanqaiser97:
@zone117x  if BTC_RPC_PORT or BTC_RPC_HOST are not defined the faucet endpoint will respond with following message.

{
    "message": "POST /extended/v1/faucets/btc not found"
}

but as per your suggestion, we can have a customized message indicating that BTC faucet is not properly configured.

zone117x (changes requested):
Yes, please make the endpoint return "BTC faucet is not configured". It should probably return status code 403, as that looks the most applicable: "indicates that the server understood the request but refuses to authorize it". This response should also be added to the OpenAPI docs for the endpoint.

Also please add a unit test to verify that the endpoint works as expected when configured and not-configured.

sulemanqaiser97:
@rafaelcr would like to hear your thoughts on this issue. Can you please share your comment on this endpoint configuration and the addition of unit tests? Thank you.

rafaelcr:
I agree that registering the endpoint and returning 403 error code is the best approach.

The unit test can be very simple as @zone117x suggested just to validate those ENV vars are accounted for @sulemanqaiser97 ... you should add your tests to this file: https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/tests/faucet-btc-tests.ts

sulemanqaiser97:
@rafaelcr need some clarification on this. 
Current Implementation: 
If these environment variables are not defined, then /faucet/btc endpoints return 500 internals server error with the message that 'the BTC faucet is not fully configured'. These checks are validated in getRpcClient() function.
image

As per our discussion, we are now not disabling this endpoint rather returning 403 ( Forbidden ) with same customised message ('BTC faucet is not configured'). So, aren't we implementing the same check which already exists and is validated when rpcClient is initialised with different status code 403.

rafaelcr:
That sounds like a good idea @sulemanqaiser97 , you could reuse that same check but just catch the exception in the endpoints to return a 403 instead of 500.

src/api/routes/faucets.ts
```diff
@@ -69,6 +69,15 @@ export function createFaucetRouter(db: PgWriteStore): express.Router {
   const router = express.Router();
   router.use(express.urlencoded({ extended: true }));
   router.use(express.json());
+  const missingBtcConfigMiddleware = (
+    err: Error,
+    req: express.Request,
+    res: express.Response,
+    next: express.NextFunction
+  ) =>
+    err.message === 'BTC Faucet not fully configured.'
```
zone117x, line 78: This error message string comparison is untyped and seems brittle. If using this pattern, then an specific Error class for this should be created. The only time we do error message string comparisons is when dealing with errors from other dependencies that we don't control.

sulemanqaiser97: Enum or constant can also be used to avoid untyped string comparison. 
 const FaucetError = {
  BTC_FAUCET_NOT_CONFIGURED: 'BTC Faucet not fully configured.'
}

Comparison:
 err.message === FaucetError.BTC_FAUCET_NOT_CONFIGURED 
 
 @zone117x what are your thoughts on this?

zone117x: That approach seems a bit strange. I'd just go with the existing pattern of creating an Error type and checking for it. Example:
https://github.com/hirosystems/stacks-blockchain-api/blob/6d66c3ccc1b270b70edf4919efde40b582c2cbd2/src/api/init.ts#L300

src/errors.ts
```diff
@@ -24,3 +24,10 @@ export class InvalidRequestError extends Error {
     this.status = status;
   }
 }
+
+export class BtcFaucetConfigError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'BtcFaucetConfigError';
```
zone117x, line 31:
```suggestion
    this.name = this.constructor.name;
```

csgui:
Hey @sulemanqaiser97 ! I am from Hiro API team.

Jumping in to close this issue with the work you've done. 

Thanks for the collaboration.

zone117x (approved):
Thanks everyone!

rafaelcr (approved):

PR #1531, opened 2023-01-26
fix: update total STX supply to the year 2050 projected amount

zone117x:
Fixes https://github.com/hirosystems/explorer/issues/973

The liquid stx amount that is calculated from the postgres db is accurate (or at least it matches the `total_liquid_supply_ustx` value reported by the stacks-node):

* `/extended/v1/stx_supply - unlocked_stx: 1357774888442244` matches:
* `/v2/pox - total_liquid_supply_ustx: 1357774888442244`

However, the `total_stx` value is incorrect, and is currently _greater_ than the liquid stx supply.

The `total_stx` value is a constant that was copied from the Stacks 1.0 codebase, see https://github.com/hirosystems/stacks-blockchain-api/pull/466
> This was copied from the legacy 1.0 API codebase.

https://github.com/hirosystems/stacks-blockchain-api/blob/b6018dc06ae2dabccc85e93090fccc267e054b8f/src/helpers.ts#L220-L222

Here's what the [Stacks 2.0 whitepaper](https://cloudflare-ipfs.com/ipfs/QmaGgiVHymeDjAc3aF1AwyhiFFwN97pme5m536cHT4FsAW) states about about the STX supply:

> Block reward: 1000 STX/block for first 4 yrs; 500 STX/block for following 4 yrs; 250 for the 4 yrs after that; and then 125 STX/block in perpetuity after that.
...
> The Stacks cryptocurrency has a predefined future supply that reaches approx 1,818M STX by year 2050 ...

This PR changes the `total_stx` value to the year 2050 projected STX amount, because "125 STX/block in perpetuity" means the total supply is not a constant.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -217,9 +217,18 @@ export function formatMapToObject<TKey extends string, TValue, TFormatted>(
   return obj;
 }
 
-export const TOTAL_STACKS = new BigNumber(1320000000)
-  .plus(322146 * 100 + 5 * 50000) // air drop
-  .toString();
+// Note: this is the legacy amount defined in the Stacks 1.0 codebase:
+// export const TOTAL_STACKS /* 1352464600000000 */ = new BigNumber(1320000000)
+//   .plus(322146 * 100 + 5 * 50000) // air drop
+//   .toString();
+
+// See the Stacks 2.0 whitepaper: https://cloudflare-ipfs.com/ipfs/QmaGgiVHymeDjAc3aF1AwyhiFFwN97pme5m536cHT4FsAW
+//   > The Stacks cryptocurrency has a predefined future supply that reaches approx 1,818M STX by year 2050
+//   > Block reward: 1000 STX/block for first 4 yrs;
+//   > 500 STX/block for following 4 yrs;
+//   > 250 for the 4 yrs after that; and then 125 STX/block in perpetuity after that.
+// We are going to use the year 2050 projected supply because "125 STX/block in perpetuity" means the total supply is infinite.
+export const TOTAL_STACKS = new BigNumber(1_818_000_000n.toString());
 
 const MICROSTACKS_IN_STACKS = 1_000_000n;
 export const STACKS_DECIMAL_PLACES = 6;
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1841,7 +1841,9 @@ paths:
         - Info
       operationId: get_stx_supply
       summary: Get total and unlocked STX supply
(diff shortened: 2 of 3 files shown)


src/helpers.ts
```diff
@@ -217,9 +217,18 @@ export function formatMapToObject<TKey extends string, TValue, TFormatted>(
...
+// Note: this is the legacy amount defined in the Stacks 1.0 codebase:
+// export const TOTAL_STACKS /* 1352464600000000 */ = new BigNumber(1320000000)
+//   .plus(322146 * 100 + 5 * 50000) // air drop
+//   .toString();
+
+// See the Stacks 2.0 whitepaper: https://cloudflare-ipfs.com/ipfs/QmaGgiVHymeDjAc3aF1AwyhiFFwN97pme5m536cHT4FsAW
+//   > The Stacks cryptocurrency has a predefined future supply that reaches approx 1,818M STX by year 2050
+//   > Block reward: 1000 STX/block for first 4 yrs;
+//   > 500 STX/block for following 4 yrs;
+//   > 250 for the 4 yrs after that; and then 125 STX/block in perpetuity after that.
```
zone117x, line 229: @kantai does the Stacks implementation match this whitepaper description?

src/helpers.ts
```diff
@@ -217,9 +217,18 @@ export function formatMapToObject<TKey extends string, TValue, TFormatted>(
...
+// We are going to use the year 2050 projected supply because "125 STX/block in perpetuity" means the total supply is infinite.
+export const TOTAL_STACKS = new BigNumber(1_818_000_000n.toString());
```
zone117x, line 231: @diwakergupta thoughts on using this value?

diwakergupta: Yeah anchoring on 2050 supply makes sense. I would also update the docs for this endpoint to include this qualified in big bold letters.

zone117x: Docs updated 👍

src/helpers.ts
kantai: Yes, though it uses bitcoin block height as opposed to calendar years:

See: 

```rust
        let effective_ht = burn_block_height.saturating_sub(first_burn_block_height);
        let blocks_per_year = 52596;
        let stx_reward = if effective_ht < blocks_per_year * 4 {
            1000
        } else if effective_ht < blocks_per_year * 8 {
            500
        } else if effective_ht < blocks_per_year * 12 {
            250
        } else {
            125
        };
```

from

https://github.com/stacks-network/stacks-blockchain/blob/master/src/chainstate/stacks/db/blocks.rs#L4136-L4171

`first_burn_block_height` is `666050`

kantai (approved):

Issue #415, opened 2021-01-25
Generate more accurate btc StackStxOp tx


zone117x:
See https://github.com/blockstack/stacks-blockchain-api/pull/414#discussion_r564097195_

An address used for the stack-stx function arg list recreation needs first output address of the bitcoin transaction. However, this requires more info from the stacks-node event.

agraebe:
is this still relevant (with the mentioned PR merged in)?

zone117x:
Yeah, the txs generated by the API for this are still incorrect. Per Aaron's comment, we need more info from the stacks-node for these events.

agraebe:
do we have an equivalent issue in the stacks-blockchain repo to track that request? I'd say we put this API issue in the icebox while we wait for the upstream info to be available

zone117x:
Issue created https://github.com/blockstack/stacks-blockchain/issues/2591

zone117x:
Closed by https://github.com/hirosystems/stacks-blockchain-api/pull/1533

PR #1533, opened 2023-01-26
fix: use correct `pox-addr` arg while parsing `stack-stx` bitcoin-op

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/415

Use the new pox-2 contract event which include enough data to now generate an accurate `stack-stx` transaction from a bitcoin-op.


rafaelcr (approved):
Sorry I missed this, LGTM!

PR #1534, opened 2023-01-27
fix: add block_height index to contract_logs

rafaelcr:
On boot, the Token Metadata Service needs to scan the `contract_logs` table to look for SIP-019 notifications that might have happened between two block heights. This index makes the performance of this query trivial on the API database.


zone117x (approved):
LGTM!

PR #1552, opened 2023-02-13
OpenAPI: Get mempool transations - Removed default value and updated max and example values

LakshmiLavanyaKasturi:
Updated maximum and example values for the Get mempool transitions.
Removed default value.

The openAPI.yaml file is updated.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -268,9 +268,8 @@ paths:
           required: false
           schema:
             type: integer
-            default: 96
-            maximum: 200
-            example: 100
+            maximum: 50
+            example: 20
         - name: offset
           in: query
           description: index of first mempool transaction to fetch
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -268,6 +268,7 @@ paths:
           required: false
           schema:
             type: integer
+            default: 20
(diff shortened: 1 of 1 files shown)


docs/openapi.yaml
```diff
@@ -268,9 +268,8 @@ paths:
           required: false
           schema:
             type: integer
-            default: 96
-            maximum: 200
-            example: 100
+            maximum: 50
```
obycode, line 271: Looks like the default is 20. Can we add that field back in?

LakshmiLavanyaKasturi: Trying to learn,  "limit": 20, is the limit considered as the default value?

LakshmiLavanyaKasturi: Done. Updated.

LakshmiLavanyaKasturi:
Relates to https://github.com/hirosystems/docs/issues/299

rafaelcr (approved):
Thanks @LakshmiLavanyaKasturi

docs/openapi.yaml
obycode: `limit` is the name of the parameter. The default value is what this limit is set to if no parameter is passed.

PR #1559, opened 2023-02-21
fix: support genesis block 0 during BNS sync

rafaelcr:
Keep backwards compatibility for Stacks node 2.05 (starting at block height 1).


zone117x (approved):
Thanks for the quick fix @rafaelcr!

Issue #1557, opened 2023-02-16
Transactions are showing twice, same nonce, same fee, same tx hash... one confirmed, one pending


314159265359879:
Not yet sure how this is happening but I have seen at least 5 cases like this with seemingly duplicate transactions. The ones with nonce 1, 2, 3 are calling the same function/making the same transfer, with the same fee and nonce twice. That shouldn't be possible and I am not sure how this is happening. 
Since the transaction hashes are the same... perhaps they are not even different instances. an API issue? if they were RBF'ed transactions I would expect them to have different fees yet same nonce. 

![image](https://user-images.githubusercontent.com/33360391/219475527-1187c66a-f7ac-45af-a927-00e4c6b0450f.png)

Is this an API issue?

rafaelcr:
I believe this is a symptom of https://github.com/hirosystems/stacks-blockchain-api/issues/1428

314159265359879:
This is causing the weirdest things, it looks like it is affecting more protocols such as ALEX:

Here is a swap that actually goes through succesfully:
![image](https://user-images.githubusercontent.com/33360391/220665759-53f5e33e-d5fa-4e8c-ac2c-8402f32d50f6.png)

Then on the address it also shows as pending (note same nonce, same tx hash)
![image](https://user-images.githubusercontent.com/33360391/220665920-1d082695-3a45-46a5-830d-0a49203e40b3.png)

In the wallet we do not see it:
it just shows all the anchored transactions
![image](https://user-images.githubusercontent.com/33360391/220666894-dc5d35c4-6edb-4ae8-bfcd-70457e2bd91b.png)

Then on ALEX a users sees "pending"
![image (5)](https://user-images.githubusercontent.com/33360391/220667138-4001c453-64fb-4ae3-93a8-3274bc9f56c5.png)

zone117x:
@314159265359879 can you verify if this issue is fixed for you now?

314159265359879:
> @314159265359879 can you verify if this issue is fixed for you now?

Looks fixed to me. Thanks! 
I will reopen if it does pop up again.

Issue #1428, opened 2022-11-09
Improve mempool data


obycode:
Recent work on the stacks node (e.g. https://github.com/stacks-network/stacks-blockchain/pull/3337, https://github.com/stacks-network/stacks-blockchain/pull/3349) exposes better sources for the mempool data. This is a generic issue to discuss how the API node is currently computing mempool information and how we can improve this data using the new techniques. Various ecosystem entities are relying on these endpoints.

Related endpoints:
* https://docs.hiro.so/api#tag/Rosetta/operation/rosetta_mempool
* https://docs.hiro.so/api#tag/Transactions/operation/get_mempool_transaction_list

zone117x:
It looks like there's a few approaches to improve mempool data for public consumption:
* Expose the mempool data that folks are interested in through the stacks-node RPC endpoints.
* Expose those stacks-node mempool prometheus metrics to the public in some fashion.
* Add the improved stacks-node mempool data to the event-stream, in which case the API could store that in postgres and return in an API endpoint.

The API endpoints you mentioned are currently returning a view of the mempool based on the `new_mempool_tx`, `drop_mempool_tx`, and `new_microblocks` events. Note that these events are "lossy" / unreliable due to race-conditions in the stacks-node. This causes events to sometimes be emitted out-of-order with respect to each other, or out-of-order in respect to the `new_block` event (which is generally the only event that is a reliable source-of-truth). In some of these situations, especially when there are also anchor-block and/or microblock re-orgs happening, the API does not have accurate enough data from the stacks-node to ensure its view of the mempool is 100% correct. 

There are known issues where the API can sometimes return a mined tx with correct data, but also have the same tx incorrectly reported as still in the mempool. There are several areas in the API's re-org handling code to try to prevent this situation, however, due to the complexity of handling anchor-block and microblock re-orgs _at the same time_ as trying to handle buggy / out-of-order data from the stacks-node, we end up with a garbage in, garbage out scenario. 

In the past when trying to resolve these discrepancies in the API, we end up adding more postgres disk/computational overhead to the already expensive process of ingesting event data, and typically also introduce regressions. For example, we fix one scenario of a tx being inaccurately reported as still in the mempool, which inadvertently causes a bug where a tx can end up being reported as completely missing from both the mempool view & the mined tx view (e.g. after some inscrutable series of out-of-order or missing events during a series of re-org events).

[...]

PR #1561, opened 2023-02-22
fix: ensure transactions are never reported as both pending and confirmed

zone117x:
Fixes https://github.com/hirosystems/stacks-blockchain-api/issues/1557, related to https://github.com/hirosystems/stacks-blockchain-api/issues/1428

This adds a function which detects any transactions that are erroneously marked as both pending in the mempool table and also confirmed in the mined txs table.

The function is called after all other sql writes during block & microblock ingestion, but before any reads or view refreshes that depend on the mempool table.

This should be a safe work-around for whatever bug is causing the underlying problem.

### DIFF
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -61,6 +61,7 @@ import {
   Pox2EventInsertValues,
   DbTxRaw,
   DbMempoolTxRaw,
+  DbChainTip,
 } from './common';
 import { ClarityAbi } from '@stacks/transactions';
 import {
@@ -152,15 +153,7 @@ export class PgWriteStore extends PgStore {
     return super.sqlTransaction(callback, false);
   }
 
-  async getChainTip(
-    sql: PgSqlClient,
-    useMaterializedView = true
-  ): Promise<{
-    blockHeight: number;
-    blockHash: string;
-    indexBlockHash: string;
-    burnBlockHeight: number;
-  }> {
+  async getChainTip(sql: PgSqlClient, useMaterializedView = true): Promise<DbChainTip> {
     if (!this.isEventReplay && useMaterializedView) {
       return super.getChainTip(sql);
     }
@@ -452,9 +445,7 @@ export class PgWriteStore extends PgStore {
 
       if (!this.isEventReplay) {
         await this.reconcileMempoolStatus(sql);
-      }
 
-      if (!this.isEventReplay) {
         const mempoolStats = await this.getMempoolStatsInternal({ sql });
         this.eventEmitter.emit('mempoolStatsUpdate', mempoolStats);
       }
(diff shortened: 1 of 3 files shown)


src/datastore/pg-write-store.ts
```diff
@@ -727,6 +731,10 @@ export class PgWriteStore extends PgStore {
         );
       }
 
+      if (!this.isEventReplay) {
+        await this.reconcileMempoolStatus(sql);
```
zone117x, line 735: @rafaelcr do these 2 call sites look safe to you?

rafaelcr: Yep, it's inside the write tx so I think we're good.

src/datastore/pg-write-store.ts
```diff
@@ -450,6 +450,10 @@ export class PgWriteStore extends PgStore {
         }
       }
 
+      if (!this.isEventReplay) {
+        await this.reconcileMempoolStatus(sql);
+      }
+
       if (!this.isEventReplay) {
```
rafaelcr, line 457: Could you consolidate these two code blocks into the same conditional? Same in the other usage

src/datastore/pg-write-store.ts
```diff
@@ -759,6 +767,34 @@ export class PgWriteStore extends PgStore {
     }
   }
 
+  // Find any transactions that are erroneously still marked as both `pending` in the mempool table
+  // and also confirmed in the mined txs table. Mark these as pruned in the mempool and log warning.
+  // This must be called _after_ any writes to txs/mempool tables during block and microblock ingestion,
+  // but _before_ any reads or view refreshes that depend on the mempool table.
+  // NOTE: this is essentially a work-around for whatever bug is causing the underlying problem.
+  async reconcileMempoolStatus(sql: PgSqlClient): Promise<void> {
```
rafaelcr, line 775: While this works great, I think we should add another similar check inside `updateMempoolTxs` that inserts them as `pruned` initially if we already have that same incoming mempool tx in a block (in a scenario where we receive a mempool tx with some huge delay for any reason).

That way, we won't display the inconsistent result from the time that delayed mempool tx arrives until the next block/microblock comes and we reconcile here.

What do you think?

src/tests/mempool-tests.ts
```diff
@@ -1405,4 +1405,125 @@ describe('mempool tests', () => {
...
+      execution_cost_read_length: 0,
+      execution_cost_runtime: 0,
+      execution_cost_write_count: 0,
+      execution_cost_write_length: 0,
+    };
+
+    // Simulate the bug with a txs being in the mempool at confirmed at the same time by
+    // directly inserting the mempool-tx and mined-tx, bypassing the normal update functions.
+    await db.updateBlock(db.sql, dbBlock1);
+    await db.updateMempoolTxs({ mempoolTxs: [mempoolTx] });
```
rafaelcr, line 1490: If we apply the fix to `updateMempoolTxs` too, you could still run this test just by inserting the tx directly into the DB here. Or maybe reversing the order of these two lines.

zone117x: Good idea, done!

rafaelcr (approved):
LGTM!

PR #1562, opened 2023-02-23
added note for mempool endpoints to clarify consistency gurantees

trichards-hiro:

## Description

A clarification was needed to explicitly state that data from mempool endpoints are not guaranteed.

1. Several Discord users noted that STX transactions were impacted by data from mempool endpoints not being guaranteed.
2. Added a note to all mempool endpoints in the openapi.yaml file


zone117x (changes requested):
Can you also add the comment to the `/extended/v1/address/{address}/mempool` endpoint?

trichards-hiro:
@zone117x I went ahead and added the note to the one endpoint I missed. Thank you for the good catch.

zone117x (commented):
Can we hold off on this PR until we see if https://github.com/hirosystems/stacks-blockchain-api/pull/1561 gets rid of the underlying problem that prompted this doc change?

trichards-hiro:
@zone117x Sure....I can hold off until we know whether the fix you are working on resolves this issue.

PR #1572, opened 2023-03-03
Fix/event replay perf

zone117x:
Removes the temporary pg table used during event-replay. This was originally in place to handle tsv files that contains 1) duplicate events, and 2) out of order events.

Now that we no longer support live appending to tsv files (only direct exports), we should no longer be dealing with these types of "dirty" tsv files. 

This should be a performance improvement now that postgres no longer has to serve reads from the temp table while also writing events during ingestion.


zone117x:
This will likely be built upon later on when we tackle event-replay perf. Leaving this PR open as draft for now.

PR #1578, opened 2023-03-13
Add arm64 build arg and update action versions

wileyj:
Two changes being proposed here:
1. adding an `arm64` platform for the docker image being built `platforms: linux/arm64, linux/amd64` - this only applies to the non-standalone image
2. update the actions to use latest versions (i.e. `docker/build-push-action@v2` -> `docker/build-push-action@v3`)

One thing to note is that in my hardcoded tests i had built for `linux/amd64` v2 and v3, but i felt it'll be better long-term to keep it simple and only build the latest amd64 version (same for arm64). this is reflected in the last commit made to my branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -881,6 +881,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           context: .
+          platforms: linux/arm64, linux/amd64, linux/amd64/v2, linux/amd64/v3
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           # Only push if (there's a new release on main branch, or if building a non-main branch) and (Only run on non-PR events or only PRs that aren't from forks)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,15 +18,15 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Use Node.js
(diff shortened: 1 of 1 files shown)


wileyj:
Note - keeping the base branch as `master` here since no code has changed, this only affects ci

zone117x (changes requested):
Thanks @wileyj! Can you make this so that the arm64 image is only built on releases (e.g. on master branch)? The docker image build steps ran for PRs are already pretty slow.

wileyj:
> Thanks @wileyj! Can you make this so that the arm64 image is only built on releases (e.g. on master branch)? The docker image build steps ran for PRs are already pretty slow.

I can look into it - but the goal of this PR was to not change the overall order of operations in the ci process here. 
it *may* be possible to assign a variable for that `platform` key in ci.yml, changing based on the type of action it is (i.e. a merge to master would build all arches, whereas non-master would only build amd64).

zone117x:
> it _may_ be possible to assign a variable for that `platform` key in ci.yml, changing based on the type of action it is (i.e. a merge to master would build all arches, whereas non-master would only build amd64).

@wileyj any luck with this? I think it's a blocker to merging this PR. Otherwise, I think this would be really nice to have.

smcclellan:
@wileyj Are you able to and interested in continuing this one?

wileyj:
> @wileyj Are you able to and interested in continuing this one?

absolutely, unless someone else wants to take it over. higher priority things on my plate at the moment, but this has been on the backburner for a while

smcclellan:
Closing this one for now. The version update is already in `master`, and the platform update here would need to be made conditional on the branch.

PR #1581, opened 2023-03-14
fix: optimize queries to retrieve BNS names

### LINKED ISSUE #1580: SPIKE: BNS `get-name` query is creating DB bottleneck (2 days)
The following query is reporting very high load times, even though indexes are complete and local tests show a healthy query plan. I suspect there is something else blocking the execution of this query, probably a bug with how queries are run inside the postgres transaction that envelops this query.

```
SELECT n.*, z.zonefile
        FROM names AS n
        LEFT JOIN zonefiles AS z USING (name, tx_id, index_block_hash)
        WHERE n.name = $1
          AND n.registered_at <= $2
          AND n.canonical = $3
          AND n.microblock_canonical = $4
        ORDER BY n.registered_at DESC,
          n.microblock_sequence DESC,
          n.tx_index DESC,
          n.event_index DESC
        LIMIT $5
```

**Acceptance Criteria**: Create an Issue for all work that needs to be done to resolve this issue.

rafaelcr:
This PR attempts to fix #1580 by grouping calls to get BNS names with latest status into a single query. It also optimizes the use of postgres transactions so we don't duplicate calls to get the latest chain tip block height.

### DIFF
--- a/src/api/routes/bns/names.ts
+++ b/src/api/routes/bns/names.ts
@@ -62,7 +62,7 @@ export function createBnsNamesRouter(db: PgStore, chainId: ChainID): express.Rou
     asyncHandler(async (req, res, next) => {
       const { name } = req.params;
       const includeUnanchored = isUnanchoredRequest(req, res, next);
-      const zonefile = await db.getLatestZoneFile({ name: name, includeUnanchored, chainId });
+      const zonefile = await db.getLatestZoneFile({ name: name, includeUnanchored });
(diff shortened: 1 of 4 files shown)


zone117x (approved):
Thanks @rafaelcr!

rafaelcr:
I think this PR helps with query performance but it doesn't solve the underlying locking issue BNS still has... I'm going to merge it to `develop` as you suggest tho 👍

PR #1591, opened 2023-03-22
fix: expand namespace discount column types to numeric

rafaelcr:


migrations/1679508966141_namespace-discount-bigint.js
```diff
@@ -0,0 +1,21 @@
...
+    type: 'numeric'
+  })
+  pgm.alterColumn('namespaces', 'no_vowel_discount', {
+    type: 'numeric'
+  })
+};
+
+exports.down = pgm => {
+  pgm.alterColumn('namespaces', 'nonalpha_discount', {
+    type: 'numeric'
```
rafaelcr, line 16: @zone117x I had to leave `numeric` here because if I set `integer` tests broke from "out of range for integer" when trying to cycle migrations... thoughts?

zone117x: I think it would have to be some `cast` expression to force the type to the smaller size. As long as things work we can probably leave it like this

zone117x (approved):

Issue #1590, opened 2023-03-22
Slow endpoint: stx_supply


CharlieC3:
This endpoint is rather slow compared to the others: https://api.hiro.so/extended/v1/stx_supply?unanchored=true

It ends up making some data on the front page of the Explorer lag a bit. Is there any way we can speed this endpoint up? Either through an optimized SQL query, or via caching?

rafaelcr:
I think a materialized view would give us the most gains and we actually wouldn't need to create a major release to do this, I believe, just a new migration that adds the view 👍🏻

zone117x:
Etag caching for this endpoint has been merged into develop. Next steps:
1. Once this makes it into a stable release and deployment, determine if the etag caching improvements are good enough.
2. If not good enough, we probably need to add this as to a materialized view.

PR #1594, opened 2023-03-23
Fix: `/stx_supply` caching

zone117x:
Fixes https://github.com/hirosystems/stacks-blockchain-api/issues/1590

Adds chaintip-based etag caching for the `/stx_supply` endpoints. This makes it so the pg query generally only needs to be ran once per block.

If the we still run into performance issues with this endpoint then we can add materialized view for it.

### DIFF
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -2346,7 +2346,7 @@ export class PgStore {
       if (result.length < 1) {
         throw new Error(`No rows returned from total supply query`);
       }
-      return { stx: BigInt(result[0].amount), blockHeight: atBlockHeight };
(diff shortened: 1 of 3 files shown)


src/datastore/pg-store.ts
```diff
@@ -2346,7 +2346,7 @@ export class PgStore {
       if (result.length < 1) {
         throw new Error(`No rows returned from total supply query`);
       }
-      return { stx: BigInt(result[0].amount), blockHeight: atBlockHeight };
+      return { stx: BigInt(result[0]?.amount ?? 0), blockHeight: atBlockHeight };
```
zone117x, line 2349: Note: also fix a minor bug where this endpoints throws and returns a 500 when no blocks have been received.

rafaelcr (approved):
LGTM

Issue #1595, opened 2023-03-23
Domain migration `stacks.co` to `hiro.so`


zone117x:
Update url references from `stacks.co` to `hiro.so` for seed nodes, API url references, and Explorer url reference.

https://github.com/search?q=repo%3Ahirosystems%2Fstacks-blockchain-api+.stacks.co&type=code

zone117x:
Can we close this or should we wait until those docs endpoints are updated? @CharlieC3 @LakshmiLavanyaKasturi ?

CharlieC3:
@zone117x I think this is done. I talked to Lavanya about it last week and it seems any stacks.co docs references left leads to content that has no plan to be migrated to hiro.so.

PR #1596, opened 2023-03-23
Fix: domain migration

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1595

* [x] toml references for seed nodes
* [x] API url references
* [x] Explorer url references

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1,8 +1,8 @@
 openapi: 3.0.2
 servers:
-  - url: https://stacks-node-api.mainnet.stacks.co/
+  - url: https://api.mainnet.hiro.so/
     description: Mainnet
-  - url: https://stacks-node-api.testnet.stacks.co/
+  - url: https://api.testnet.hiro.so/
     description: Testnet
   - url: http://localhost:3999/
     description: Local
@@ -95,7 +95,7 @@ paths:
         specified testnet address.
 
         The endpoint returns the transaction ID, which you can use to view the transaction in the
-        [Stacks Explorer](https://explorer.stacks.co/?chain=testnet). The tokens are delivered once the transaction has
+        [Stacks Explorer](https://explorer.hiro.so/?chain=testnet). The tokens are delivered once the transaction has
         been included in an anchor block.
 
         A common reason for failed faucet transactions is that the faucet has run out of tokens. If you are experiencing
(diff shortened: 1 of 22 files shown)


stacks-blockchain/Stacks-testnet.toml
```diff
@@ -1,7 +1,7 @@
 [node]
 rpc_bind = "0.0.0.0:20443"
 p2p_bind = "0.0.0.0:20444"
-bootstrap_node = "047435c194e9b01b3d7f7a2802d6684a3af68d05bbf4ec8f17021980d777691f1d51651f7f1d566532c804da506c117bbf79ad62eea81213ba58f8808b4d9504ad@testnet.stacks.co:20444"
+bootstrap_node = "029266faff4c8e0ca4f934f34996a96af481df94a89b0c9bd515f3536a95682ddc@seed.testnet.hiro.so:20444"
 # working_dir = "/tmp/stacks-blockchain-data"
```
zone117x, line 7: @CharlieC3 what about the reference below to `bitcoind.testnet.stacks.co`, should these be updated to a `hiro.so` domain?

docs/openapi.yaml
```diff
@@ -1,8 +1,8 @@
 openapi: 3.0.2
 servers:
-  - url: https://stacks-node-api.mainnet.stacks.co/
+  - url: https://api.mainnet.hiro.so/
     description: Mainnet
-  - url: https://stacks-node-api.testnet.stacks.co/
+  - url: https://api.testnet.hiro.so/
     description: Testnet
   - url: http://localhost:3999/
     description: Local
```
zone117x, line 8: @LakshmiLavanyaKasturi @CharlieC3 what about the `docs.stacks.co` references in this file and in various markdown files under `content/`? Should these be replaced with `docs.hiro.so`?

CharlieC3 (approved):
Just calling out here that the Explorer URL changes are a bit ahead of the migration. The Explorer isn't planned to be migrated until April 10th. But it shouldn't be a big issue to make these updates now.

LGTM! :shipit:

zone117x:
> @zone117x Could you also change the reference here? https://github.com/hirosystems/stacks-blockchain-api/blob/master/client/src/generated/runtime.ts#L16

This should be changed automatically during CI build and publish (since this is a generated file). Right now committing the generated client files results in a huge diff from lots of changed added since last time they were committed. For now I can do a one-off edit to that file. 

At some point we should either delete and not commit the generated client files, or do an isolated PR for the latest updates then enforce updates in new PRs somehow.

smcclellan (approved):

Issue #1391, opened 2022-10-27
Offline Rosetta node error connecting to postgres


zone117x:
Bug report from one of our partners. Offline Rosetta instances are throwing errors related to postgres and will not boot:

 > rosetta online nodes are coming up fine but offline nodes are not.
After a bit exploration, I found that we have to set new env var STACKS_API_MODE=offline for those nodes.
We did set, but no luck, still getting the same Pg connection error. AFAIK offline nodes does not need postgres, right?

> are there any more env vars that you are aware of which were recently added or you think which we may be missing? 4.1.x versions are running fine, the issue is with 6.x.x versions.

@rafaelcr is this something you can take a look at?

DSdatsme:
The offline rosetta instance keeps throwing the following error.

```bash
Pg connection failed: Error: connect ECONNREFUSED 127.0.0.1:5432, retrying..
```

PR #1599, opened 2023-03-27
fix: postgres should not be required in `STACKS_API_MODE=offline` mode

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1391

The dummy storage instance was not being used when initializing in offline mode, which was causing the postgres connection to retry forever.


rafaelcr (approved):
Thanks @zone117x

PR #1603, opened 2023-04-06
fix: sha256 tx cache etags to make them shorter

### LINKED ISSUE #1600: All TX endpoints show the cache is expired
Etag caching does not seem to be working as expected for TX endpoints.

For example, hitting this endpoint returns the following `cf-cache-status` of EXPIRED:
https://api.hiro.so/extended/v1/tx/0xe4661994dccf80c4268adf6a90931c8897b02176bc255e2e04d587e93f4a7f55
Screenshot 2023-03-27 at 1 14 49 PM

Instead, I would expect it to work like the `[status](https://api.hiro.so/extended/v1/status)` endpoint and return REVALIDATED like so:
Screenshot 2023-03-27 at 1 36 59 PM

I'm not sure if etag caching is broken for other endpoints too, so this should be verified across the board just to be sure.

rafaelcr:
Possibly fixes #1600

### DIFF
--- a/src/api/controllers/cache-controller.ts
+++ b/src/api/controllers/cache-controller.ts
@@ -1,9 +1,8 @@
 import { RequestHandler, Request, Response } from 'express';
 import * as prom from 'prom-client';
-import { bufferToHexPrefixString, logger, normalizeHashString } from '../../helpers';
+import { logger, normalizeHashString, sha256 } from '../../helpers';
 import { asyncHandler } from '../async-handler';
 import { PgStore } from '../../datastore/pg-store';
-import { sha256 } from 'bitcoinjs-lib/src/crypto';
 
 const CACHE_OK = Symbol('cache_ok');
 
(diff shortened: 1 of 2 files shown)


src/api/controllers/cache-controller.ts
```diff
@@ -3,6 +3,7 @@ import * as prom from 'prom-client';
 import { bufferToHexPrefixString, logger, normalizeHashString } from '../../helpers';
 import { asyncHandler } from '../async-handler';
 import { PgStore } from '../../datastore/pg-store';
+import { sha256 } from 'bitcoinjs-lib/src/crypto';
```
zone117x, line 6: Bitcoinjs-lib uses a slow js implementation of sha256. Can you use the built-in nodejs one?
```js
import { createHash } from 'node:crypto'

function sha256(content) {  
  return createHash('sha256').update(content).digest('hex')
}
```

zone117x (approved):
Thanks @rafaelcr!

PR #1604, opened 2023-04-06
fix: materialized view refresh after IBD height has passed

rafaelcr:
Fixes remaining issue for #1316 caused by a faulty materialized view refresh logic that was looking at the `chain_tip` view to determine whether or not we should resume refreshes after IBD (causing them never to be refreshed).


zone117x (approved):
LGTM!

PR #1605, opened 2023-04-07
fix: log level issues

### LINKED ISSUE #1452: log level is always "http"
**Describe the bug**
The log level printed by the API always reads "http".
e.g. `{"level":"http","message":"HTTP GET /extended/v1/tx/0x7756798e9f6a0f1104c56b82d232b5ed9e0de950191a84953f29e96657ccfaf4","req":{"headers":{"accept":"application/json, text/plain,...`

**Expected behavior**
Instead, the log level should represent different severity scopes, like debug, info, warning, error, etc.
Ideally, the scopes would be compatible with what Grafana Loki recognizes here: https://grafana.com/docs/grafana/latest/explore/logs-integration/#log-level

Correcting the log level would greatly help with debugging and log searches.

csgui:
This PR fixes #1452.

Changes:
- default logging level is set to `info`;
- 4xx statuses code are logged as a warning.


src/api/init.ts
```diff
@@ -155,10 +155,20 @@ export async function startApiServer(opts: {
...
-      statusLevels: {
-        error: 'error',
-        warn: httpLogLevel ?? 'http',
-        success: httpLogLevel ?? 'http',
+      statusLevels: false,
+      level: function (req, res) {
+        let level = '';
+        if (res.statusCode >= 100) {
+          level = 'info';
+        }
```
rafaelcr, line 163: I think you could simplify this by setting this as `info` initially:
```suggestion
        let level = 'info';
```

rafaelcr (commented):
Adding @CharlieC3 as reviewer too

src/api/init.ts
```diff
@@ -155,10 +155,20 @@ export async function startApiServer(opts: {
       format: logger.format,
       transports: logger.transports,
       metaField: (null as unknown) as string,
-      statusLevels: {
-        error: 'error',
-        warn: httpLogLevel ?? 'http',
-        success: httpLogLevel ?? 'http',
+      statusLevels: false,
```
CharlieC3, line 158: If you set this to `true`, I believe it will essentially implement the same logic as below [according to the docs](https://www.npmjs.com/package/express-winston):
> If you set statusLevels to true express-winston will log sub 400 responses at info level, sub 500 responses as warnings and 500+ responses as errors.

csgui: Interesting... this passed by my side. Way simpler than customizing every response code. 

👍

CharlieC3 (approved):
LGTM! :shipit:

zone117x (approved):
I think this should probably be merged into `develop` branch (i.e. I don't think there's a pressing need to release this immediately like a hotfix). Otherwise looks great! Thanks @csgui!

rafaelcr (approved):
LGTM 🏁

Issue #1065, opened 2022-02-15
Fix RPC proxy middleware errors showing up in logs


zone117x:
Expected-looking http errors occurring inside the RPC proxy are showing up as `level: error` logs.

Example: 
`Response 400 during POST /v2/transaction to extra endpoint http://stacks-node-follower-events-blue-0.stacks-node-follower-events-blue.prod-stacks.svc.cluster.local:20443/v2/transactions`

MHassanTariq:
@zone117x what is the required behavior for this issue? 
With the current approach, the [specified error](https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/api/routes/core-node-rpc-proxy.ts#L209-L211) is logged [here](https://github.com/hirosystems/stacks-blockchain-api/blob/master/src/helpers.ts#L195). Are we supposed to ignore the http error even if the `p.value.ok` is not true? Or should such errors be shown as `level: warning` logs?

zone117x:
These should probably be `warning` level logs.

PR #1612, opened 2023-04-14
fix: warning logger level for RPC proxy errors

csgui:
Fix for issue #1065.

400 errors should be logged as `warning`.


zone117x (approved):
Thanks @csgui!

Issue #1566, opened 2023-03-01
Log output cleanup


zone117x:
* [x] `[socket.io] ...` messages should include a property to easily filter by, and most of them should be changed from `info` to `verbose` level
* [x] `[stacks-node event] ...` messages should include a property to easily filter by
* [x] `getPagingQueryLimit` error messages like `'limit' must be equal to or less than 50` are logged without any error level, these should be http-400 messages

zone117x:
@csgui is this something you can take on? The migration from winston to pino that Rafael mentioned above can be done in a follow-up PR.

csgui:
@zone117x Sure. I'll have a look. Assigning it to me.

csgui:
Hello @zone117x and @rafaelcr !

> * `[socket.io] ...` messages should include a property to easily filter by

Does `socket.io` text can't be used as a filter? I.e.: if the property `message` contains `socket.io` filter by that.

> * `[stacks-node event] ...` messages should include a property to easily filter by

Does `stacks-node event` text can't be used as a filter? I.e.: if the property `message` contains `stacks-node event` filter by that.

PR #1613, opened 2023-04-17
fix: log cleanup

csgui:
Fixes for issue https://github.com/hirosystems/stacks-blockchain-api/issues/1566.

Added:

1 - metadata component to logs.
2 - `getPagingQueryLimit` error handling.

### DIFF
--- a/src/api/routes/address.ts
+++ b/src/api/routes/address.ts
@@ -533,7 +533,7 @@ export function createAddressRouter(db: PgStore, chainId: ChainID): express.Rout
       const untilBlock = parseUntilBlockQuery(req, res, next);
 
       const nftListResponse = await db.sqlTransaction(async sql => {
-      const blockHeight = await getBlockHeight(untilBlock, req, res, next, db);
+        const blockHeight = await getBlockHeight(untilBlock, req, res, next, db);
 
         const response = await db.getAddressNFTEvent({
           stxAddress,
--- a/src/api/routes/address.ts
+++ b/src/api/routes/address.ts
@@ -214,7 +214,15 @@ export function createAddressRouter(db: PgStore, chainId: ChainID): express.Rout
       const principal = req.params['principal'];
       validatePrincipal(principal);
       const untilBlock = parseUntilBlockQuery(req, res, next);
-      const limit = getPagingQueryLimit(ResourceType.Tx, req.query.limit);
+
+      let limit: number;
+      try {
+        limit = getPagingQueryLimit(ResourceType.Tx, req.query.limit);
+      } catch (error: any) {
+        res.status(400).json({ error: error.message });
+        return;
+      }
+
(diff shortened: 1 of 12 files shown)


rafaelcr (approved):
Nice work!

src/api/routes/address.ts
```diff
@@ -317,7 +325,15 @@ export function createAddressRouter(db: PgStore, chainId: ChainID): express.Rout
...
           blockHeight = await getBlockHeight(untilBlock, req, res, next, db);
         }
-        const limit = getPagingQueryLimit(ResourceType.Tx, req.query.limit);
+
+        let limit: number;
+        try {
+          limit = getPagingQueryLimit(ResourceType.Tx, req.query.limit);
+        } catch (error: any) {
+          res.status(400).json({ error: error.message });
+          return;
```
zone117x, line 334: Do you know why this isn't being handled as expected in the error handling code for `InvalidRequestError` types?

I.e. that `getPagingQueryLimit` throws an error which is supposed to indicate to the rest of the route handling that it's a 400, see:
* https://github.com/hirosystems/stacks-blockchain-api/blob/05d6a9e241d56b05066e8dc4895b4ca0a5086c75/src/api/init.ts#L303
and
* https://github.com/hirosystems/stacks-blockchain-api/blob/05d6a9e241d56b05066e8dc4895b4ca0a5086c75/src/api/init.ts#L325

src/helpers.ts
```diff
@@ -192,7 +192,9 @@ const defaultLogLevel: LogLevel = (() => {
 export const logger = winston.createLogger({
   level: defaultLogLevel,
   exitOnError: false,
+  defaultMeta: { component: 'core-api' },
   format: winston.format.combine(
+    winston.format.metadata(),
```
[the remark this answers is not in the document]
csgui: It adds a `metada` object to the log output where it can be used for API components filtering purpose. For instance:

`{"level":"info","message":"API server listening on: http://127.0.0.1:3999","metadata":{"component":"core-api"},"timestamp":"2023-04-17T16:49:45.955Z"}`
or
`{"level":"verbose","message":"new connection: rVVvoH_n-He7JvDIAAAB","metadata":{"component":"socket-io"},"timestamp":"2023-04-17T16:54:26.597Z"}`

src/api/routes/address.ts
csgui: I don't have an immediate answer for that. But probably due to Winston misconfiguration.

Should I take care of that? Or would be better to start working on the new logging using Pino?

zone117x: There's a few other areas depending on the correct behavior for InvalidRequestError handling. I think it makes sense to try to fix that before the pino logging migration, since that might take a while to land, but it can be done in PR after this one (we'd just want to make sure to revert these extra try/catches)

zone117x (approved):

PR #1618, opened 2023-04-18
fix(subnets): use boot address for burnchain-tx signer

### LINKED ISSUE #1616: [Subnets] Synthetic subnet burnchain-tx parsing causes address nonce bug
From @hugocaillard:
> On subnet, the “nonces” endpoint
> (extended/v1/address/ST2REHHS5J3CERCRBEPMGH7921Q6PYKAADT7JP2VB/nonces?unanchored=true) returns:
> ```json
> {
>   "last_executed_tx_nonce": 0,
>   "last_mempool_tx_nonce": null,
>   "possible_next_nonce": 1,
>   "detected_missing_nonces": []
> }
> ```
> For an address that hasn’t executed any transaction yet, but that received NFTs via a deposit. (For an address that didn’t interact at all with the subnet, the response is "possible_next_nonce": 0  as expected).

This is caused by the subnet burnchain-tx parsing code:
https://github.com/hirosystems/stacks-blockchain-api/blob/1f7ad4119ef7c0a19b4cbe92ec0ad21a878f6d41/src/event-stream/reader.ts#L205

The tx "sender" (i.e. signer) is specified as the address associated with the asset transfer. This throws off the results of the `/nonces` query, because it will see this as a regular, nonce-incrementing tx.

The fix is to change the tx sender to another address, probably the subnet boot contract principal.

zone117x:
Fixes #1616

Use the boot address as the tx signer for synthetic subnet burnchain-tx parsing. Previously it was using the address associated with the asset transfer, which would cause other queries to view those addresses as having issued nonce-incrementing txs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
       - master
       - develop
       - beta
+      - subnets
     tags-ignore:
       - '**'
     paths-ignore:
--- a/src/event-stream/reader.ts
+++ b/src/event-stream/reader.ts
@@ -48,6 +48,8 @@ import {
   bufferToHexPrefixString,
   hexToBuffer,
   SubnetContractIdentifer,
+  BootAddress,
+  getBootAddressForChain,
 } from '../helpers';
 import {
   TransactionVersion,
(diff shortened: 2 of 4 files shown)


rafaelcr (approved):
LGTM

Issue #1602, opened 2023-04-02
subnet block sync doesn't start when BNS enabled


deantchi:
It was discovered that BNS data needs to be disabled in the api in order for the sync to start.  I'm not sure if BNS is needed or supported in the context of subnets but I had to [disable it](https://github.com/hirosystems/k8s/commit/50d8b1946d0c5f4af0060de3122255ec3295d777) to get the subnet node syncing.

deantchi:
That's a good idea. As the bns setting is stated to be optional in the docs. We should mention the possibility of conflicting flags.

smcclellan:
Next: Add env variable (or lookup) to determine subnet, then throw error if this comes up.

PR #1619, opened 2023-04-18
add check for stacks node type before BNS import

csgui:
Issue: #1602 .

Summary: BNS imports are mean to run for L1 stacks node. When running for Subnets the node sync is not starting.

Solution: Create an environment variable that explicits the underlying node type. And check for the node type before starting the BNS importer.

### DIFF
--- a/src/import-v1/index.ts
+++ b/src/import-v1/index.ts
@@ -35,8 +35,8 @@ const SUBDOMAIN_BATCH_SIZE = 2000;
 const STX_VESTING_BATCH_SIZE = 2000;
 
 const enum StacksNodeType {
-  L1 = 'l1',
-  Subnet = 'subnet'
+  L1 = 'L1',
+  Subnet = 'subnet',
 }
 
 class LineReaderStream extends stream.Duplex {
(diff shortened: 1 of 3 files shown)


src/import-v1/index.ts
```diff
@@ -548,6 +553,12 @@ export async function importV1TokenOfferingData(db: PgWriteStore) {
 }
 
 export async function handleBnsImport(db: PgWriteStore) {
+  const stacksNodeType = process.env.STACKS_NODE_TYPE;
+  if (stackNodeType === StacksNodeType.Subnet) {
+    logger.error('BNS imports should not be enabled for a Subnet');
```
zone117x, line 558: With this behavior, we will log and keep running (as opposed to exiting with error). That seems okay to me, but we should make the error a bit more explicit in that case:
> `BNS imports should not be enabled for a Subnet, skipping ...`

src/import-v1/index.ts
```diff
@@ -548,6 +553,12 @@ export async function importV1TokenOfferingData(db: PgWriteStore) {
 }
 
 export async function handleBnsImport(db: PgWriteStore) {
+  const stacksNodeType = process.env.STACKS_NODE_TYPE;
```
zone117x, line 556: We should also add this to the `.env` file where we informally document env vars

zone117x (approved):
LGTM!

PR #1625, opened 2023-04-20
Merge subnets into develop

zone117x:
An initial subnet-support PR had accidentally already been merged into the develop branch a while ago, and is difficult to revert.

So, we might as well merge in the latest subnet support into develop. This means the next release cut from develop will support subnets. 

Note: if additional API changes related to subnets are needed, then those will continue to go only into the `subnets` branch of the API.


rafaelcr (approved):
LGTM!

PR #1636, opened 2023-04-28
fix: support Stacks 2.2 force pox-2 unlocks

### LINKED ISSUE #1634: Update locked amount to reflect Stacks 2.2 changes
In Stacks 2.2 the `/new_block` payload has a new property `pox_v2_unlock_height` which should be stored and used in account balance and stacking related queries. Delegated and locked STX should be treated as unlocked after the specified height, similar to how pox-v1 force unlocks via the `pox_v1_unlock_height` value.

zone117x:
Closes #1634


csgui (approved):
LGTM!

Issue #1584, opened 2023-03-15
API does not reconnect to stacks-blockchain node in some situations


CharlieC3:
**Describe the bug**
When the postgres database and stacks-node are terminated while an API writer continues running, the API writer _sometimes_ is unable to re-connect to the stacks-blockchain nodes after it and the postgres database come back online _after being down for an extended period (~30 minutes or so)._
Despite this, the API logs still continue to indicate the API writer is attempting to connect to the stacks-node as one would expect:
```
Warning: failed to connect to node RPC server at mainnet-stacks-blockchain-snapshot:20443
```

However, I still think there is something wrong here. If I try restarting the stacks-node again, the API will continue printing the above log and won't be able to connect. If I try restarting the API writer after the stacks-node and postgres db have already come back online, it starts working.
Thus, this test seems to indicate the issue resides in the API rather than the stacks-node.

zone117x:
If you exec into the pod when this is happening, is `mainnet-stacks-blockchain-snapshot` reachable with curl? I'm wondering if there is some DNS caching issue somewhere.

CharlieC3:
I'll test that next time it happens, but it doesn't occur every day. More like every two weeks or so. However, we explicitly have DNS caching _disabled_, so I would not expect it to be related to DNS caching.
In the meantime, anything you can find here would be a big help.

rafaelcr:
Next steps: set up some alerts to get more info about this

CharlieC3:
I was able to find this happening again. I've confirmed a few interesting things:

* DNS caching is not a problem (IPs matched)
* The API writer could not ping the stacks-node on port `20443` because `20443` was not being opened by the node
* the API writer could not ping localhost:3999 because `3999` was not open
* Restarting the API writer did not solve the problem
* Restarting both the API writer and stacks-node fixed the problem

What's strange is this deployment _was_ previously working, then broke without the API writer ever restarting. This means port `3999` was open, but the API closes port `3999` if it cannot contact the stacks-node for a while and enters this circular dependency.
I think that’s the actual bug, and the solution is the keep port 3999 open and responding even when the stacks-node goes down.

As for why the stacks-node needed to be restarted, @rafaelcr and I think there may have been a retry limit on the stacks-node before giving up (unconfirmed). There should not be a limit to the number of retries in that logic, but anyways it's outside the scope of this issue.

CharlieC3:
[...]

PR #1640, opened 2023-05-01
fix: option to skip RPC request during init which may avoid startup deadlocks

zone117x:
May close https://github.com/hirosystems/stacks-blockchain-api/issues/1584

For context, see comment https://github.com/hirosystems/stacks-blockchain-api/issues/1584#issuecomment-1530010417


rafaelcr (approved):
LGTM

PR #1644, opened 2023-05-04
feat: Stacks 2.4 / pox-3

zone117x:
Support Stacks 2.4 / PoX-3

- [x] Sql migration to create new table for pox3 events (should be able to copy the schema from the `pox2_events` table)
- [x] Ingest/parse pox-3 events and in store in new `pox3_events` table (should be able to re-use [same code from pox-2](https://github.com/hirosystems/stacks-blockchain-api/blob/feat/stacks-2.4/src/event-stream/pox2-event-parsing.ts#L0-L1))
- [x] Update sql queries for locked balances and other stacking related data to also read for the new `pox3_events` table
- [x] Update Rosetta to generate stx_unlock ops from `pox3_events` table
- [x] Update Rosetta to generate stacking transactions against pox-3
- [x] Update `/tests-2.1` to instead test Stacks 2.4 / `pox-3`
- [x] We can probably just remove the `/tests-2.1-transition` tests now

### Notes:
* No breaking schema changes or event-replay will be required for this upgrade.
* So far, PoX-3 events are all identical in shape to PoX-2 events, so much of the code implemented for pox-2 is re-used (e.g. functions, types, sql) -- sometimes aliased, sometimes there will be some pox-3 code with references to pox-2, and some sql queries are copy and pasted. These can be refactored and abstracted for clarity later on.
* Many of the test suites were running against a stacks-node booted only into epoch 2.0 -- these have been upgraded so that almost all test suites are ran against epoch 2.4. Many of the LoC changes in tests were to account for issues when making these upgrades.


package.json
```diff
@@ -78,8 +79,8 @@
         "prerelease": true
       },
       {
-        "name": "stacks-2.1",
-        "channel": "stacks-2.1",
+        "name": "stacks-2.4",
+        "channel": "feat/stacks-2.4",
```
rafaelcr, line 83: You have these reversed, `name` should be the branch name

src/api/routes/rosetta/construction.ts
```diff
@@ -399,7 +400,7 @@ export function createRosettaConstructionRouter(db: PgStore, chainId: ChainID):
         res.status(400).json(RosettaErrorsTypes.missingTransactionSize);
         return;
       }
-      const feeValue = (BigInt(feeInfo) * BigInt(options.size)).toString();
+      const feeValue = Math.round(Number(feeInfo) * Number(options.size) * 1.5).toString();
```
rafaelcr, line 403: I might be missing something with pox-3, but why is this fee value function being adjusted?

package.json
zone117x: Thanks! Fixed

src/test-utils/test-helpers.ts
```diff
@@ -142,15 +142,21 @@ export async function standByForNextPoxCycle(): Promise<CoreRpcPoxInfo> {
   return lastPoxInfo;
 }
 
-export async function standByForPoxCycle(): Promise<CoreRpcPoxInfo> {
-  const firstPoxInfo = await testEnv.client.getPox();
+export async function standByForPoxCycle(
```
rafaelcr, line 145: Thanks for doing these changes haha we had a mess here 💯

rafaelcr (approved):
Looks good to me!

src/api/routes/rosetta/construction.ts
rafaelcr: OK gotcha, thanks

zone117x:
Close in favor of https://github.com/hirosystems/stacks-blockchain-api/pull/1650

csgui (commented):
Left some comments... my main remark is that there are some references to PoX2 that can be changed to PoX3. Or even PoX, in general. To an user not much familiar with Stacks that can sound as an issue. Do we have available time to improve that?

Anyway, great work. Thanks for the effort on this task!

PR #1651, opened 2023-05-09
Chore: update rosetta dockerfile

wileyj:
## Description

This close to a full rewrite of the built image. The filename is also changed to a more appropriate `rosetta.Dockerfile`

- runit is removed
- `/entrypoint.sh` now starts all processes via single script
- all logs are printed to stdout
- remove printf in favor of heredoc
- add a script to seed the chainstate when env var `SEED_CHAINSTATE=true`
- adds a postgres init script in case of genesis sync and DB != `postgres`

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

N/A

## Testing information
Tested by building the docker image and launching with:
- Testnet without seeding the chainstate
- Testnet with seeding the chainstate
- Mainnet without seeding the chainstate
- Mainnet with seeding the chainstate

In all cases, the chain progressed - of note, seeding the chainstate by supplying the env var: `SEED_CHAINSTATE=true` only works if the archives for the defined versions exists (at this moment, there is not 2.3.0.0.2 archive available, but was tested with 2.3.0.0.1). Also note that the time to process an archive can be highly variable depending on the system. 

The other caveat to this change is that when `SEED_CHAINSTATE=true`, the container will need to be launched with `--shm-size`. 
ex: `--shm-size=256MB`

Else, postgres will crash with an error regarding "out of space" (this applies to memory, not disk. details here: https://stackoverflow.com/questions/56751565/pq-could-not-resize-shared-memory-segment-no-space-left-on-device )


docker/rosetta.Dockerfile
```diff
@@ -0,0 +1,300 @@
...
+export STACKS_LOG_JSON=1
+# configure postgres and start it
+mkdir -p "${PGDATA}" || exit 1
+chown -R postgres:postgres "${PGDATA}" || exit 1
+gosu postgres /usr/lib/postgresql/${PG_VERSION}/bin/initdb -D "${PGDATA}" --wal-segsize=512  || exit 1
+echo "host all all all trust" >> "$PGDATA/pg_hba.conf" || exit 1
+gosu postgres /usr/lib/postgresql/${PG_VERSION}/bin/pg_ctl start -W -D ${PGDATA} -o "-c listen_addresses='*'"
+# give postgres 10s to startup
+echo "waiting 10s for postgres to start"
+sleep 10
```
zone117x, line 136: I think if you change the `pg_ctl start` flag `-W` to `-w` (lowercase) then it will wait correctly for startup, and you could remove this sleep.

docker/rosetta.Dockerfile
```diff
@@ -0,0 +1,300 @@
...
+fi
+# create DB/schema if using other than default 'postgres'
+if [[ "${SEED_CHAINSTATE}" = "false" && "${PG_DATABASE}" != "postgres" ]]; then
+    /scripts/postgres-initdb.sh || exit 1
+fi
+# set chain_id based on network
+case "${STACKS_NETWORK}" in
+    testnet)
+        export STACKS_CHAIN_ID=0x80000000
+        export V2_POX_MIN_AMOUNT_USTX=90000000260
```
zone117x, line 149: This `V2_POX_MIN_AMOUNT_USTX` is no longer in use and can be removed (looks like we still have references in the docs that need cleaned up, but it's not used anywhere in the code)

zone117x (approved):
Thanks @wileyj! This looks super helpful. Awesome way to use the archives 💯 

Left a couple minor comments, otherwise LGTM!

zone117x (commented):
Oh there's also a few references to the file that need to be updated with the rename, this CI build in particular here:
https://github.com/hirosystems/stacks-blockchain-api/blob/9dc72488bff40de008d4245172259185840fc670/.github/workflows/ci.yml#L900

The others are in markdown docs

docker/rosetta.Dockerfile
wileyj: Indeed it does!  I had changed that line from a `postmaster` exec (where -w doesn't work), but i like this idea - i'll have to re-test it, but shouldn't be an issue to update. for whatever reason, `-w` when using postmaster was throwing an error.

docker/rosetta.Dockerfile
wileyj: Thanks for confirming! it will be removed

wileyj:
> Oh there's also a few references to the file that need to be updated with the rename, this CI build in particular here:
> 
> https://github.com/hirosystems/stacks-blockchain-api/blob/9dc72488bff40de008d4245172259185840fc670/.github/workflows/ci.yml#L900
> 
> The others are in markdown docs

got it - i'll address when i make the other changes

docker/rosetta.Dockerfile
wileyj: updated and tested with testnet, all good

zone117x (approved):
Thanks again @wileyj!

Issue #1555, opened 2023-02-15
Rosetta cli construction fails while trying to run stacks-blockchain-api with mocknet


HassanAmed:
## Describe the bug
I tried to run stacks-api-blockchain api as mocknet using this command 
`npm run dev:integrated` which failed
I guess the problem is this, in `docker/docker-compose.dev.rosetta.cli.yaml`
`command: ${CMD}` 
`volume: ${OUTPUT}` 
`${CMD}` and `${OUTPUT}` are undefined
 I get this output 
 ```
WARN[0000] The "OUTPUT" variable is not set. Defaulting to a blank string. 
WARN[0000] The "CMD" variable is not set. Defaulting to a blank string. 
1 error(s) decoding:

* error decoding 'Volumes[0]': invalid empty volume spec
 ```
## Steps To Reproduce

1. `git clone https://github.com/hirosystems/stacks-blockchain-api.git`
2. `npm run dev:integrated`

## Expected behavior

should successfully run stacks-blockchain-api with mocknet

## Environment

 - OS: macOS Monterey M1
 - Rust version: rustc 1.64.0

## Additional context
When i set 
`command: /bin/rosetta-cli --configuration-file /app/rosetta-config-docker.json check:construction`
`volume: - ./rosetta-output-construction:/app/rosetta-output`
 in `docker/docker-compose.dev.rosetta.cli.yaml` it runs successfully

rafaelcr:
Hi @HassanAmed we will update documentation and tooling because this is a very old command that should be removed in favor of using Clarinet

neilmehta31:
Hi @rafaelcr, I would like to fix this. Does changing the command and volume to 
`command: /bin/rosetta-cli --configuration-file /app/rosetta-config-docker.json check:construction`
`volume: - ./rosetta-output-construction:/app/rosetta-output`
fix the issue. I was able to run the stacks-blockchain-api successfully after the change. Let me know if anything else needs to be modified. I'll make a PR with the updates

rafaelcr:
Hi @neilmehta31 that's great! Please send the PR so we can review, thanks

neilmehta31:
@rafaelcr Please review the PR : https://github.com/hirosystems/stacks-blockchain-api/pull/1647

neilmehta31:
[...]

PR #1652, opened 2023-05-10
Fix: rosetta cli construction command and volumes updated

neilmehta31:
Description
Refer: https://github.com/hirosystems/stacks-blockchain-api/issues/1555
Changed the commands to run the stacks-blockchain-api in `docker-compose.dev.rosetta-cli.yml` to
`command: /bin/rosetta-cli --configuration-file /app/rosetta-config-docker.json check:construction`
`volume: - ./rosetta-output-construction:/app/rosetta-output`

To note: values hardcoded here to run the rosetta-cli on mocknet

### DIFF
--- a/docker/docker-compose.dev.rosetta-cli.yml
+++ b/docker/docker-compose.dev.rosetta-cli.yml
@@ -4,6 +4,6 @@ services:
     build:
       context: ../rosetta-cli-config
       dockerfile: docker/Dockerfile
-    command: ${CMD}
+    command: /bin/rosetta-cli --configuration-file /app/rosetta-config-docker.json check:construction
     volumes:
-      - ${OUTPUT}
+      - ./rosetta-output-construction:/app/rosetta-output
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,7 +162,7 @@ jobs:
 
       - name: Upload coverage to Coveralls
         uses: coverallsapp/github-action@master
-        if: always()
+        if: ${{ false }}
         with:
           github-token: ${{ secrets.github_token }}
           flag-name: run-${{ github.job }}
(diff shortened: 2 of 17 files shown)


docker/docker-compose.dev.rosetta-cli.yml
```diff
@@ -4,6 +4,6 @@ services:
     build:
       context: ../rosetta-cli-config
       dockerfile: docker/Dockerfile
-    command: ${CMD}
+    command: /bin/rosetta-cli --configuration-file /app/rosetta-config-docker.json check:construction
     volumes:
-      - ${OUTPUT}
+      - ./rosetta-output-construction:/app/rosetta-output
```
zone117x, line 9: The only intended purpose of this dockerfile is for integration tests, which this change appears will break: https://github.com/hirosystems/stacks-blockchain-api/blob/9dc72488bff40de008d4245172259185840fc670/src/tests-rosetta-cli-construction/validate-construction.ts#L45

It sounds like from issue https://github.com/hirosystems/stacks-blockchain-api/issues/1555 that you're trying to get `npm run dev:integrated` to work? As Rafael mentioned, this is a very old command that should be removed. 

Can you give more context into what you are trying to achieve? Then we can point you towards the intended workflow, or more appropriate areas to modify.

zone117x: Also note that when you change target git branches for a PR, you need to rebase your commit on top the new branch. Otherwise you end up with a PR file diff that has changes other than your own.

neilmehta31: @zone117x As you pointed out the issue #1555, the way I thought to remove the command was to replace it with the updated one. If this breaks the integration tests, this PR stands NULL, and I will close it with your confirmation. 

I tried to rebase the changes on top of my changes but couldn't achieve that. I will look into it for future reference. Thanks for your help.
Also, could you please suggest any open issue that I could work on? I wanna contribute to the repo.

PR #1653, opened 2023-05-10
Added headers to filter the API traffic

LakshmiLavanyaKasturi:
Added headers to WebSockets to understand the API traffic.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -33,7 +33,11 @@ Here is example code that subscribes to updates for a specific Stacks address:
 import { connectWebSocketClient } from '@stacks/blockchain-api-client';
 
 // for testnet, replace with wss://api.testnet.hiro.so/
-const client = await connectWebSocketClient('wss://api.mainnet.hiro.so/');
+const client = await connectWebSocketClient('wss://api.mainnet.hiro.so/', {
+  headers: {
+    'x-hiro-product': 'blockchain-api-client',
+  },
+});
 
 const sub = await client.subscribeAddressTransactions('ST3GQB6WGCWKDNFNPSQRV8DY93JN06XPZ2ZE9EVMA', event =>
   console.log(event);
@@ -54,7 +58,11 @@ const socketUrl = "https://api.mainnet.hiro.so/";
 const socket = io(socketUrl, {
   transports: [ "websocket" ]
 });
-const sc = new stacks.StacksApiSocketClient(socket);
+const sc = new stacks.StacksApiSocketClient(socket, {
+  headers: {
+    'x-hiro-product': 'blockchain-api-client',
+  },
+});
 
 sc.subscribeAddressTransactions('ST3GQB6WGCWKDNFNPSQRV8DY93JN06XPZ2ZE9EVMA');
 ```


client/README.md
```diff
@@ -33,7 +33,11 @@ Here is example code that subscribes to updates for a specific Stacks address:
 import { connectWebSocketClient } from '@stacks/blockchain-api-client';
 
 // for testnet, replace with wss://api.testnet.hiro.so/
-const client = await connectWebSocketClient('wss://api.mainnet.hiro.so/');
+const client = await connectWebSocketClient('wss://api.mainnet.hiro.so/', {
+  headers: {
+    'x-hiro-product': 'blockchain-api-client',
+  },
+});
```
zone117x, line 40: I don't think this is invalid code, there's no second parameter where headers can be specified. In typescript this won't compile, in javascript this second param will be ignored: 
https://github.com/hirosystems/stacks-blockchain-api/blob/9dc72488bff40de008d4245172259185840fc670/client/src/ws/index.ts#L318-L322

Also it looks like according to the WebSocket spec, clients are not permitted to set request headers, see https://stackoverflow.com/a/4361358/794962

client/README.md
```diff
@@ -54,7 +58,11 @@ const socketUrl = "https://api.mainnet.hiro.so/";
 const socket = io(socketUrl, {
   transports: [ "websocket" ]
 });
-const sc = new stacks.StacksApiSocketClient(socket);
+const sc = new stacks.StacksApiSocketClient(socket, {
+  headers: {
+    'x-hiro-product': 'blockchain-api-client',
+  },
+});
```
zone117x, line 65: Same here, a second parameter isn't valid code: https://github.com/hirosystems/stacks-blockchain-api/blob/9dc72488bff40de008d4245172259185840fc670/client/src/socket-io/index.ts#L32-L37

While socket.io has an option that we could use called ["extraHeaders"](https://socket.io/docs/v4/client-options/#extraheaders), they are not actual _http request_ headers and thus will not show up in the metric dashboards. It may be possible in the socket.io server code to detect `extraHeaders` and inject them into the http request such that it is reported in the logs, but that would need some investigation and testing.

zone117x (changes requested):
These changes are currently not valid code. And because custom http headers are not permitted in the websocket protocol, it would take some research and testing to figure out if it's even possible for us to report custom http headers over websockets somehow.

I'm assuming we would do something like add a custom payload to an initial handshake payload, then on server-side we'd try to inject them into the http reporting such that they look like custom http headers (unclear to me how possible or practical this is).

client/README.md
rafaelcr: There might be a way for us to modify the client code to always include this header instead... not sure if that could be auto-generated though, we might need to do that manually

LakshmiLavanyaKasturi:
> These changes are currently not valid code. And because custom http headers are not permitted in the websocket protocol, it would take some research and testing to figure out if it's even possible for us to report custom http headers over websockets somehow.
> 
> I'm assuming we would do something like add a custom payload to an initial handshake payload, then on server-side we'd try to inject them into the http reporting such that they look like custom http headers (unclear to me how possible or practical this is).

@sabbyanandan: ^

PR #1699, opened 2023-07-26
Added an example to get the contracts by trait

LakshmiLavanyaKasturi:
Added an example for trait_abi to the endpoint https://docs.hiro.so/api#tag/Smart-Contracts/operation/get_contracts_by_trait


docs/openapi.yaml
```diff
@@ -976,6 +976,15 @@ paths:
...
       operationId: get_contracts_by_trait
+      requestBody:
+        description: Trait Application Binary Interface(ABI) in the JSON format
+        required: true
+        content:
+          application/json:
+            schema:
+              type: string
+            example: 
+              $ref: ./entities/contracts/contracts-by-trait.example.json
```
zone117x, line 987: I'm not sure if this is correct. The endpoint is `GET /extended/v1/contract/by_trait?trait_abi=<value>` where `<value>` is the [URI encoded](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) JSON-stringified ABI object.

So rather than `requestBody` here (which I think is mostly for POST requests), seems like it should be something like:
```
parameters:
  - in: query
    name: by_trait
    content:
      application/json:
        schema:
          type: object
          properties: ...
```

This suggestion was just from a cursory look over how to specify json in a query param -- you should double check if this is actually the expected approach.

zone117x: If OpenAPI schema doesn't actually allow specifying a JSON-stringified query param, then you may just have to specify a string query param and document that it needs to be the JSON-stringified object, e.g.:
```
parameters:
  - in: query
    name: by_trait
    description: JSON stringified ABI object e.g. `JSON.stringify(abi)`
    schema:
      type: string
```

docs/entities/contracts/contracts-by-trait.example.json
rafaelcr, line 1: The `.example.json` files are used mainly for response bodies... I think in this case the example should be a string on the `openapi.yaml` file

docs/openapi.yaml
LakshmiLavanyaKasturi: I tried to add the following code to the yaml file, but I don't see any change in the output. May I know if this is the right way to add parameters to the OpenAPI.yaml?
```
parameters:
        - name: trait_abi
          in: query
          description: JSON stringified ABI object e.g. `JSON.stringify(abi)`.
          required: true
           schema:  # The request schema
              type: object  # JSON object type
              properties:
                functions:
                  type: array
                  items:
                    type: object
                    properties:
                      access:
                        type: string
                      args:
                        type: array
                        items: {}
                      name:
                        type: string
                      outputs:
                        type: object
                        properties:
                          type:
                            type: object
                            properties:
                              response:
                                type: object
                                properties:
                                  ok:
                                    type: string
                                    format: uint128
                                  error:
                                    type: string
                                    format: uint128
                variables:
                  type: array
                  items: {}
                maps:
                  type: array
                  items: {}
                fungible_tokens:
                  type: array
                  items: {}
                non_fungible_tokens:
                  type: array
                  items: {} ```

LakshmiLavanyaKasturi: If the above approach is incorrect, I will go ahead with [this](https://github.com/hirosystems/stacks-blockchain-api/pull/1699/commits/76f53d824c1b58f9b533061721365fb984dd36cc) commit.

zone117x: The OpenAPI 3.0 spec on object parameters is here: https://swagger.io/specification/v3/#parameter-object-examples

It looks like the first example I gave may be correct. I don't think you should be trying to define the entire possible ABI object here though, you're going to end up having to specify a large complex enum of clarity types. I seems fine to me if the spec simple states a `string` and just explain the JSON encoded abi in the description.

Issue #1685, opened 2023-07-06
getAddressMempoolTransactions does not support smart contracts


friedger:
**Describe the bug**
transactionApi.getAddressMempoolTransactions fails with error `Invalid query parameter for "SP21YTSM60CAY6D011EZVEVNKXVW8FVZE198XEFFP.pox-fast-pool-v2"` while accountsApi.getAccountTransactions works with contract parameters.

It would be nice to also retrieve mempool transactions filtered by the contract address.

**Additional context**
getHasPendingTransactions fails for FAST pool
https://github.com/hirosystems/lockstacks/blob/b31f6cd53d8c127458a2e1a4be9fd6b147967748/src/pages/stacking/direct-stacking-info/utils-pending-txs.ts#L72

zone117x:
This seems reasonable and relatively simple to implement.

PR #1704, opened 2023-08-07
fix: allow contract-principals in `/extended/v1/address/:principal/mempool` endpoint

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1685

The endpoint `/extended/v1/address/:principal/mempool` (AKA client function `getAddressMempoolTransactions`) was too strict and rejected contract-type principals to be queried.


rafaelcr (approved):
LGTM

PR #1711, opened 2023-09-06
Update rate limit for faucet

LakshmiLavanyaKasturi:
Updated rate-limit for STX faucet.

### DIFF
--- a/content/feature-guides/rate-limiting.md
+++ b/content/feature-guides/rate-limiting.md
@@ -19,4 +19,4 @@ You can refer to the rate limit for each endpoint in the table below:
 
 ## STX faucet
 
-The Stacks faucet rate limits depend on the type of request. For stacking requests, there is a limit of **1 request per 2 days**. In case of regular Stacks faucet requests, the limits are set to **1 request per minute**.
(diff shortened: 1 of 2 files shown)


content/feature-guides/rate-limiting.md
```diff
@@ -19,4 +19,4 @@ You can refer to the rate limit for each endpoint in the table below:
 
 ## STX faucet
 
-The Stacks faucet rate limits depend on the type of request. For stacking requests, there is a limit of **1 request per 2 days**. In case of regular Stacks faucet requests, the limits are set to **1 request per minute**.
+The Stacks faucet rate limits depend on the type of request. For stacking requests, there is a limit of **1 request per 2 days**. In case of regular Stacks faucet requests, the limits are set to **1 request per day**.
```
CharlieC3, line 22: I don't think we actually have a different rate limit depending on whether it's a stacking request or not. It's always 1 request per day.

andresgalante:
We've increased the rate limit of the faucet for the hackathon. My questions are:

- Did we bring it back down again?
- if so, what's the rate limit so I can update this docs and merge it?

CharlieC3:
@andresgalante We did not yet bring the rate limit down. We were going to have @aravindgee handle that change as part of [this PR](https://github.com/hirosystems/k8s/pull/449#issuecomment-1787590984), but we can implement that sooner if you'd like.

zone117x:
Seems like this isn't needed anymore.

PR #1741, opened 2023-10-19
fix: move nft custody view into a table

rafaelcr:
This PR transforms the `nft_custody` and `nft_custody_unanchored` materialized views into tables that are written to with each new block/microblock. This migration increases block ingestion performance considerably.

### DIFF
--- a/migrations/1696872367486_nft-custody-tables.js
+++ b/migrations/1696872367486_nft-custody-tables.js
@@ -1,0 +1,187 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = pgm => {
+  pgm.dropIndex('nft_custody', ['recipient', 'asset_identifier']);
+  pgm.dropIndex('nft_custody', 'value');
+  pgm.renameMaterializedView('nft_custody', 'nft_custody_old');
+  pgm.createTable('nft_custody', {
+    asset_identifier: {
+      type: 'string',
+      notNull: true,
+    },
+    value: {
+      type: 'bytea',
+      notNull: true,
+    },
+    recipient: {
+      type: 'text',
+    },
+    block_height: {
+      type: 'integer',
+      notNull: true,
+    },
+    index_block_hash: {
+      type: 'bytea',
+      notNull: true,
+    },
+    parent_index_block_hash: {
+      type: 'bytea',
+      notNull: true,
+    },
+    microblock_hash: {
+      type: 'bytea',
+      notNull: true,
+    },
+    microblock_sequence: {
+      type: 'integer',
+      notNull: true,
+    },
+    tx_id: {
+      type: 'bytea',
+      notNull: true,
+    },
+    tx_index: {
+      type: 'smallint',
+      notNull: true,
+    },
+    event_index: {
+      type: 'integer',
+      notNull: true,
+    },
+  });
+  pgm.createConstraint('nft_custody', 'nft_custody_unique', 'UNIQUE(asset_identifier, value)');
+  pgm.createIndex('nft_custody', ['recipient', 'asset_identifier']);
+  pgm.createIndex('nft_custody', 'value');
+  pgm.createIndex('nft_custody', [
+    { name: 'block_height', sort: 'DESC' },
+    { name: 'microblock_sequence', sort: 'DESC' },
+    { name: 'tx_index', sort: 'DESC' },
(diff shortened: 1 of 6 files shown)


zone117x:
Very clean 💯 
I'm locally testing out block ingestion times with this

csgui (approved):
Good change! 👍 LGTM!

migrations/1696872367486_nft-custody-tables.js
```diff
@@ -0,0 +1,187 @@
...
+    { name: 'microblock_sequence', sort: 'DESC' },
+    { name: 'tx_index', sort: 'DESC' },
+    { name: 'event_index', sort: 'DESC' }
+  ]);
+  pgm.sql(`
+    INSERT INTO nft_custody (asset_identifier, value, recipient, tx_id, block_height) (
+      SELECT asset_identifier, value, recipient, tx_id, block_height
+      FROM nft_custody_old
+    )
+  `);
```
zone117x, line 68: I'm getting a migration error from this query:
```json
{
  "level": "error",
  "time": "2023-10-25T12:51:37.063Z",
  "pid": 1,
  "hostname": "6750bf061f1a",
  "name": "stacks-blockchain-api",
  "component": "core-api",
  "err": {
    "type": "DatabaseError",
    "message": "null value in column \"index_block_hash\" of relation \"nft_custody\" violates not-null constraint",
    "stack": "error: null value in column \"index_block_hash\" of relation \"nft_custody\" violates not-null constraint\n    at Parser.parseErrorMessage (/app/node_modules/pg-protocol/src/parser.ts:369:69)\n    at Parser.handlePacket (/app/node_modules/pg-protocol/src/parser.ts:188:21)\n    at Parser.parse (/app/node_modules/pg-protocol/src/parser.ts:103:30)\n    at Socket.<anonymous> (/app/node_modules/pg-protocol/src/index.ts:7:48)\n    at Socket.emit (node:events:513:28)\n    at addChunk (node:internal/streams/readable:315:12)\n    at readableAddChunk (node:internal/streams/readable:289:9)\n    at Socket.Readable.push (node:internal/streams/readable:228:10)\n    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)",
    "length": 500,
    "name": "error",
    "severity": "ERROR",
    "code": "23502",
    "detail": "Failing row contains (SP000000000000000000002Q6VF78.bns::names, \\x0c00000002046e616d6502000000012d096e616d6573706163650200000002..., SPJ5BG9TBW0ZE36ZKK8W59PS1JZT2HSYESR90BKD, 94746, null, null, null, null, \\x445b1dde3854841d4c13a621bfd759ad5fa43ec99c178ebc3937e2d234f205..., null, null).",
    "schema": "stacks_blockchain_api",
    "table": "nft_custody",
    "column": "index_block_hash",
    "file": "execMain.c",
    "line": "1883",
    "routine": "ExecConstraints"
  },
  "msg": "Error running pg-migrate"
}
```

My local deployment is based off a Hiro Archive pg dump (`stacks-blockchain-api-pg-15-7.3.0-20230926.dump`). If `index_block_hash` should never be null, then there may be an issue with the archives (perhaps bns import wasn't ran?). Otherwise, could be an issue with this migration query?

zone117x: Ah I see the issue. The original `nft_custody` materialized view doesn't have the columns required for this INSERT query. In particular: `index_block_hash, parent_index_block_hash, microblock_hash, microblock_sequence, tx_index, event_index`.

I think this INSERT query would need to join against the `nft_events` and/or `txs` table to retrieve these values.

zone117x: Should be fixed with this commit: https://github.com/hirosystems/stacks-blockchain-api/pull/1741/commits/8a69698a042de63378bb060b807d2f2e72ada8db

@rafaelcr can you check that this looks correct?

zone117x (approved):
The migration takes about 30 seconds on my machine, which I think might be a few minutes on the production pg server. Seems okay to me.

I also verified that the block ingestion times are consistently 1-3 seconds with this change.

migrations/1696872367486_nft-custody-tables.js
zone117x: Latest commit works. Migration only takes around 30 seconds on my machine, not bad at all.

PR #1744, opened 2023-10-26
update link references to api endpoints

ryanwaits:
related issue: https://github.com/hirosystems/docs/issues/183

### DIFF
--- a/content/how-to-guides/how-to-handle-errors.md
+++ b/content/how-to-guides/how-to-handle-errors.md
@@ -17,13 +17,13 @@ The Stacks 2.0 Blockchain API is centrally hosted. However, every running Stacks
 
 While the Node RPC API doe not provide the same functionality as the hosted Stacks 2.0 Blockchain API, you will have similar functionality in a way that is scoped to that specific node. The RPC API includes the following endpoints:
 
-- [POST /v2/transactions](https://docs.hiro.so/api#operation/post_core_node_transactions)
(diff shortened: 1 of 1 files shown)


content/how-to-guides/how-to-handle-errors.md
```diff
@@ -17,13 +17,13 @@ The Stacks 2.0 Blockchain API is centrally hosted. However, every running Stacks
 
 While the Node RPC API doe not provide the same functionality as the hosted Stacks 2.0 Blockchain API, you will have similar functionality in a way that is scoped to that specific node. The RPC API includes the following endpoints:
```
sabbyanandan, line 18:
```suggestion
While the Stacks Node RPC API does not provide the same functionality as the hosted Stacks Blockchain API, you will have similar functionality in a way that is scoped to that specific node. The RPC API includes the following endpoints:
```

sabbyanandan: Aside from the nit and a typo, I do not follow this statement. Suppose the goal is to say Hiro's Stacks Blockchain API supports the following endpoints that act as a proxy to fetch the raw Stacks RPC endpoint responses. In that case, the latter part of the statement (i.e., "The RPC API includes the following endpoints:") is confusing. Perhaps we could rephrase.

sabbyanandan (changes requested):

content/how-to-guides/how-to-handle-errors.md
ryanwaits: it prob makes sense just to remove that last line, will go ahead and do that

sabbyanandan (approved):

PR #1747, opened 2023-11-08
refactor: remove old token metadata processor

rafaelcr:
Removes the old FT/NFT token metadata processor. We already have the Token Metadata API live and all current traffic is being redirected there.

This change will optimize block ingestion times because the API will no longer need to analyze the ABI of all deployed contracts to determine if they deploy FTs or NFTs. Also, it will no longer need to maintain token queues and notify listeners when the queues are updated.

---

### Rosetta

This change impacts Rosetta, because FT operations still need to be displayed in blocks and transactions with the proper `symbol` and `decimals`. To solve this, a FT metadata retrieval class is added that just queries the relevant contracts for this info before operations are displayed, keeping it in an in-memory LRU cache for quicker access in the future.


src/rosetta/rosetta-ft-metadata-client.ts
```diff
@@ -0,0 +1,187 @@
...
+    default:
+      return RosettaTokenMetadataErrorMode.warning;
+  }
+}
+
+/**
+ * LRU cache that keeps `RosettaFtMetadata` entries for FTs used in the Stacks chain and retrieved
+ * by the Rosetta endpoints.
+ */
+const ftMetadataCache = new LRUCache<string, RosettaFtMetadata>({ max: 5_000 });
```
zone117x, line 53: I don't think the caller functions do any de-duplication when looking up metadata for a set of asset IDs in a given block/tx, right? For example, if a block/tx contained say 100 transfers associated with a single asset, then this client would likely issue 100 concurrent lookups to the same RPC endpoints.

To handle that case, we probably want to change this to cache a `Promise<RosettaFtMetdata>`. That way the first call to `getFtMetdata` would immediate cache the pending lookup promise, and the additional calls would all await that first one. 

What do you think? I can open a PR against this one with that change.

zone117x: The change was pretty minor so I just pushed it in https://github.com/hirosystems/stacks-blockchain-api/pull/1747/commits/ac4282ef5112156c48053b060ef3fa227f20d5d8

Let me know what you think

rafaelcr: Looks great! Thanks for the change

migrations/1621511823100_token_metadata_queue.js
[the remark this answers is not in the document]
rafaelcr: Good idea, I'll do that

src/api/routes/rosetta/account.ts
```diff
@@ -122,22 +121,21 @@ export function createRosettaAccountRouter(db: PgStore, chainId: ChainID): expre
           ];
 
           // Add Fungible Token balances.
-          if (isFtMetadataEnabled()) {
```
zone117x, line 125: Is there still a way for Rosetta users to disable ft metadata? I think at least 1 entity isn't using it, and idk if they might run into problems if it's forced enabled for them.

rafaelcr: Hmm yeah good point, I'll re-add this ENV

zone117x (approved):
LGTM!

PR #1751, opened 2023-11-16
fix: convert `chain_tip` materialized view into a table

rafaelcr:
This PR removes the `chain_tip` materialized view, because it relied on `COUNT(*)` operations that are considerably slow. Instead it updates the chain tip with totals as blocks are ingested.


zone117x (approved):
Nice job on the reorg handling 💯

PR #1760, opened 2023-12-07
fix: insert block transaction data in batches

rafaelcr:


src/datastore/pg-write-store.ts
```diff
@@ -418,63 +301,128 @@ export class PgWriteStore extends PgStore {
...
+      const missingTxs = data.txs.filter(entry =>
+        missingMicroblockHashes.has(entry.tx.microblock_hash)
+      );
+      await this.insertMicroblockData(sql, missingMicroblocks, missingTxs);
+      return missingMicroblockHashes;
+    }
+    return new Set();
+  }
+
+  private async updatePoxStateUnlockHeight(sql: PgSqlClient, data: DataStoreBlockUpdateData) {
```
zone117x, line 387: Do you want to also move the `pox_v3_unlock_height` update into this function?

zone117x (approved):
Awesome optimizations! 👏

PR #1762, opened 2023-12-09
PoX-4 tests

zone117x:
* Update the stacks-node docker image to latest Nakamato code which boots into epoch 2.5 with `pox-4` enabled: https://github.com/hirosystems/stacks-regtest-env/pull/26
* Update all pox-3 tests to now test pox-4


src/tests-2.5/pox-4-delegate-revoked-stacking.ts
```diff
@@ -340,14 +351,15 @@ describe('PoX-3 - Delegate Revoked Stacking', () => {
...
+      fee: 10000n,
     });
     const delegateTxResult = await testEnv.client.sendTransaction(
       Buffer.from(delegateTx.serialize())
     );
     const delegateDbTx = await standByForTx(delegateTxResult.txId);
-    expect(delegateDbTx.status).not.toBe(DbTxStatus.Success);
     const delegateResult = decodeClarityValue(delegateDbTx.raw_result);
     expect(delegateResult.repr).toEqual('(err 3)'); // ERR_STACKING_ALREADY_STACKED
+    expect(delegateDbTx.status).not.toBe(DbTxStatus.Success);
```
zone117x, line 362: @janniks this test is showing that revoke delegation behavior in pox-4 is different than pox-3. Does this seem like a pox-4 regression to you?

zone117x: It seems to be only this individual test `Try to perform delegate-stx - again`, if skipped then all the rest of the revoke tests pass.

janniks: If I understand correctly, this is intended in pox-4. `Try to perform delegate-stx - again` tests that if still locked a `delegate-stx` will fail.

In pox-4 we can delegate-stx, even if stacking:
https://github.com/stacks-network/stacks-core/blob/de8129cced156bb180bd19e42ac4ad1b2bca8d53/stackslib/src/chainstate/stacks/boot/pox-4.clar#L660-L661

zone117x: Good catch! I removed that test. All should be passing now

rafaelcr (approved):
LGTM

src/tests-2.5/pox-4-delegate-revoked-stacking.ts
```diff
@@ -271,9 +276,14 @@ describe('PoX-3 - Delegate Revoked Stacking', () => {
...
       network: testEnv.stacksNetwork,
       anchorMode: AnchorMode.OnChainOnly,
+      fee: 10000n,
     });
     const revokeTxResult = await testEnv.client.sendTransaction(Buffer.from(revokeTx.serialize()));
-    await standByForTxSuccess(revokeTxResult.txId);
+    const revokeStackDbTx = await standByForTxSuccess(revokeTxResult.txId);
+
+    const revokeStackResult = decodeClarityValue(revokeStackDbTx.raw_result);
+    expect(revokeStackResult.repr).toEqual('(ok true)'); // ERR_STACKING_PERMISSION_DENIED
```
janniks, line 285: nit: misleading comment?

janniks (approved):
👏4️⃣

PR #1765, opened 2023-12-11
fix: use the mempool etag for the /nonces endpoint

rafaelcr:


zone117x (approved):
Thanks for the quick fix 💯

PR #1769, opened 2023-12-13
feat: add `/extended/v2/blocks` endpoint with burn block filters

### LINKED ISSUE #1755: Get STX blocks by BTC block
Hello team,
We are currently working on a new feature in the explorer to display a list of STX blocks associated with a specific BTC block. For this, we need a new endpoint that accepts either the BTC block hash or height and returns a paginated list of STX blocks.
Suggested interface:
GET `/blocks/by_burn_block_hash/{burn_block_hash}?limit={limit}&offset={offset}`
Response:
```
{
    limit: number;
    offset: number;
    total: number;
    results: Array<Block>;
}
```

rafaelcr:
* Validates request query schema with TypeBox
* Creates `/v2` router

Fixes #1755


src/tests/block-tests.ts
```diff
@@ -630,4 +630,111 @@ describe('block tests', () => {
     expect(blockQuery.body.execution_cost_write_count).toBe(3);
     expect(blockQuery.body.execution_cost_write_length).toBe(3);
   });
+
+  test('blocks v2 filtered by burn block', async () => {
```
zone117x, line 634: Should there be a test for querying a burn block that has multiple stacks blocks?

src/api/routes/v2/schemas.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const BlockLimitParam = Type.Integer({
+  minimum: 1,
+  maximum: pagingQueryLimits[ResourceType.Block].maxLimit,
+  default: pagingQueryLimits[ResourceType.Block].defaultLimit,
+  title: 'Block limit',
+  description: 'Blocks per page',
+});
+
+const BurnBlockHashParam = Type.RegExp(/^[0]{8}[a-fA-F0-9]{56}$/, {
```
zone117x, line 23: The `0` prefix check here should be removed. The number of leading zeros in a bitcoin block hash is dependent on network difficulty, which changes overtime on mainnet, and is completely different on testnet and regtest.

zone117x, line 23: Should this also allow for a `0x` prefix? It's not common for bitcoin block hashes, but we use it for the rest of the hash strings, i.e. `/^(0x)?[a-fA-F0-9]{64}$/i`

src/tests/block-tests.ts
rafaelcr: This test creates a few Stacks blocks per Bitcoin block and then filters by burn block hash to check for result count, etc. Is that what you mean? Or you mean calling the `/burn_block` endpoint also in this test?

src/api/routes/v2/schemas.ts
rafaelcr: Done

rafaelcr: Done

zone117x (approved):
💯

PR #1774, opened 2023-12-15
feat: add `/extended/v2/blocks/:height_or_hash`

rafaelcr:
Support filtering by block height, block hash, index block hash or `latest`


zone117x (approved):
It kinda seems like the `/:index_block_hash` response should be a 3xx redirect to the `/:block_hash` endpoint, but I don't feel strongly.

LGTM!

PR #1776, opened 2023-12-17
fix: do not load duckdb binary unless required

zone117x:
Fixes an error when starting the API (without event-replay) where duckdb has not been installed. The dependency is now lazy loaded on demand.

```
Error: /stacks-api/node_modules/duckdb/lib/binding/duckdb.node: cannot open shared object file: No such file or directory
```


zone117x:
The `unused-exports` lint seems buggy, it's probably not catching the dynamic `import` of the a module. Seems like we should remove that check and replace it with something from eslint (if available). Otherwise, I don't think we get much value of that anyway.

csgui (approved):
LGTM!

PR #1778, opened 2023-12-18
feat: add `tx_count` property to `/extended/v2/blocks`

rafaelcr:
* Add `tx_count` column to `blocks` table for faster tx count reads
* Remove `txs` list since it was taking a lot of response space
* Add `/extended/v2/blocks/:hash_or_height/transactions` to get all transactions for a block
* Switch to Ajv type compiler/validation for easier type coercion


src/tests/tx-tests.ts
```diff
@@ -3283,6 +3298,101 @@ describe('tx tests', () => {
...
+          canonical: true,
+          microblock_canonical: true,
+          microblock_sequence: I32_MAX,
+          microblock_hash: '',
+          fee_rate: 1234n,
+          sender_address: 'sender-addr',
+        })
+        .build()
+    );
+    const result = await supertest(api.server).get(
```
zone117x, line 3347: A test for http 400 response would be useful too

zone117x (approved):
LGTM!

PR #1781, opened 2023-12-21
fix: optimize mempool transaction reads and writes

rafaelcr:
* Delete `mempool_digest` materialized view, keep data in `chain_tip` table
* Change the digest algorithm from a hash to a simple "updated at" timestamp so mempool writes are extremely cheap
* Keep `tx_count` column to avoid running ad-hoc counts over unpruned txs
* Insert mempool txs in batches

### DIFF
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -713,20 +713,28 @@ export class PgWriteStore extends PgStore {
   // NOTE: this is essentially a work-around for whatever bug is causing the underlying problem.
   async reconcileMempoolStatus(sql: PgSqlClient): Promise<void> {
     const txsResult = await sql<{ tx_id: string }[]>`
-      UPDATE mempool_txs
-      SET pruned = true
-      FROM txs
-      WHERE
-        mempool_txs.tx_id = txs.tx_id AND
-        mempool_txs.pruned = false AND
-        txs.canonical = true AND
-        txs.microblock_canonical = true AND
-        txs.status IN ${sql([
-          DbTxStatus.Success,
-          DbTxStatus.AbortByResponse,
-          DbTxStatus.AbortByPostCondition,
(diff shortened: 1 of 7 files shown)


src/datastore/pg-write-store.ts
```diff
@@ -2747,28 +2761,6 @@ export class PgWriteStore extends PgStore {
...
-   * Called when a full event import is complete.
-   */
-  async finishEventReplay() {
-    if (!this.isEventReplay) return;
-    await this.refreshMaterializedView('mempool_digest', this.sql, false);
-  }
-
   /**
    * batch operations (mainly for event-replay)
    */
```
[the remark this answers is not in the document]
rafaelcr: Great catch, done!

src/tests/mempool-tests.ts
```diff
@@ -1546,7 +1546,7 @@ describe('mempool tests', () => {
     // directly inserting the mempool-tx and mined-tx, bypassing the normal update functions.
     await db.updateBlock(db.sql, dbBlock1);
     const chainTip = await db.getChainTip();
-    await db.insertDbMempoolTx(mempoolTx, chainTip, db.sql);
+    await db.insertDbMempoolTxs([mempoolTx], chainTip, db.sql);
     await db.updateTx(db.sql, dbTx1);
 
     // Verify tx shows up in mempool (non-pruned)
```
zone117x, line 1552: Would it be reasonable to check for the new behavior in the `Reconcile mempool pruned status` test? Not sure if that makes sense, up to you

zone117x (approved):
Left a comment about adding another test, otherwise LGTM!

src/tests/mempool-tests.ts
rafaelcr: Good idea, done

PR #1807, opened 2024-01-09
fix(docs): URL query arrays should be formatted with `form` rather than comma-separated

zone117x:
Addresses issue https://github.com/hirosystems/stacks-blockchain-api/issues/1806

Some OpenAPI tooling default to comma-separated strings for arrays in URL query params. In particular, the `docusaurus` generator used in Hiro docs.

This specifies the default behavior used by Express which is multiple parameters with the same name. E.g.

Before:
   `?tx_id=1111,2222`
After:
   `?tx_id=1111&tx_id=2222`

PR #1809, opened 2024-01-09
fix: support comma-separated strings in array query params

zone117x:
Related to https://github.com/hirosystems/stacks-blockchain-api/pull/1807

This adds support for parsing comma-separated array strings from url query params. Now both common formats for specifying arrays should work:
* `?tx_id=ID1,ID2,ID3`
* `?tx_id=ID1&tx_id=ID2&tx_id=ID3`


rafaelcr (approved):
Nice!

PR #1810, opened 2024-01-09
feat: add `order_by` and `order` params to `/extended/v1/tx/mempool`

rafaelcr:
`Age`, `size`, and `fee` for sorting options with `asc` and `desc` for order


src/api/routes/tx.ts
```diff
@@ -152,10 +152,28 @@ export function createTxRouter(db: PgStore): express.Router {
           InvalidRequestErrorType.invalid_param
         );
       }
+
+      const orderBy = req.query.order_by;
+      if (orderBy !== undefined && orderBy != 'fee' && orderBy != 'age' && orderBy != 'size') {
```
zone117x, line 157: I think we typically put constants like these into an enum or string union type (either a manually defined type, or one generated in the .d.ts from the openapi spec)

rafaelcr: Done!

zone117x (approved):

PR #1815, opened 2024-01-10
feat: remove reconcile mempool & debounce stats

bestmike007:
## Description

Currently when the event server handles a new_mempool_tx event, it reconcile mempool status and recalculate mempool stats. This is really unnecessary, especially when an influx of new transactions occurs. As the stacks node awaits the event server processing, it slows down the new block processing.

This PR suggests a method to reconcile and update mempool statistics through a debounce scheme. It significantly decreases the processing time for new_mempool_tx events from over 1s to less than 10ms. 

## Type of Change

Performance optimization.

No.

No.

## Testing information

Since the reconcile process is deferred, a method to wait for next reconcile is added for testing purposes. And the tests are updated accordingly.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -1678,51 +1678,38 @@ export class PgWriteStore extends PgStore {
         tenure_change_signature: tx.tenure_change_signature ?? null,
         tenure_change_signers: tx.tenure_change_signers ?? null,
       }));
+      // The incoming mempool transactions might have already been settled
+      // We need to mark them as pruned to avoid inconsistent tx state
       const result = await sql<{ tx_id: string }[]>`
         WITH inserted AS (
           INSERT INTO mempool_txs ${sql(values)}
           ON CONFLICT ON CONSTRAINT unique_tx_id DO NOTHING
           RETURNING tx_id
         ),
+        settled AS (
+          SELECT tx_id
+          FROM txs
+          WHERE
+            tx_id IN ${sql(values.map(b => b.tx_id))} AND
+            canonical = true AND
+            microblock_canonical = true
+        ),
+        pruned AS (
+          UPDATE mempool_txs
+          SET pruned = true
+          WHERE
+            tx_id IN (SELECT tx_id FROM settled) AND
+            pruned = false
+          RETURNING tx_id
+        ),
         count_update AS (
           UPDATE chain_tip SET
-            mempool_tx_count = mempool_tx_count + (SELECT COUNT(*) FROM inserted),
+            mempool_tx_count = mempool_tx_count + (SELECT COUNT(*) FROM inserted) - (SELECT COUNT(*) FROM pruned),
             mempool_updated_at = NOW()
         )
         SELECT tx_id FROM inserted
       `;
       txIds.push(...result.map(r => r.tx_id));
-      // The incoming mempool tx might have already been settled
-      // We need to mark them as pruned to avoid inconsistent tx state
-      const pruned_tx = await sql<{ tx_id: string }[]>`
-        SELECT tx_id
-        FROM txs
-        WHERE
-          tx_id IN ${sql(batch.map(b => b.tx_id))} AND
-          canonical = true AND
-          microblock_canonical = true AND
-          status IN ${sql([
-            DbTxStatus.Success,
-            DbTxStatus.AbortByResponse,
-            DbTxStatus.AbortByPostCondition,
-          ])}`;
-      if (pruned_tx.length > 0) {
-        await sql<{ tx_id: string }[]>`
-          WITH pruned AS (
-            UPDATE mempool_txs
-            SET pruned = true
-            WHERE
-              tx_id IN ${sql(pruned_tx.map(t => t.tx_id))} AND
-              pruned = false
-            RETURNING tx_id
-          ),
-          count_update AS (
-            UPDATE chain_tip SET
-              mempool_tx_count = mempool_tx_count - (SELECT COUNT(*) FROM pruned),
-              mempool_updated_at = NOW()
-          )
-          SELECT tx_id FROM pruned`;
-      }
     }
     return txIds;
   }
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -1686,38 +1686,46 @@ export class PgWriteStore extends PgStore {
         tenure_change_signature: tx.tenure_change_signature ?? null,
         tenure_change_signers: tx.tenure_change_signers ?? null,
       }));
-      // The incoming mempool transactions might have already been settled
-      // We need to mark them as pruned to avoid inconsistent tx state
       const result = await sql<{ tx_id: string }[]>`
         WITH inserted AS (
           INSERT INTO mempool_txs ${sql(values)}
           ON CONFLICT ON CONSTRAINT unique_tx_id DO NOTHING
           RETURNING tx_id
         ),
-        settled AS (
-          SELECT tx_id
-          FROM txs
-          WHERE
-            tx_id IN ${sql(values.map(b => b.tx_id))} AND
-            canonical = true AND
-            microblock_canonical = true
-        ),
-        pruned AS (
-          UPDATE mempool_txs
-          SET pruned = true
-          WHERE
-            tx_id IN (SELECT tx_id FROM settled) AND
-            pruned = false
-          RETURNING tx_id
-        ),
         count_update AS (
           UPDATE chain_tip SET
-            mempool_tx_count = mempool_tx_count + (SELECT COUNT(*) FROM inserted) - (SELECT COUNT(*) FROM pruned),
+            mempool_tx_count = mempool_tx_count + (SELECT COUNT(*) FROM inserted),
             mempool_updated_at = NOW()
         )
         SELECT tx_id FROM inserted
       `;
       txIds.push(...result.map(r => r.tx_id));
+      // The incoming mempool transactions might have already been settled
+      // We need to mark them as pruned to avoid inconsistent tx state
+      const pruned_tx = await sql<{ tx_id: string }[]>`
+        SELECT tx_id
+        FROM txs
+        WHERE
+          tx_id IN ${sql(batch.map(b => b.tx_id))} AND
+          canonical = true AND
+          microblock_canonical = true`;
+      if (pruned_tx.length > 0) {
+        await sql<{ tx_id: string }[]>`
+          WITH pruned AS (
+            UPDATE mempool_txs
+            SET pruned = true
+            WHERE
+              tx_id IN ${sql(pruned_tx.map(t => t.tx_id))} AND
+              pruned = false
+            RETURNING tx_id
+          ),
+          count_update AS (
+            UPDATE chain_tip SET
+              mempool_tx_count = mempool_tx_count - (SELECT COUNT(*) FROM pruned),
+              mempool_updated_at = NOW()
+          )
+          SELECT tx_id FROM pruned`;
+      }
     }
     return txIds;
   }
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -1698,7 +1698,13 @@ export class PgWriteStore extends PgStore {
     lastTrigger?: number | null;
     debounce?: NodeJS.Timeout | null;
     running: boolean;
-  } = { running: false };
+    listeners: (() => unknown)[];
+  } = { running: false, listeners: [] };
+  public waitForNextReconcileMempool() {
+    return new Promise<void>(f => {
+      this._debounceReconcileMempool.listeners.push(f);
+    });
+  }
   private debounceReconcileMempool() {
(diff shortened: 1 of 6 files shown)


bestmike007:
@rafaelcr @zone117x please help me review this. I'm not sure how to update the changelog.

zone117x (commented):
Thanks for the PR! This is an interesting approach, but it looks like it will cause data consistency issues. For example, here is the comment on the `reconcileMempoolStatus` function:
```
  // Find any transactions that are erroneously still marked as both `pending` in the mempool table
  // and also confirmed in the mined txs table. Mark these as pruned in the mempool and log warning.
  // This must be called _after_ any writes to txs/mempool tables during block and microblock ingestion,
  // but _before_ any reads or view refreshes that depend on the mempool table.
  // NOTE: this is essentially a work-around for whatever bug is causing the underlying problem.
```

This PR moves the mempool tx status updating to outside of the sql transaction for /new_block ingestion. Which means clients will have period of time where a tx can appear to exist as both mined and still in the mempool.

As the above comment mentions, this reconcile function is a work-around for a bug that causes this inconsistent state. If we are to make changes around this, it would be better to track down and fix that bug, then we can remove the `reconcileMempoolStatus` function all together.

bestmike007:
I see, that will require more efforts. Does the mempool stats need to be consistent as well?

Is the bug cause by: the new mempool transaction event occurs after, not before, the new block event which settles the corresponding transaction?

bestmike007:
This avoid inconsistent states caused by new inserts.

> Is the bug cause by: the new mempool transaction event occurs after, not before, the new block event which settles the corresponding transaction?

This is possible if the stacks-node does not check tx state before emitting new_mempool_transaction event. Theoretically, it SHOULD broadcast already settled transactions: they might be reorged and would be absent from the mempool table.

src/datastore/pg-write-store.ts
```diff
@@ -1696,21 +1692,82 @@ export class PgWriteStore extends PgStore {
         SELECT tx_id FROM inserted
       `;
       txIds.push(...result.map(r => r.tx_id));
+      // The incoming mempool tx might have already been settled
+      // We need to mark them as pruned to avoid inconsistent tx state
+      const pruned_tx = await sql<{ tx_id: string }[]>`
```
zone117x, line 1697: Instead of multiple sql queries after the inserting into `mempool_txs`, can you instead modify the above query so that if a tx is already settled, then it should be inserted with `pruned=true`?

bestmike007:
Doesn't look like the received transactions are filtered before forwarding to event observer:

https://github.com/stacks-network/stacks-core/blob/2.4.0.0.4/testnet/stacks-node/src/neon_node.rs#L2267-L2294
https://github.com/stacks-network/stacks-core/blob/2.4.0.0.4/src/net/relay.rs#L1815-L1857
https://github.com/stacks-network/stacks-core/blob/2.4.0.0.4/src/net/relay.rs#L1567-L1596

src/datastore/pg-write-store.ts
bestmike007: The added query is run using the index: `Index Scan using txs_tx_id_index on txs  (cost=0.00..12.04 rows=1 width=33)`; so mempool transactions event should be processed within 20ms.

zone117x: Is there a reason it can't be done in the first query with a CTE or similar?

bestmike007: Not sure how to write the CTE query to insert `${sql(values)}` and override the pruned state, and in the meantime update the tx already exists with correct pruned state, and also get the correct number of new mempool tx count. Looks like it will be more complex and maybe less performant.

Can you give me an example so I can follow?

zone117x: It's likely more performant to perform the updates in one query as it gives the postgres query planner more room to optimize. I can take a look at helping with the PR later unless we can determine that this is high priority right now

bestmike007: Is this what you're suggesting?

```typescript
const result = await sql<{ tx_id: string }[]>`
  WITH inserted AS (
    INSERT INTO mempool_txs ${sql(values)}
    ON CONFLICT ON CONSTRAINT unique_tx_id DO NOTHING
    RETURNING tx_id
  ),
  settled AS (
    SELECT tx_id
    FROM txs
    WHERE
      tx_id IN ${sql(values.map(b => b.tx_id))} AND
      canonical = true AND
      microblock_canonical = true AND
      status IN ${sql([
        DbTxStatus.Success,
        DbTxStatus.AbortByResponse,
        DbTxStatus.AbortByPostCondition,
      ])}
  ),
  pruned AS (
    UPDATE mempool_txs
    SET pruned = true
    WHERE
      tx_id IN (SELECT tx_id FROM settled) AND
      pruned = false
    RETURNING tx_id
  ),
  count_update AS (
    UPDATE chain_tip SET
      mempool_tx_count = mempool_tx_count + (SELECT COUNT(*) FROM inserted) - (SELECT COUNT(*) FROM pruned),
      mempool_updated_at = NOW()
  )
  SELECT tx_id FROM inserted
`;
```

zone117x: That looks about right. Is the `status IN ...` necessary? And would it be possible to avoid passing in tx_id twice? E.g.
```sql
...
settled AS (
  SELECT tx_id
  FROM txs
  WHERE
    tx_id IN (SELECT tx_id FROM inserted) AND
    canonical = true AND
    microblock_canonical = true
)
...
```

bestmike007: > Is the status IN ... necessary?

I don't think so, but it's in the `reconcileMempoolStatus` function. I thought there were other possibilities.

The `inserted` is a subset of the original transactions, because duplicated tx_id will be excluded.

bestmike007: The cost is way higher than checking current inserted transactions.

image

src/datastore/pg-write-store.ts
```diff
@@ -1700,17 +1714,47 @@ export class PgWriteStore extends PgStore {
...
+        this.eventEmitter.emit('mempoolStatsUpdate', mempoolStats);
+      } catch (e) {
+        logger.error(e, `failed to reconcile mempool`);
+      } finally {
+        this._debounceMempoolStat.running = false;
+        if (this._debounceMempoolStat.lastTrigger != null) {
+          this.debounceMempoolStat();
+        }
+      }
+    }, 1000);
```
zone117x, line 1746: This can default to 1000, but can you make it a configurable env var?

src/datastore/pg-write-store.ts
bestmike007: Pushed the changes.

src/datastore/pg-write-store.ts
bestmike007: Done.

zone117x: Please also add associated entry to `.env` file

bestmike007: Added, please review the wording.

src/datastore/pg-write-store.ts
```diff
@@ -1700,17 +1717,47 @@ export class PgWriteStore extends PgStore {
...
+        this.eventEmitter.emit('mempoolStatsUpdate', mempoolStats);
+      } catch (e) {
+        logger.error(e, `failed to reconcile mempool`);
+      } finally {
+        this._debounceMempoolStat.running = false;
+        if (this._debounceMempoolStat.lastTrigger != null) {
+          this.debounceMempoolStat();
+        }
+      }
+    }, MEMPOOL_STATS_DEBOUNCE_INTERVAL);
```
zone117x, line 1749: Should we be concerned about a possible very long delay in the case of receiving a new mempool tx at least once per MEMPOOL_STATS_DEBOUNCE_INTERVAL for a prolonged period of time? Could be solvable by having another timeout for "max stats wait period" where it runs the stats if too much time has passed?

src/datastore/pg-write-store.ts
```diff
@@ -1700,17 +1717,47 @@ export class PgWriteStore extends PgStore {
...
+  }
+
   async updateMempoolTxs({ mempoolTxs: txs }: { mempoolTxs: DbMempoolTxRaw[] }): Promise<void> {
     const updatedTxIds: string[] = [];
     await this.sqlWriteTransaction(async sql => {
       const chainTip = await this.getChainTip();
       updatedTxIds.push(...(await this.insertDbMempoolTxs(txs, chainTip, sql)));
-      if (!this.isEventReplay) {
-        await this.reconcileMempoolStatus(sql);
```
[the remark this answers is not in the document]
bestmike007: We still need it to update mempool tx pruned status after new block processed, no?

src/datastore/pg-write-store.ts
zone117x: LGTM thanks!

src/datastore/pg-write-store.ts
zone117x: LGTM, @rafaelcr can you also review this query?

src/datastore/pg-write-store.ts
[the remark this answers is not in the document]
zone117x: That should be done by the calls to `pruneMempoolTxs` (assuming there isn't a bug elsewhere causing inconsistent state)

src/datastore/pg-write-store.ts
```diff
@@ -1682,15 +1681,33 @@ export class PgWriteStore extends PgStore {
...
+        settled AS (
+          SELECT tx_id
+          FROM txs
+          WHERE
+            tx_id IN ${sql(values.map(b => b.tx_id))} AND
+            canonical = true AND
+            microblock_canonical = true
+        ),
+        pruned AS (
+          UPDATE mempool_txs
```
rafaelcr, line 1701: This query is looking great! However, can you add a test for this scenario @bestmike007 ? I'm worried that a row that is both inserted and updated within the same query will trigger a "cannot affect row twice" error in Postgres.

If that is the case you'll be forced to split the query in two stages. A test will help us determine that.

src/datastore/pg-write-store.ts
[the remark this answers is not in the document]
rafaelcr: There is already a function that marks inserted transactions as pruned in the mempool in another step of the block insert

src/datastore/pg-write-store.ts
bestmike007: Added the logic.

src/helpers.ts
```diff
@@ -764,3 +764,13 @@ export enum BootContractAddress {
   mainnet = 'SP000000000000000000002Q6VF78',
   testnet = 'ST000000000000000000002AMW42H',
 }
+
+export function getUintEnvOrDefault(envName: string, defaultValue = 0) {
+  const v = BigInt(process.env['MEMPOOL_STATS_DEBOUNCE_INTERVAL'] ?? defaultValue);
+  if (v < 0n) {
+    throw new Error(
+      `Expecting ENV ${envName} to be non-negative number but it is configured as ${process.env['MEMPOOL_STATS_DEBOUNCE_INTERVAL']}`
```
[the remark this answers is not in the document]
bestmike007: My bad.

src/datastore/pg-write-store.ts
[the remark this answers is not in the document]
bestmike007: Removing the `reconcileMempoolStatus` fails the test: https://github.com/bestmike007/stacks-blockchain-api/actions/runs/7476647698/job/20347512997#step:10:2173

Am I missing something?

src/datastore/pg-write-store.ts
bestmike007: It's worse, it succeeds with unexpected behavior. Let's revert it to using 2 stages.

image

src/helpers.ts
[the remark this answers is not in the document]
bestmike007: Fixed.

src/datastore/pg-write-store.ts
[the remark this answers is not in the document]
bestmike007: Let's leave it to a subsequent PR, shall we?

src/datastore/pg-write-store.ts
bestmike007: Reverted.

src/datastore/pg-write-store.ts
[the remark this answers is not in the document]
bestmike007: Removed `reconcileMempoolStatus`.

Is this one flaky? The first run failed.

https://github.com/bestmike007/stacks-blockchain-api/actions/runs/7477045881/job/20348840926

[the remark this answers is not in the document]
zone117x: Yeah that one can be flaky

zone117x (approved):

rafaelcr (approved):
LGTM, thanks for the contribution @bestmike007 💯

PR #1818, opened 2024-01-10
feat: run inserts in batch and in parallel when processing new block

bestmike007:
## Description

This PR improves the new block process improvements by means of running inserts in batch and in parallel.

Previously when the event server process a new block, it roughly process less than 10 transactions per second.
When tested with concurrency=100, it processes more than 30 transactions per second.

## Type of Change

Performance improvements.

No.

No.

## Testing information

No changes required.

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/.env
+++ b/.env
@@ -43,6 +43,10 @@ PG_APPLICATION_NAME=stacks-blockchain-api
 # Limit to how many concurrent connections can be created, defaults to 10
 # PG_CONNECTION_POOL_MAX=10
 
+# Insert concurrency when processing new blocks
+# It'd better set it higher, e.g. 8 or 16, if your postgres is running on SSD and multiple CPU cores
+# INSERT_CONCURRENCY=4
+
 # If specified, controls the Stacks Blockchain API mode. The possible values are:
 # * `readonly`: Runs the API endpoints without an Event Server that listens to events from a node and
 #       writes them to the local database. The API will only read data from the PG database
--- a/.env
+++ b/.env
@@ -44,7 +44,7 @@ PG_APPLICATION_NAME=stacks-blockchain-api
 # PG_CONNECTION_POOL_MAX=10
 
 # Insert concurrency when processing new blocks
-# It'd better set it higher, e.g. 8 or 16, if your postgres is running on SSD and multiple CPU cores
+# If your PostgreSQL is operating on SSD and has multiple CPU cores, consider raising this value, for instance, to 8 or 16.
 # STACKS_BLOCK_DATA_INSERT_CONCURRENCY=4
 
 # If specified, controls the Stacks Blockchain API mode. The possible values are:
--- a/.env
+++ b/.env
@@ -45,7 +45,7 @@ PG_APPLICATION_NAME=stacks-blockchain-api
 
 # Insert concurrency when processing new blocks
 # It'd better set it higher, e.g. 8 or 16, if your postgres is running on SSD and multiple CPU cores
-# INSERT_CONCURRENCY=4
+# STACKS_BLOCK_DATA_INSERT_CONCURRENCY=4
 
 # If specified, controls the Stacks Blockchain API mode. The possible values are:
 # * `readonly`: Runs the API endpoints without an Event Server that listens to events from a node and
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -900,9 +900,7 @@ export class PgWriteStore extends PgStore {
       const res = await sql`
         INSERT INTO ${sql(String(poxTable))} ${sql(batch)}
       `;
-      if (res.count !== batch.length) {
-        throw new Error(`Expected ${batch.length} inserts, got ${res.count}`);
-      }
+      assert(res.count === batch.length, `Expecting ${batch.length} inserts, got ${res.count}`);
     }
   }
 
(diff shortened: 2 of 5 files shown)


bestmike007:
@rafaelcr @zone117x please help me review this.

src/datastore/pg-write-store.ts
```diff
@@ -256,20 +255,30 @@ export class PgWriteStore extends PgStore {
...
-          await this.updatePoxSyntheticEvents(sql, entry.tx, 'pox3_events', entry.pox3Events);
-          await this.updatePoxSyntheticEvents(sql, entry.tx, 'pox4_events', entry.pox4Events);
-          await this.updateStxLockEvents(sql, entry.tx, entry.stxLockEvents);
-          await this.updateFtEvents(sql, entry.tx, entry.ftEvents);
-          await this.updateNftEvents(sql, entry.tx, entry.nftEvents);
-          await this.updateSmartContracts(sql, entry.tx, entry.smartContracts);
-          await this.updateNamespaces(sql, entry.tx, entry.namespaces);
-          await this.updateNames(sql, entry.tx, entry.names);
+        if (batchedTxData.length > 0) {
+          const q = new PQueue({ concurrency: 4 });
```
rafaelcr, line 259: Can you make `concurrency` configurable via ENV? Or why is it set to 4? Curious if you did any benchmarks or how did you get to that number

rafaelcr: Also, can you initialize the `PQueue` outside this block so we can reuse it across blocks? Will save CPU cycles especially during event replays

bestmike007: 4 is better than 1, and a default value not being too radical if someone is running postgres instance on non-SSD disk.

Added the env entry, please help me review the wording.

.env
```diff
@@ -43,6 +43,10 @@ PG_APPLICATION_NAME=stacks-blockchain-api
 # Limit to how many concurrent connections can be created, defaults to 10
 # PG_CONNECTION_POOL_MAX=10
 
+# Insert concurrency when processing new blocks
+# It'd better set it higher, e.g. 8 or 16, if your postgres is running on SSD and multiple CPU cores
+# INSERT_CONCURRENCY=4
```
rafaelcr, line 48: Please rename this to `STACKS_BLOCK_DATA_INSERT_CONCURRENCY`, also add a comment stating that this value should be set in harmony with `PG_CONNECTION_POOL_MAX` so all the concurrent writes have enough available connections to service them.

src/datastore/pg-write-store.ts
bestmike007: > initialize the PQueue outside this block

It's not safe, `await q.onIdle()` will wait for all tasks in the queue, which might come from another process.

Besides, the queue initialization overhead is small compares to all the tasks during event replays.

.env
bestmike007: It doesn't relate to connection pool size, the batch inserts happens within a transaction, which uses a single connection. Batching reduces interaction rounds, and sending requests in parallel reduces time to wait for responses of other irrelevant requests.

bestmike007: Renamed.

src/datastore/pg-write-store.ts
```diff
@@ -746,14 +759,35 @@ export class PgWriteStore extends PgStore {
...
+        case 'pox3_events':
+          assert('pox3Events' in entry);
+          events = entry.pox3Events;
+        case 'pox4_events':
+          assert('pox4Events' in entry);
+          events = entry.pox4Events;
+      }
+      if (events == null) {
+        throw new Error(`unknown pox table: ${poxTable}`);
+      }
```
zone117x, line 788: Fix switch-case fall-through and simplify types

```suggestion
  async updatePoxSyntheticEvents(
    sql: PgSqlClient,
    poxTable: PoxSyntheticEventTable,
    entries: DataStoreTxEventData[]
  ) {
    const values: PoxSyntheticEventInsertValues[] = [];
    for (const entry of entries) {
      let events: DbPoxSyntheticEvent[];
      switch (poxTable) {
        case 'pox2_events':
          events = entry.pox2Events;
          break;
        case 'pox3_events':
          events = entry.pox3Events;
          break;
        case 'pox4_events':
          events = entry.pox4Events;
          break;
        default:
          throw new Error(`unknown pox table: ${poxTable}`);
      }
```

bestmike007: The `entries` does not have to be `DataStoreTxEventData`, using typescript type narrowing is better:

image

Fixed switch-case fall-through.

zone117x: The readability of `updatePoxSyntheticEvents`'s function signature is questionable, although it is type safer for callers, so I'm okay with it 👍

src/datastore/pg-write-store.ts
rafaelcr: We process blocks in series so they would not overlap across blocks, but I think we can start with this approach as we test this PR in an env.

.env
rafaelcr: Ah, you're right, I was thinking from the read side of the API. New name is good.

rafaelcr (approved):
This looks good to me. We'd like to cut a pre-release Docker image with this PR and run a few tests so we can measure how it performs in the kubernetes pods, especially when interacting with postgres over a network connection.

I'll create a `beta` branch and target your PR into it so we can merge there and we can run tests on our end @bestmike007 , sound good? cc @zone117x

zone117x (approved):
Thanks a lot for this contribution!

bestmike007:
It'd be better if you run with different concurrency values, and then decide the default value.

Issue #1817, opened 2024-01-10
Ensure re-broadcasted mempool txs are "unpruned"


zone117x:
Mempool transactions need to be "restored" in the following scenario:
1. Mempool tx is received
2. Mempool tx is dropped after not being mined for a period of time
3. Mempool tx is re-broadcasted

It seems like step 3 may not be handled properly during re-ingestion of the mempool tx event:
https://github.com/hirosystems/stacks-blockchain-api/blob/c4cba69c270a3d823b61f8ead8a95fca68bf0cc3/src/datastore/pg-write-store.ts#L1770-L1771

The `ON CONFLICT` might need to instead 1) ensure the tx isn't already mined because these events can come out-of-order from /new_block, then 2) update `pruned` to false.

cc @lgalabru who identified this potential bug

zone117x:
Building off the query which is being updated in https://github.com/hirosystems/stacks-blockchain-api/pull/1815, I think the fix would look something like:

```sql
WITH settled AS (
    SELECT tx_id
    FROM txs
    WHERE
        tx_id IN ${sql(values.map(b => b.tx_id))} AND
        canonical = true AND
        microblock_canonical = true
),
inserted AS (
    INSERT INTO mempool_txs ${sql(values)}
    ON CONFLICT (tx_id) DO 
    UPDATE SET
        pruned = CASE 
            WHEN EXCLUDED.tx_id NOT IN (SELECT tx_id FROM settled) THEN false
            ELSE mempool_txs.pruned
        END
    RETURNING tx_id
),
pruned AS (
    UPDATE mempool_txs
    SET pruned = true
    WHERE
        tx_id IN (SELECT tx_id FROM settled) AND
        pruned = false
    RETURNING tx_id
),
count_update AS (
    UPDATE chain_tip SET
        mempool_tx_count = mempool_tx_count + (SELECT COUNT(*) FROM inserted),
        mempool_tx_count = mempool_tx_count - (SELECT COUNT(*) FROM pruned),
        mempool_updated_at = NOW()
)
SELECT tx_id FROM inserted

```

zone117x:
Closed by https://github.com/hirosystems/stacks-blockchain-api/pull/1823

PR #1823, opened 2024-01-12
fix: revive dropped mempool rebroadcasts

zone117x:
Fixes https://github.com/hirosystems/stacks-blockchain-api/issues/1817

### DIFF
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -1648,7 +1648,10 @@ export class PgWriteStore extends PgStore {
       // Revive mempool txs that were previously dropped
       const revivedTxs = await sql<{ tx_id: string }[]>`
         UPDATE mempool_txs
-        SET pruned = false, status = ${DbTxStatus.Pending}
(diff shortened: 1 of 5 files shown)


src/datastore/pg-write-store.ts
```diff
@@ -2329,7 +2349,7 @@ export class PgWriteStore extends PgStore {
     const updatedRows = await sql<{ tx_id: string }[]>`
       WITH restored AS (
         UPDATE mempool_txs
-        SET pruned = FALSE
+        SET pruned = FALSE, status = ${DbTxStatus.Pending}
```
zone117x, line 2352: I think this may have fixed an super long standing bug where after a series of re-orgs and dropped mempool txs, a mempool tx could "disappear" because this "restore" function wouldn't reset the `status` back to pending.

rafaelcr: ahh man, great catch

rafaelcr (approved):
LGTM, thanks!

Issue #1761, opened 2023-12-08
Implement parsing for `revoke-delegate-stx` pox event


zone117x:
PR https://github.com/stacks-network/stacks-core/pull/4094 adds a new pox event for the `revoke-delegate-stx` action. This should be parsed and stored in the pox events table similar to the other synthetic pox events.

zone117x:
These changes are now in PR https://github.com/stacks-network/stacks-core/pull/4157

janniks:
fixed by https://github.com/hirosystems/stacks-blockchain-api/pull/1829

PR #1829, opened 2024-01-16
fix: pox 4 revoke events

janniks:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1761

- updates rosetta for signer-key
- updates tests for signer-key
- adds helper for using descriptor wallet for address format tests
- adds timeouts to flaky tests
- some assumptions I'm not 100% certain about, but tests are green 😶

### DIFF
--- a/src/test-utils/test-helpers.ts
+++ b/src/test-utils/test-helpers.ts
@@ -54,9 +54,6 @@ import { coerceToBuffer, hexToBuffer, runMigrations, timeout } from '@hirosystem
 import { MIGRATIONS_DIR } from '../datastore/pg-store';
 import { getConnectionArgs } from '../datastore/connection';
 
-/** A (buff 33) signer key of all zeroes, to be used for testing */
-export const ZERO_SIGNER_KEY_BYTES = hexToBytes('00'.repeat(33));
-
 export async function migrate(direction: 'up' | 'down') {
   await runMigrations(MIGRATIONS_DIR, direction, getConnectionArgs());
 }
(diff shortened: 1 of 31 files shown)


migrations/1702134678728_pox_4_events.js
```diff
@@ -171,6 +171,8 @@ exports.up = pgm => {
       WHEN 'stack-aggregation-increase' THEN
         reward_cycle IS NOT NULL AND
         amount_ustx IS NOT NULL
+      WHEN 'revoke-delegate-stx' THEN
+        delegate_to IS NOT NULL
```
zone117x, line 175: Can you make this change in a new migration file? Otherwise it's a breaking change and requires a major version bump (and costly upgrade)

janniks: Ah yes, wanted to ask about this, if it's live somewhere. Will do 👍🏻

src/test-utils/test-helpers.ts
```diff
@@ -160,8 +162,10 @@ export async function standByForPoxCycle(
   do {
     await standByUntilBurnBlock(lastPoxInfo.current_burnchain_block_height! + 1, api, client);
     lastPoxInfo = await client.getPox();
-  } while (lastPoxInfo.current_cycle.id <= firstPoxInfo.current_cycle.id);
-  expect(lastPoxInfo.current_cycle.id).toBe(firstPoxInfo.next_cycle.id);
```
janniks, line 164: Not sure about this change. Basically `current_cycle.id` can be incorrect and shouldn't really be tested against. I'm thinking most tests rely on the rough behavior here and not the exact value.

zone117x (approved):

PR #1833, opened 2024-01-17
feat: add `/extended/v2/smart-contracts/status` endpoint

rafaelcr:
Retrieves the deployment status of multiple smart contracts at once

### DIFF
--- a/docs/api/smart-contracts/get-smart-contracts-status.example.json
+++ b/docs/api/smart-contracts/get-smart-contracts-status.example.json
@@ -1,19 +1,31 @@
-[
-  {
-    "contract_id": "SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.swap-helper-bridged-v1-1",
-    "status": "success",
-    "tx_id": "0x8542d28e427256ea3c29dcd8793222891999ceff4ef1bb062e2f21cb6def6884",
-    "block_height": 111021
+{
+  "SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.swap-helper-bridged-v1-1": {
+    "found": true,
+    "status": {
+      "contract_id": "SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.swap-helper-bridged-v1-1",
+      "status": "success",
+      "tx_id": "0x8542d28e427256ea3c29dcd8793222891999ceff4ef1bb062e2f21cb6def6884",
+      "block_height": 111021
+    }
   },
-  {
-    "contract_id": "SP1JTCR202ECC6333N7ZXD7MK7E3ZTEEE1MJ73C60.name-registrar",
-    "status": "success",
-    "tx_id": "0x6e1114cce8c6f2e9c8130f9acd75d67bb667ae584f882acdd2db6dd74e6cbe5e",
-    "block_height": 113010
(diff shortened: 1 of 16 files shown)


docs/entities/smart-contracts/smart-contract-status.schema.json
```diff
@@ -0,0 +1,29 @@
...
+      "description": "Deployment transaction ID"
+    },
+    "contract_id": {
+      "type": "string",
+      "description": "Smart contract ID"
+    },
+    "block_height": {
+      "type": "integer",
+      "description": "Height of the transaction confirmation block"
+    }
```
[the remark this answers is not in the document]
zone117x: I guess not if this endpoint is explicitly for getting "status" of contracts -- as opposed to something similar to /v1/contract with support for multiple contract ids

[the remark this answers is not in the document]
rafaelcr: Exactly, this is only for deployment status

src/datastore/pg-store-v2.ts
```diff
@@ -230,4 +234,40 @@ export class PgStoreV2 extends BasePgStoreModule {
...
+
+      // Search confirmed txs.
+      const confirmed = await sql<DbSmartContractStatus[]>`
+        SELECT DISTINCT ON (smart_contract_contract_id) smart_contract_contract_id, tx_id, block_height, status
+        FROM txs
+        WHERE type_id IN ${sql([DbTxTypeId.SmartContract, DbTxTypeId.VersionedSmartContract])}
+          AND smart_contract_contract_id IN ${sql(contractArray)}
+          AND canonical = TRUE
+          AND microblock_canonical = TRUE
+          AND status = ${DbTxStatus.Success}
```
rafaelcr, line 251: Not sure if I should be doing this condition @zone117x because it would be useful to know if a contract failed or something (even if it's not in the mempool anymore). Thoughts?

docs/api/smart-contracts/get-smart-contracts-status.example.json
```diff
@@ -0,0 +1,31 @@
+{
+  "SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.swap-helper-bridged-v1-1": {
+    "found": true,
+    "status": {
```
BLuEScioN, line 4: Maybe it would be better to call this result, since not all the properties are related to status

docs/generated.d.ts
```diff
@@ -3059,6 +3063,40 @@ export interface TxSearchResult {
...
+ */
+export interface SmartContractStatus {
+  /**
+   * Smart contract deployment transaction status
+   */
+  status: string;
+  /**
+   * Deployment transaction ID
+   */
+  tx_id: string;
```
BLuEScioN, line 3087: Just curious, how hard is it to get the tx_status added to this call? This would mean needing to one less call to go from contract txid to requesting the tx list endpoint for tx status

rafaelcr: `status` is actually the `tx_status`, perhaps I should rename it

src/datastore/pg-store-v2.ts
zone117x: Maybe you could instead add the `status` to the ORDER BY clause, that way for a given contract, you'll get either the successful deployment, or the last failed one?

rafaelcr: Great idea, I'll do that

PR #1836, opened 2024-01-19
fix: line spacing in `/v2/fees/transaction`

obycode:
## Description

This fixes some line spacing issues in the documentation.

1. The formatting is not as intended due to the treatment of line-breaks

Example:
image

In the screenshot above, it is clear that the information about 200 responses is not intended to be in the bullet about `CostEstimationDisabled`.


obycode:
Confirmed this looks much better in the vercel page 
image

rafaelcr (approved):
Thanks 👌🏻

zone117x (approved):

PR #1850, opened 2024-01-26
test: fix flaky postgres teardown in tests

zone117x:
* Move postgres down-migration logic out of teardown
* Remove timeouts in postgres shutdown call

These changes appear to remove the error log messages (which are sometimes fatal deadlocks) that would appear after tests finished.

### DIFF
--- a/src/tests-2.5/global-setup.ts
+++ b/src/tests-2.5/global-setup.ts
@@ -40,6 +40,7 @@ export default async (): Promise<void> => {
   process.env.PG_DATABASE = 'postgres';
   process.env.STACKS_CHAIN_ID = '0x80000000';
 
+  await migrate('down');
   await migrate('up');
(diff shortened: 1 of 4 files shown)


src/tests-2.5/global-setup.ts
```diff
@@ -40,6 +40,7 @@ export default async (): Promise<void> => {
   process.env.PG_DATABASE = 'postgres';
   process.env.STACKS_CHAIN_ID = '0x80000000';
 
+  await migrate('down');
```
janniks, line 43: This needed for running multiple tests with the same postgres?

janniks (approved):

src/tests-2.5/global-setup.ts
zone117x: The `/test-2.5/*` tests are one shot, only one can run since they pollute the stacks-node and bitcoind. Preserving postgres state makes no difference

PR #1854, opened 2024-01-29
update to new url path

ryanwaits:
## Description

Main API overview page [here](https://docs.hiro.so/api) currently references an old url path. This change updates to the new path that renders the proper sidebar in docs.


ryanwaits:
reference: https://github.com/hirosystems/docs/issues/600

Issue #1855, opened 2024-01-31
Rosetta: use `/v2/fees/transaction` for fees


GentaKaieda01:
**Describe the bug**
The [getEstimatedTransferFee](https://github.com/hirosystems/stacks-blockchain-api/blame/develop/src/core-rpc/client.ts#L296) rpc endpoint is calling 'v2/fees/transfer' which is deprecated and giving low fee estimation (0.00027 STX). Can we update the endpoint to call [v2/fees/transaction](https://github.com/stacks-network/stacks-core/blob/master/docs/rpc/openapi.yaml#L283) instead? 

**To Reproduce**
1. Call the getEstimatedTransferFee endpoint

PR #1858, opened 2024-02-02
Use updated fee estimation endpoint in Rosetta transaction construction

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1855

(also cherrypicked an unrelated fix from the Nakamoto branch)


rafaelcr (approved):
Thanks @zone117x

PR #1861, opened 2024-02-05
update broken anchors for deprecated endpoints

ryanwaits:
## Description

Caught a few more broken anchors causing some build issues with docs for some of the deprecated endpoints.
 
## Checklist
- [x] Update anchors to point to new active endpoints
- [x] Prioritize `docs.hiro.so` routing and update urls to `/api`

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -123,7 +123,7 @@ paths:
         required: true
         schema:
           type: string
-          example: '2N4M94S1ZPt8HfxydXzL2P7qyzgVq7MHWts'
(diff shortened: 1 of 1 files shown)


ryanwaits:
so im noticing - aside from the `get-block` and `get-blocks` inconsistency, that the relative paths for the `openapi.yaml` actually work properly within the context of redoc, but those paths do not work properly relative to docusaurus when they get pulled in to the docs repo

will try out a few things, but in the meantime will move this pr into draft

diwakergupta (approved):
There are a bunch of unrelated changes in this PR (spaces, quotes etc). No objection from my end, but let's a 👍🏽 from @rafaelcr or @zone117x

PR #1867, opened 2024-02-19
fix: report placeholder in prom metrics for invalid request paths

### LINKED ISSUE #1826: Prometheus metric label exceeds length limit
I'm seeing errors in Promtheus when collecting some metrics from the API:

```
server returned HTTP status 400 Bad Request: received a series whose label value length exceeds the limit, value: '/v1/addresses/stacks/%!D(MISSING)9%!A(MISSING)%!D(MISSING)8%!A(MISSING)7%D8%!A(MISSING)3/Nuri%!A(MISSING)nja%20%!E(MISSING)2%!A(MISSING)%!A(MISSING)0%!E(MISSING)F%!B(MISSING)8%!F(MISSING)%!F(MISSING)0%!F(MISSING)%87%!D(MISSING)8%!A(MISSING)A%!D(MISSING)9%20%!D(MISSING)8%!A(MISSING)5%!D(MISSING)8%!B(MISSING)1%!D(MISSING)8%!A(MISSING)C%!D(MISSING)8%!A(MISSING)7%!D(MISSING)8%!B(MISSING)9%D8%!A(MISSING)5%!D(MISSING)9%!A(MISSING)%!D(MISSING)8%!B(MISSING)5%!D(MISSING)8%!A(MISSING)7%!D(MISSING)9%20%!D(MISSING)8%!A(MISSING)7%!D(MISSING)9%D8%!A(MISSING)3%!D(MISSING)8%!B(MISSING)1%!D(MISSING)8%!B(MISSING)' (truncated) series: 'http_request_duration_seconds_bucket{cluster=\"<redacted>\", container=\"<redacted>\", endpoint=\"tcp-metrics\", instance=\"10.0.0.157:9153\", job=\"<redacted>\", le=\"0.8\", method=\"get\"'
```

It seems a label being added to the `http_request_duration_seconds_bucket` metric is too long, and caused some metrics to be dropped.

zone117x:
Fixes #1826

When a request path is invalid because it doesn't have a route handler, change the prometheus reporting so that the path name is the placeholder `<invalid_path>`. This fixes the issue of very long path names, and fixes the issue of log spam from many unique invalid paths.

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -120,7 +120,7 @@ export async function startApiServer(opts: {
                 break;
               }
             }
-            return pathTemplate;
+            return '<invalid_path>';
           } catch (error) {
             logger.warn(`Warning: ${error}`);
             return path;
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -110,14 +110,13 @@ export async function startApiServer(opts: {
           // Get the url pathname without a query string or fragment
           // (note base url doesn't matter, but required by URL constructor)
           try {
-            let pathTemplate = new URL(path, 'http://x').pathname;
+            const pathTemplate = new URL(path, 'http://x').pathname;
             // Match request url to the Express route, e.g.:
             // `/extended/v1/address/ST26DR4VGV507V1RZ1JNM7NN4K3DTGX810S62SBBR/stx` to
             // `/extended/v1/address/:stx_address/stx`
             for (const pathRegex of routes) {
               if (pathRegex.regexp.test(pathTemplate)) {
-                pathTemplate = pathRegex.path;
-                break;
+                return pathRegex.path;
               }
             }
             return '<invalid_path>';


src/api/init.ts
```diff
@@ -120,7 +120,7 @@ export async function startApiServer(opts: {
                 break;
               }
             }
-            return pathTemplate;
+            return '<invalid_path>';
```
CharlieC3, line 123: I might be misunderstanding something, but won't this always return `<invalid_path>`, even if `pathTemplate` is successfully set to a valid `pathRegex.path` in the for loop's condition above?

zone117x: Good catch, fixed!

CharlieC3 (approved):
Thanks for addressing this! We should wait for @rafaelcr but LGTM! :shipit:

rafaelcr (approved):

Issue #1877, opened 2024-02-28
Add block receipt time property


zone117x:
Stacks blocks currently use Bitcoin block timestamps. Nakamoto blocks will need their own timestamps. Nakamoto blocks do not include this data, so the best we can do is record time of receipt by the API, similar to how mempool tx times are recorded.

zone117x:
Implemented in https://github.com/hirosystems/stacks-blockchain-api/pull/1886

PR #1886, opened 2024-03-14
feat: nakamoto block timestamps

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1877

Adds timestamps to Nakamoto blocks, with some important caveats. 

Nakamoto block headers do not include timestamps, see related issues in core:
https://github.com/stacks-network/stacks-core/issues/4220
https://github.com/stacks-network/stacks-core/issues/4541

To work around this, we record the receipt time (i.e. `Date.now()`) of when we receive a block and store that as the block time. The code is structured such if Nakamoto block times are added to the event payload, it will be trivial to swap our usage to the values from the events.

If IDB mode (e.g. event-replay) is enable, then Nakamoto block timestamps just use the burn block timestamps.

To preserve backwards compatibility with older sql databases, the burn block timestamp will be used if the receipt time was not recorded.


docs/entities/blocks/block.schema.json
```diff
@@ -7,6 +7,8 @@
     "canonical",
     "height",
     "hash",
+    "stacks_block_time",
+    "stacks_block_time_iso",
```
rafaelcr, line 11: Could a better name be just `block_time`? Given that we use `block_` prefix in all other Stacks block properties

src/datastore/pg-write-store.ts
```diff
@@ -1790,10 +1791,20 @@ export class PgWriteStore extends PgStore {
...
-        logger.error(e, `failed to run mempool stats update`);
+      } catch (e: unknown) {
+        if (
+          e instanceof Error &&
+          'code' in e &&
+          ['CONNECTION_ENDED', 'CONNECTION_ENDED'].includes((e as any).code)
+        ) {
+          logger.info(`Skipping mempool stats query because ${(e as any).code}`);
+        } else {
+          logger.error(e, `failed to run mempool stats update`);
```
rafaelcr, line 1806: Haha thank you! I've been seeing this pop up in all tests

src/event-stream/event-server.ts
```diff
@@ -232,10 +235,20 @@ async function handleBlockMessage(
 ): Promise<void> {
   const ingestionTimer = stopwatch();
   const counts = newCoreNoreBlockEventCounts();
+  // If running in IBD mode, we use the parent burn block timestamp as the receipt date,
+  // otherwise, use the current timestamp.
+  // TOOD: if the core node can give use the stacks-block count/index for the current tenure/burn_block then we should
```
rafaelcr, line 240:
```suggestion
  // TODO: if the core node can give use the stacks-block count/index for the current tenure/burn_block then we should
```

src/helpers.ts
```diff
@@ -31,6 +31,11 @@ export function getIbdBlockHeight(): number | undefined {
   }
 }
 
+export function isBlockInIbdMode(blockHeight: number): boolean {
+  const ibdBlockHeight = getIbdBlockHeight();
```
rafaelcr, line 35: I think you should include the `isEventReplay` property from `PgStore` here because event replay does not necessarily mean IBD if I remember correctly

docs/entities/blocks/block.schema.json
zone117x: Yep! Done

src/helpers.ts
zone117x: Ahh that's right, fixed!

rafaelcr (approved):

PR #1897, opened 2024-03-19
feat: add new PoX endpoints

rafaelcr:
Add new pox endpoints:

```
/extended/v2/pox/cycles
/extended/v2/pox/cycles/:number
/extended/v2/pox/cycles/:number/signers
/extended/v2/pox/cycles/:number/signers/:key
/extended/v2/pox/cycles/:number/signers/:key/stackers
```


src/datastore/pg-store-v2.ts
```diff
@@ -269,4 +279,133 @@ export class PgStoreV2 extends BasePgStoreModule {
...
+        await sql`SELECT cycle_number FROM pox_cycles WHERE cycle_number = ${args.cycle_number} LIMIT 1`;
+      if (cycleCheck.count === 0)
+        throw new InvalidRequestError(`PoX cycle not found`, InvalidRequestErrorType.invalid_param);
+      const results = await sql<(DbPoxCycleSigner & { total: number })[]>`
+        SELECT
+          signing_key, weight, stacked_amount, weight_percent, stacked_amount_percent,
+          COUNT(*) OVER()::int AS total
+        FROM pox_sets
+        WHERE canonical = TRUE AND cycle_number = ${args.cycle_number}
+        ORDER BY weight DESC
```
zone117x, line 336: I think it should be common for multiple entries to have the same weight, so I think this needs to also sort by another column so the order is deterministic per page. Probably something like
```sql
ORDER BY weight DESC, stacked_amount DESC, signing_key
```

zone117x (approved):
Left a comment about fixing one of the ORDER clauses, otherwise this looks great! Thanks @rafaelcr!

PR #1904, opened 2024-03-21
test: add regtest solo stacking

janniks:
Added one of the tests from the TSVs. Some stuff still seems off, _double checking._

> Haven't validated every value here, assuming some are correct. 

- fixes two minor issues (one covered in #1907)


src/tests/pox-tests.ts
```diff
@@ -141,4 +141,116 @@ describe('PoX tests', () => {
...
+      limit: 100,
+      offset: 0,
+      results: [
+        {
+          pox_address: '1NQAp2ecV2zUQfz7y2hLvJTU5scnFTSACM',
+          stacked_amount: '2500170000000000',
+          stacker_address: 'ST3NBRSFKX28FQ2ZJ1MAKX58HKHSDGNV5N7R21XCP', // signer
+        },
+        {
+          pox_address: 'mhYeZXrSEuyf2wbJ14qZ2apG7ofMLDj9Ss',
```
janniks, line 248: This pox_address is different in the test run -- not sure if an encoding error somewhere

src/tests/pox-tests.ts
```diff
@@ -141,4 +141,116 @@ describe('PoX tests', () => {
...
+      total: 3,
+    });
+
+    const signer = await supertest(api.server).get(
+      `/extended/v2/pox/cycles/6/signers/0x029fb154a570a1645af3dd43c3c668a979b59d21a46dd717fd799b13be3b2a0dc7`
+    );
+    expect(signer.status).toBe(200);
+    expect(signer.type).toBe('application/json');
+    expect(JSON.parse(signer.text)).toStrictEqual({
+      signing_key: '0x029fb154a570a1645af3dd43c3c668a979b59d21a46dd717fd799b13be3b2a0dc7',
```
janniks, line 226: This signer_key is different in the test run -- I'm expecting the same as given in the path parameter.

src/tests/pox-tests.ts
```diff
@@ -141,4 +141,116 @@ describe('PoX tests', () => {
...
+
+    const cycles = await supertest(api.server).get(`/extended/v2/pox/cycles`);
+    expect(cycles.status).toBe(200);
+    expect(cycles.type).toBe('application/json');
+    expect(JSON.parse(cycles.text)).toStrictEqual({
+      limit: 20,
+      offset: 0,
+      results: [
+        {
+          block_height: 14,
```
janniks, line 164: What is the meaning of `block_height` here?

src/tests/pox-tests.ts
janniks: It's `132hGUmTRtYQFq7gHVsBCfbwFp4eNsQvBi`, so I'm assuming either core or the API is encoding as mainnet, instead of testnet

src/tests/pox-tests.ts
janniks: fixed

src/datastore/pg-store-v2.ts
```diff
@@ -298,7 +298,7 @@ export class PgStoreV2 extends BasePgStoreModule {
         limit,
         offset,
         results: results,
-        total: results[0].total,
+        total: results[0]?.total ?? 0,
```
janniks, line 301: throws if no cycle have been registered yet (not likely in prod, but can happen in testing)

src/tests/pox-tests.ts
```diff
@@ -141,4 +141,160 @@ describe('PoX tests', () => {
...
+          cycle_number: 6, // !!! next cycle (even though we're still in cycle 5)
+          total_signers: 3, // no addition signer
+          total_weight: 21, // additional weight from steph's stacking
+        })
+      );
+    });
+
+    test('snapshot 3', async () => {
+      await importEventsFromTsv(
+        'src/tests/tsv/regtest-env-pox-4-stack-stx-in-reward-phase-S3.tsv',
```
rafaelcr, line 193: I think this TSV is just 1 block shy of when the prepare phase ends (assuming it's 5 blocks wide) 😅 
Could you generate another snapshot that is further ahead?

rafaelcr (approved):
LGTM

src/tests/pox-tests.ts
zone117x: Hmm yes we get the bitcoin address directly from the node. Perhaps the `pox-addr` clarity value tuple generated in the tests used a mainnet version byte? Otherwise it does sound like a potential issue in core. Either way, I don't think it's an issue in the API codebase, so we can merge this.

janniks: Found it. The TSV import gets the chainId from ENV and uses that to parse the pox_addr_raw

zone117x (approved):

PR #1910, opened 2024-03-25
fix: add nakamoto testnet to openapi docs

rafaelcr:


ryanwaits:
nice! i actually already added this url manually, so this lgtm!

ryanwaits (approved):

PR #1915, opened 2024-03-28
fix: other empty result responses

rafaelcr:


janniks (commented):
count or length?

rafaelcr:
@janniks `count` is a property added by postgres.js that returns the number of resulting rows 👍

janniks (approved):
Cool 🫡

PR #1925, opened 2024-04-05
feat: ensure events are inserted into the raw event request table

zone117x:
Issue found by @csgui where the `event_observer_requests` table can be missing events even if the rest of the event ingestion was successful. 

This appears to be because the function for inserting into `event_observer_requests` was ran after the event server http request has been completed. So when a large block was received, it may have been ingested into the regular tables, but could run into an OOM error on the `event_observer_requests` insert.

This PR fixes the issue by ensuring the http request is not completed until insert into `event_observer_requests` is completed. Also the amount of JSON parsing/stringifying has been reduced in an attempt to reduce likelihood of an OOM when receiving a large `/new_block` event payload.


src/datastore/pg-write-store.ts
```diff
@@ -155,27 +155,28 @@ export class PgWriteStore extends PgStore {
...
+        {
+          id: string;
+          receive_timestamp: string;
+          event_path: string;
+        }[]
+      >`INSERT INTO event_observer_requests(
         event_path, payload
-      ) values(${eventPath}, ${payload})
-      RETURNING id, receive_timestamp::text, event_path, payload::text
+      ) values(${eventPath}, ${sql.unsafe(`'${payload}'::jsonb`)})
```
zone117x, line 171: This lets us pass a serialized json string directly to postgres without the memory overhead of a JSON.parse/JSON.stringify in the client.

src/event-stream/event-server.ts
```diff
@@ -866,23 +866,22 @@ export async function startEventServer(opts: {
...
       }
       logger.debug(`${eventPath} ${payload}`, { component: 'stacks-node-event' });
     }
-  });
+  };
 
   app.use(loggerMiddleware);
-
-  app.use(bodyParser.json({ type: 'application/json', limit: '500MB' }));
+  app.use(bodyParser.text({ type: 'application/json', limit: '500MB' }));
```
rafaelcr, line 884: What happens is the size limit is reached here? Should we make this size larger?

src/event-stream/event-server.ts
```diff
@@ -911,101 +910,96 @@ export async function startEventServer(opts: {
...
+        await handleRawEventRequest(req);
         res.status(200).json({ result: 'ok' });
-        next();
       } catch (error) {
         logger.error(error, 'error processing core-node /new_block');
         res.status(500).json({ error: error });
       }
-    }),
-    handleRawEventRequest
+    })
```
zone117x, line 926: This change where `handleRawEventRequest` is called fixes the sql inconsistency bug that likely caused event ingestion to succeed even if the payload failed to write into the raw events table.

src/event-stream/event-server.ts
zone117x: The request would be rejected with an HTTP 4xx error (which is generally what we want, as opposed to silently fail). We can increase the limit if we expect an event to be larger than 500MB. However, 500MB raw http request body would already use something like 2GB of memory after JSON parsing and postgres client serialization. I don't think our writer environments expect to have much more memory than that (2GB in payload processing + normal operating memory)

rafaelcr (approved):

csgui (approved):
Code verified and tested. LGTM!

csgui:
@zone117x Just validated this fix against an archive synced from genesis. Please check the [conversation](https://github.com/hirosystems/devops/issues/1674#issuecomment-2140941824) on this devops issue for more context. 

Blocks containing a large number of transactions are now included in the archive file.

For example, block 137843, which was the root cause of this issue, is now present in the archive file after the fix. Therefore, for a block to be present in the archive file, it must have also been present in the `event_observer_requests` table.

Fix LGTM.

zone117x:
Awesome, thanks for the thorough testing @csgui!

Issue #1901, opened 2024-03-20
Support pox-4 in Rosetta stacking operations


zone117x:
This is mentioned in https://github.com/hirosystems/stacks-blockchain-api/issues/1735. I'm creating a new issue explicitly to track this.

PR #1928, opened 2024-04-08
feat: rosetta pox4 support

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1901
Implement Rosetta pox-4 `stack-stx` support.

This PR introduces three modes to use pox4 `stack-stx` in Rosetta:
1) Specify `signer-private-key` and the Rosetta layer will generate the signer signature used by the stack-stx transaction.
2) Specify `signer-signature` (generated external to the Rosetta layer) and it will be used in the stack-stx transaction.
3) Do not specify `signer-private-key` or `signer-signature`, and the stack-stx transaction will specify `none` for this field -- only works if the signer has added the caller to their allow-list.

Additionally:
* The stack-stx `auth-id` field is optional: if not specified, the Rosetta layer will automatically generate a random ID.
* The `reward-cycle-id` field is optional: if not specified, the Rosetta layer will pick the next cycle similar to how it already does with the `burn-block-height` field.
* The `pox-max-amount` field is optional: if not specified, the STX `amount` will be used.

### DIFF
--- a/docs/entities/rosetta/rosetta-construction-options.schema.json
+++ b/docs/entities/rosetta/rosetta-construction-options.schema.json
@@ -102,7 +102,7 @@
     },
     "pox_max_amount": {
       "type": "string",
-      "description": "The maximum amount of STX to stack for PoX"
+      "description": "The maximum amount of STX to stack for PoX. If not specified, the `amount` will be used as the `max-amount` for the PoX transaction."
     },
     "pox_auth_id": {
       "type": "string",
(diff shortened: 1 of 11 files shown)


docs/entities/rosetta/rosetta-construction-options.schema.json
```diff
@@ -87,6 +91,22 @@
     "signer_key": {
       "type": "string",
       "description": "The hex-encoded signer key (buff 33) for PoX."
+    },
+    "signer_private_key": {
+      "type": "string",
+      "description": "The hex-encoded signer private key for PoX. Must be specified for PoX transactions if `signer_signature` is not specified"
```
janniks, line 97: `is not specified` is now outdated for these both

zone117x: Fixed

janniks (approved):
👏

PR #1929, opened 2024-04-09
chore: update rosetta pox-4 checks

janniks:
minor updates

- add check for `authID`
- allow stacking with `none` signature (possible, but we can disallow this flow if we want to)


zone117x (approved):
Thanks @janniks!

Issue #1945, opened 2024-04-15
Improve STX testnet faucet


zone117x:
We've ran into the TOO_MUCH_CHAINING issue a few times during congestion, bitcoin re-orgs and block storm. 

One approach to improving the faucet is to use something like the send-many stx smart contract, and when new faucet request is made, perform an RBF for any previous faucet send-many contract calls in the mempool.

If we can't depend on the send-many contract to already be deployed, it may be possible to use the same approach except with performing a contract-deploy, assuming it's possible to send stx during contract instantiation. 

Another approach could be just spreading out the faucet funds to multiple seed accounts, and adjusting the API code to try sending from all given accounts rather than just the one.

smcclellan:
Closing as the above PR is solving this issue.

PR #1946, opened 2024-04-15
feat: support multiple STX faucet source accounts

rafaelcr:
Related to #1945

### DIFF
--- a/src/api/routes/faucets.ts
+++ b/src/api/routes/faucets.ts
@@ -310,8 +310,13 @@ export function createFaucetRouter(db: PgWriteStore): express.Router {
           } else if (
             sendTxResults.every(res => res.status === TxSendResultStatus.TooMuchChaining)
           ) {
-            retrySend = true;
-            stxKeyIndex++;
+            // Try with the next key in case we have one.
+            if (stxKeyIndex + 1 === STX_FAUCET_KEYS.length) {
+              retrySend = false;
+            } else {
+              retrySend = true;
+              stxKeyIndex++;
+            }
           } else {
             retrySend = false;
           }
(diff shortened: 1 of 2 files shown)


src/api/routes/faucets.ts
```diff
@@ -271,9 +267,9 @@ export function createFaucetRouter(db: PgWriteStore): express.Router {
         let sendSuccess: { txId: string; txRaw: string } | undefined;
         let lastSendError: Error | undefined;
         do {
-          const tx = await generateTx(networks[0], nextNonce, fee);
+          const tx = await generateTx(STX_FAUCET_NETWORKS[0], nextNonce, fee);
```
rafaelcr, line 270: @zone117x I think this is an old bug, even though there's a foreach loop right below, the only transaction generated is for the network with index zero... Do we still need to support multiple networks?

src/api/routes/faucets.ts
```diff
@@ -148,6 +144,10 @@ export function createFaucetRouter(db: PgWriteStore): express.Router {
   const FAUCET_STACKING_WINDOW = 2 * 24 * 60 * 60 * 1000; // 2 days
   const FAUCET_STACKING_TRIGGER_COUNT = 1;
 
+  const STX_FAUCET_NETWORKS = getStxFaucetNetworks();
+  const STX_FAUCET_KEYS = (process.env.FAUCET_PRIVATE_KEY ?? testnetKeys[0].secretKey).split(',');
+  let stxKeyIndex = 0;
```
zone117x, line 149: I think this needs to be stateless, so the loop just tries sending from all accounts when getting a TooMuchChaining

rafaelcr: Just so I understand: You mean always try with the first one and upon failure retry with all others at the same time? Wouldn't that send more STX than necessary in case all other txs succeed?

zone117x: Start on first one, upon failure go to next one, upon success then break the loop. Does that make sense?

rafaelcr: Done

zone117x (approved):
LGTM!

Issue #1955, opened 2024-04-19
Rosetta should default to block tip


zone117x:
A regression was made in the Rosetta layer where endpoints started requiring the block height rather than assuming the latest block when left out.

PR #1956, opened 2024-04-19
fix: rosetta account endpoint should assume chain tip if block not specified

zone117x:
Fixes https://github.com/hirosystems/stacks-blockchain-api/issues/1955


rafaelcr (approved):
LGTM

PR #1975, opened 2024-05-08
feat: add signer_address to pox signer endpoints

zone117x:
Adds the p2pkh stacks address of the signer pubkey, which is used for "vote for aggregate key" txs. Helpful for debugging and matching signers to known identities.

cc @hstove


hstove (approved):
LGTM, ty!

rafaelcr (approved):
LGTM

PR #1987, opened 2024-06-03
feat: include solo and pooled stackers in signer stacker endpoints

zone117x:
Include pooled and solo stackers in the pox signer endpoints.

* The properties `solo_stacker_count` and `pooled_stacker_count` added to `/extended/v2/pox/cycles/:cycle_number/signers` and `/extended/v2/pox/cycles/:cycle_number/signers/:signer_key`
* The `/extended/v2/pox/cycles/:cycle_number/signers/:signer_key/stackers` endpoint now returns the list of all stackers (both pooled and solo), and a `stacker_type: 'solo' | 'pooled'` field.

### DIFF
--- a/src/datastore/common.ts
+++ b/src/datastore/common.ts
@@ -1086,8 +1086,8 @@ export interface DbPoxCycleSigner {
   stacked_amount: string;
   weight_percent: number;
   stacked_amount_percent: number;
-  // TODO: Figure this out
-  // total_stackers: number;
+  pooled_stacker_count: number;
+  solo_stacker_count: number;
 }
 
 export interface DbPoxCycleSignerStacker {
--- a/src/datastore/common.ts
+++ b/src/datastore/common.ts
@@ -1096,6 +1096,7 @@ export interface DbPoxCycleSignerStacker {
   pox_addr: string;
   name: string;
   amount_ustx: string;
+  stacker_type: 'solo' | 'pooled';
 }
 
 interface ReOrgEntities {
--- a/src/datastore/pg-store-v2.ts
+++ b/src/datastore/pg-store-v2.ts
@@ -767,19 +767,66 @@ export class PgStoreV2 extends BasePgStoreModule {
           InvalidRequestErrorType.invalid_param
         );
       const results = await sql<(DbPoxCycleSignerStacker & { total: number })[]>`
-        WITH stackers AS (
-          SELECT DISTINCT ON (stacker) stacker, locked, pox_addr, amount_ustx, name
-          FROM pox4_events
-          WHERE canonical = true
-            AND microblock_canonical = true
-            AND start_cycle_id <= ${args.cycle_number}
-            AND (end_cycle_id >= ${args.cycle_number} OR end_cycle_id IS NULL)
-            AND signer_key = ${args.signer_key}
-          ORDER BY stacker, block_height DESC, tx_index DESC, event_index DESC
+        WITH signer_keys AS (
+            SELECT DISTINCT ON (stacker) stacker, signer_key
+            FROM pox4_events
+            WHERE canonical = true AND microblock_canonical = true
+                AND name in ('stack-aggregation-commit-indexed', 'stack-aggregation-commit')
+                AND start_cycle_id <= ${args.cycle_number}
+                AND end_cycle_id >= ${args.cycle_number}
+            ORDER BY stacker, block_height DESC, tx_index DESC, event_index DESC
+        ), delegated_stackers AS (
+            SELECT DISTINCT ON (main.stacker) 
+                main.stacker, 
+                sk.signer_key,
+                main.locked,
+                main.pox_addr,
+                main.name,
+                main.amount_ustx,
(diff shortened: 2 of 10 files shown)


src/datastore/common.ts
```diff
@@ -1086,8 +1086,8 @@ export interface DbPoxCycleSigner {
   stacked_amount: string;
   weight_percent: number;
   stacked_amount_percent: number;
-  // TODO: Figure this out
```
janniks, line 1089: 👏

src/datastore/pg-store-v2.ts
```diff
@@ -629,12 +629,52 @@ export class PgStoreV2 extends BasePgStoreModule {
...
-        FROM pox_sets
-        WHERE canonical = TRUE AND cycle_number = ${args.cycle_number}
-        ORDER BY weight DESC, stacked_amount DESC, signing_key
+        WITH signer_keys AS (
+            SELECT DISTINCT ON (stacker) stacker, signer_key
+            FROM pox4_events
+            WHERE canonical = true AND microblock_canonical = true
+                AND name in ('stack-aggregation-commit-indexed', 'stack-aggregation-commit')
+                AND start_cycle_id <= ${args.cycle_number}
+                AND end_cycle_id >= ${args.cycle_number}
```
janniks, line 638: Can this be an exact match for start and an exact match +1 for end, since the agg events will always target exactly one cycle? Might be better perf, either way the query might be more ~~readable~~understandable.

src/datastore/pg-store-v2.ts
```diff
@@ -629,12 +629,52 @@ export class PgStoreV2 extends BasePgStoreModule {
...
+            ORDER BY main.stacker, main.block_height DESC, main.microblock_sequence DESC, main.tx_index DESC, main.event_index DESC
+        ), solo_stackers AS (
+            SELECT DISTINCT ON (stacker) 
+                stacker, 
+                signer_key
+            FROM pox4_events
+            WHERE canonical = true AND microblock_canonical = true
+                AND name in ('stack-stx', 'stacks-increase', 'stack-extend')
+                AND start_cycle_id <= ${args.cycle_number} 
+                AND end_cycle_id >= ${args.cycle_number}
```
janniks, line 660: This one I'll want to double check or create a test for. Might be `>` for end. (Non inclusive style)

janniks (changes requested):

src/datastore/pg-store-v2.ts
zone117x: That should be 
```sql
	AND start_cycle_id = ${args.cycle_number}
	AND end_cycle_id = (${args.cycle_number} + 1) -- this one probably isn't required?
```
right?

rafaelcr (approved):
💯

src/datastore/pg-store-v2.ts
janniks: This condition works for the aggregation events

src/datastore/pg-store-v2.ts
janniks: Yes, I'm pretty sure all events have the end as non-inclusive (the unlock height cycle), so the end condition should be `end_cycle_id > ${args.cycle_number}` for all cases. _(for agg `== +1` is fine, but both would work)_

janniks (changes requested):

src/datastore/pg-store-v2.ts
```diff
@@ -654,10 +694,50 @@ export class PgStoreV2 extends BasePgStoreModule {
...
+                AND main.start_cycle_id <= ${args.cycle_number} 
+                AND main.end_cycle_id >= ${args.cycle_number}
+            ORDER BY main.stacker, main.block_height DESC, main.microblock_sequence DESC, main.tx_index DESC, main.event_index DESC
+        ), solo_stackers AS (
+            SELECT DISTINCT ON (stacker) 
+                stacker, 
+                signer_key
+            FROM pox4_events
+            WHERE canonical = true AND microblock_canonical = true
+                AND name in ('stack-stx', 'stacks-increase', 'stack-extend')
```
janniks, line 723: Comment for `stack-extend`, but I think it's ok as is. 

`stack-extend` can be used to change the signing-key, so in some cases a `DISTINCT` might strip one of the keys in a result.
Here, I believe we should only have one valid key/event, since the extend can't override a valid event, but would only append a later cycle entry. But might be trickier for a potential future range query.

src/datastore/pg-store-v2.ts
zone117x: Thanks for the verification ❤️

src/datastore/pg-store-v2.ts
zone117x: Perfect, thanks!

BLuEScioN (approved):

src/datastore/pg-store-v2.ts
zone117x: Ah got it, thanks for the head up

janniks (approved):
👏

PR #2012, opened 2024-06-18
feat: tx to/from address options

### LINKED ISSUE #1997: Add to/from address filter options to `/extended/v1/tx`
Add query params to the tx list endpoint that can filter based on the sender and/or recipient address. The sender filter can apply to any tx type (they all have a sender), however, the recipient filter probably only makes sense for `token_transfer` tx types.

Example usage: `/tx?from_address=SP2DRM5DKDJX6ECFJH9XDT66DZV8FK2H1A27JZHCA&to_address=SP11M3BVG5TGNWQ1FFE5EJM86TMV08YF3EK0JSMQ8`

* One or both can be specified.
* The usage of `to_address` will implicitly also filter for only `token_transfer` tx types.

He1DAr:
Closes #1997

### DIFF
--- a/src/api/routes/tx.ts
+++ b/src/api/routes/tx.ts
@@ -80,6 +80,28 @@ export function createTxRouter(db: PgStore): express.Router {
         }
       }
 
+      let fromAddress: string | undefined;
+      if (typeof req.query.from_address === 'string') {
+        if (!isValidC32Address(req.query.from_address)) {
+          throw new InvalidRequestError(
+            `Invalid query parameter for "from_address": "${req.query.from_address}" is not a valid STX address`,
+            InvalidRequestErrorType.invalid_param
+          );
+        }
+        fromAddress = req.query.from_address;
+      }
+
+      let toAddress: string | undefined;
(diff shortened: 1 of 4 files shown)


src/api/routes/tx.ts
```diff
@@ -80,6 +80,28 @@ export function createTxRouter(db: PgStore): express.Router {
...
+            `Invalid query parameter for "from_address": "${req.query.from_address}" is not a valid STX address`,
+            InvalidRequestErrorType.invalid_param
+          );
+        }
+        fromAddress = req.query.from_address;
+      }
+
+      let toAddress: string | undefined;
+      if (typeof req.query.to_address === 'string') {
+        if (!isValidC32Address(req.query.to_address)) {
```
zone117x, line 96: Recipients can be also be a contract, e.g. https://api.hiro.so/extended/v1/tx/0xcb13a41095cca7f264ab0800d8badf612c54b0e036908d8ebf2d38adc193d2f8

You can use `isValidPrincipal` here instead

zone117x (approved):
LGTM! Thanks @He1DAr!

Issue #1996, opened 2024-06-12
Add timestamp filter options to `/extended/v1/tx`


zone117x:
Add query params to the tx list endpoint that can filter based on start and/or end times. Should be based off the `burn_block_time`.

Example usage: `/tx?start_time=1704067200&end_time=1706745599` (one or both can be specified)

PR #2015, opened 2024-06-20
feat: tx list timestamp filter options

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1996

Add timestamp (by burn_block_time) filter options to `/extended/v1/tx`

Example usage: `/tx?start_time=1704067200&end_time=1706745599` (one or both can be specified)


zone117x:
@rafaelcr yep we have a burn_block_time index 👍

Issue #1999, opened 2024-06-12
Add function name filtering to `/extended/v1/tx`


zone117x:
Related issue: https://github.com/hirosystems/explorer/issues/964

Add query params to the tx list endpoint that filter by contract-call `function_name`. This would implicitly filter for only `contract_call` type transactions.

The query param could require the fully qualified contract id + function name -- something like:
`/tx?=function_name=SP000000000000000000002Q6VF78.pox-4.delegate-stx`

... or it could be broken up to allow filter by contract id _and/or_ function name, like
`/tx?contract_id=SP000000000000000000002Q6VF78.pox-4&function_name=delegate-stx`

The latter probably makes more sense because it's more flexible, and also I'm not aware of a convention for a "fully qualified function name". E.g. in Clarity these are broken up as separate arguments `(contract-call? 'SP000000000000000000002Q6VF78.pox-4 delegate-stx ...)`

PR #2018, opened 2024-06-20
feat: tx list contract id/name filter options

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1999

Add query params to the tx list endpoint that filter by contract-call `contract_id` and `function_name`.

Example usage: `/tx?contract_id=SP000000000000000000002Q6VF78.pox-4&function_name=delegate-stx`


rafaelcr (approved):
LGTM

PR #2033, opened 2024-07-17
fix: `/v2/addresses/{addr}/transactions` incorrect when address only involved with token events

zone117x:
Closes https://github.com/hirosystems/explorer/issues/1709

Fixes two bugs with the `/extended/v2/addresses/{addr}/transactions` endpoint:
* Missing results for addresses that only had sent/received ft/nft tokens and were not involved with any stx transfers.
* The associated tx fee was included in the `stx_sent` value even when the address was not the tx sender/sponsor.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1938,9 +1938,11 @@ paths:
     get:
       summary: Get account transactions
       description: |
-          Retrieves a list of all Transactions for a given Address or Contract Identifier. More information on Transaction types can be found [here](https://docs.stacks.co/understand-stacks/transactions#types).
+        **NOTE:** This endpoint is deprecated in favor of [Get address transactions](/api/get-address-transactions).
 
-          If you need to actively monitor new transactions for an address or contract id, we highly recommend subscribing to [WebSockets or Socket.io](https://github.com/hirosystems/stacks-blockchain-api/tree/master/client) for real-time updates.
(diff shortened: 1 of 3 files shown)


rafaelcr (approved):
Nice catch, thanks!

diwakergupta:
Are these `extended/v2/` endpoints correctly documented? cc/ @ryanwaits 

For instance, I hit `https://api.mainnet.hiro.so/extended/v1/address/SP29H5FH57AZVJSBWHCTJB5ATS2ZAH9SXN1XJDNK/transactions` and got a different output than `https://api.mainnet.hiro.so/extended/v2/addresses/SP29H5FH57AZVJSBWHCTJB5ATS2ZAH9SXN1XJDNK/transactions`

There are also subtle differences between the two variants (`address` vs. `addresses`) that are easy to get wrong.

As a user it's a bit confusing to understand which endpoints should one be using. For instance, if I go to https://docs.hiro.so/stacks/api/accounts/transactions I see that endpoints is deprecated but it doesn't tell me which endpoint I should use instead.

There are other discrepancies too, e.g. on https://docs.hiro.so/stacks/api/accounts/info the documentation has `extended/v1` endpoint but the code sample uses `extended/v2` 

Screenshot 2024-07-17 at 10 54 14 AM

ryanwaits:
+1 thanks for flagging, i noticed this when experimenting with the api playground

PR #2042, opened 2024-08-05
fix: batch insert length assertion

### LINKED ISSUE #2037: Code has an Assertion Error that causes the API to exit abnormally
**Describe the bug**
Code has an Assertion Error that causes the API to exit abnormally，the correct code should compare batch.length and res.count, but here it compares values.length and res.count.

https://github.com/hirosystems/stacks-blockchain-api/blob/develop/src/datastore/pg-write-store.ts#L1077
https://github.com/hirosystems/stacks-blockchain-api/blob/develop/src/datastore/pg-write-store.ts#L1122

**To Reproduce**
Code issues

**Expected behavior**
Fix the code to compare the incoming quantity with the actual inserted quantity to avoid abnormal exit of the API server.

**Console log**
```log
{"level":"error","time":"2024-07-31T03:31:51.266Z","pid":23,"hostname":"stx-1-0","name":"api","err":{"type":"Ass
ertionError","message":"Expecting 880 inserts, got 500","stack":"AssertionError [ERR_ASSERTION]: Expecting 880 i
nserts, got 500\n    at PgWriteStore.updateStxBalances (/app/src/datastore/pg-write-store.ts:1077:7)\n    at pro
cessTicksAndRejections (node:internal/process/task_queues:95:5)\n    at run (/app/node_modules/p-queue/dist/inde
x.js:163:29)","generatedMessage":false,"code":"ERR_ASSERTION","actual":false,"expected":true,"operator":"=="},"m
sg":"unhandledRejection"}
```

rafaelcr:
Fixes #2037

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -100,7 +100,7 @@
         "@typescript-eslint/eslint-plugin": "5.46.1",
         "@typescript-eslint/parser": "5.51.0",
         "concurrently": "7.3.0",
-        "docker-compose": "0.23.14",
+        "docker-compose": "0.24.8",
         "eslint": "8.29.0",
         "eslint-plugin-prettier": "4.2.1",
         "eslint-plugin-tsdoc": "0.2.17",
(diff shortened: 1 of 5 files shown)


zone117x (approved):
💯

Issue #2002, opened 2024-06-13
Isolate & organize Rosetta code


smcclellan:
With the new Express to Fastify migration in progress, there may be some special handling involved with Rosetta. This task is to improve folder & code organization around Rosetta support.

smcclellan:
Relates to https://github.com/hirosystems/stacks-blockchain-api/issues/1626.

Issue #1626, opened 2023-04-20
migrate from Express to Fastify


rafaelcr:
Adding this umbrella issue to track individual migrations:

* [ ] #1627
* [ ] #2002
* [x] #1628

smcclellan:
Large refactor that will pay off for p95s, logging, validation, ease of development, etc. Good investment.

This will also affect API documentation in a positive way, automating the OpenAPI spec.

PR #2045, opened 2024-08-08
Refactor/fastify

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1626
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2002
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1627

* Refactor all endpoints except for `/rosetta` from express to fastify
  * Rosetta endpoints, types, and schemas now largely isolated
* Refactor event-server to fastify
* Openapi & json schemas now defined in the actual route handlers
* Client library now generated with [openapi-ts](https://openapi-ts.dev/openapi-fetch/)
  * Websocket & socket.io usage remains unchanged
  * The `@stacks/stacks-blockchain-api-types` package is no longer needed or published now that client generation types are correct
  * See the [client migration notes](https://github.com/hirosystems/stacks-blockchain-api/blob/refactor/fastify-wip/client/MIGRATION.md) for more details

TODO:
* [x] Migration guide for client library
* [ ] Review openapi changes with docs team
* [ ] Canary testing for unintentional breaking changes


docs/openapi.json
rafaelcr, line 1: Do we need to check in this file to the repo? I think now that we'll have the neat generator we can afford putting this one on gitignore like the rest of the APIs

docs/openapi.yaml
rafaelcr, line 1: Same here, I think we can leave it out of the repo

src/api/routes/address.ts
```diff
@@ -76,32 +72,58 @@ async function getBlockHeight(
...
+      schema: {
+        operationId: 'get_account_stx_balance',
+        summary: 'Get account STX balance',
+        description: `Retrieves STX token balance for a given Address or Contract Identifier.`,
+        tags: ['Accounts'],
+        params: Type.Object({
+          principal: Type.String({
+            description: 'Stacks address or a Contract identifier',
+            examples: ['SP31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZW97B5P0'],
+          }),
```
rafaelcr, line 93: I think we can reuse a simple `AddressSchema` or similar in a bunch of these endpoints

src/api/routes/address.ts
```diff
@@ -76,32 +72,58 @@ async function getBlockHeight(
...
+        }),
+        querystring: Type.Object({
+          unanchored: UnanchoredParamSchema,
+          until_block: Type.Optional(
```
rafaelcr, line 97: This is also repeated a few times

src/api/routes/address.ts
```diff
@@ -277,27 +393,62 @@ export function createAddressRouter(db: PgStore, chainId: ChainID): express.Rout
...
   /**
    * @deprecated See `/v2/addresses/:address/transactions`
    */
-  router.get(
-    '/:stx_address/transactions_with_transfers',
-    cacheHandler,
-    asyncHandler(async (req, res, next) => {
-      const stxAddress = req.params['stx_address'];
+  fastify.get(
+    '/:principal/transactions_with_transfers',
```
rafaelcr, line 407: Question: should we take this opportunity to get rid of this one for good? (Not sure if this refactor will be a breaking change or not)

src/api/routes/bns/names.ts
```diff
@@ -147,26 +329,21 @@ export function createBnsNamesRouter(db: PgStore, chainId: ChainID): express.Rou
...
+            await reply.status(404).send(error);
           }
         });
-    })
+    }
   );
 
-  return router;
-}
+  await Promise.resolve();
```
rafaelcr, line 348: To avoid having to do this, you could declare all these route plugins as `FastifyPluginCallback` (instead of `Async`) and just call `done()` here at the end

src/api/routes/core-node-rpc-proxy.ts
```diff
@@ -166,12 +108,25 @@ export function createCoreNodeRpcProxyRouter(db: PgStore): express.Router {
     }
   }
 
-  router.post(
-    '/transactions',
-    asyncHandler(async (req, res, next) => {
+  const maxBodySize = 10_000_000; // 10 MB max POST body size
```
rafaelcr, line 111: I know we had this cap in the past but should we have this be configurable just in case?

src/api/routes/status.ts
```diff
@@ -1,46 +1,63 @@
...
       const response: ServerStatusResponse = {
         server_version: `stacks-blockchain-api ${SERVER_VERSION.tag} (${SERVER_VERSION.branch}:${SERVER_VERSION.commit})`,
         status: 'ready',
       };
-      const poxForceUnlockHeights = await db.getPoxForceUnlockHeights();
-      if (poxForceUnlockHeights.found) {
-        response.pox_v1_unlock_height = poxForceUnlockHeights.result.pox1UnlockHeight as number;
-        response.pox_v2_unlock_height = poxForceUnlockHeights.result.pox2UnlockHeight as number;
-        response.pox_v3_unlock_height = poxForceUnlockHeights.result.pox3UnlockHeight as number;
+      try {
```
rafaelcr, line 30: We should wrap this in a pg transaction

src/api/routes/bns/names.ts
zone117x: I'm a fan of the async one because the type checker / eslint enforces the correct behavior. With `FastifyPluginCallback`, the `done()` is easily omitted and then there's just a mysterious timeout at runtime.

src/api/routes/core-node-rpc-proxy.ts
zone117x: IIRC this is pretty much only here to override the default cap (1MB I think). But yeah probably still useful to make configurable. Maybe a good thing to do in a PR where we migrate to `env-schema` (I left that refactor out of this PR because of the already huge change set)

rafaelcr: Sounds good to me

src/event-stream/event-server.ts
```diff
@@ -931,174 +928,145 @@ export async function startEventServer(opts: {
     eventHost = hostname;
   }
 
-  const app = express();
+  const bodyLimit = 1_000_000 * 500; // 500MB body limit
```
rafaelcr, line 931: Another value we should set as configurable in the next PR

rafaelcr (approved):
Other than the nits I've commented on, this LGTM 👍🏻 Can't wait to try it out

src/api/routes/address.ts
zone117x: Good catch

src/api/routes/status.ts
zone117x: Good catch

src/api/routes/core-node-rpc-proxy.ts
zone117x: https://github.com/hirosystems/stacks-blockchain-api/issues/2055

src/event-stream/event-server.ts
zone117x: https://github.com/hirosystems/stacks-blockchain-api/issues/2055

PR #2048, opened 2024-08-12
fix: export events tsv directly to postgres instance

rafaelcr:
This PR adds a `local` option to the `export-events` API CLI command. If the argument to `--file` starts with `local:`, it will export events using only postgres to a file path local to its container.


src/event-replay/event-requests.ts
```diff
@@ -1,30 +1,27 @@
...
-    client.release();
-    await pool.end();
-  }
+  await sql`
+    COPY (
+      SELECT id, receive_timestamp, event_path, payload
+      FROM event_observer_requests
+      ORDER BY id ASC
+    )
+    TO '${sql.unsafe(filePath)}' WITH (FORMAT TEXT, DELIMITER E'\t', ENCODING 'UTF8')
```
rafaelcr, line 22: @zone117x thoughts here? Postgres can't take positional arguments for this path

src/event-replay/event-replay.ts
```diff
@@ -31,23 +31,16 @@ enum EventImportMode {
  * @param filePath - Path to TSV file to write
  * @param overwriteFile - If we should overwrite the file
```
CharlieC3, line 32: This line should be removed since the `overwriteFile` option was removed.

src/index.ts
```diff
@@ -274,7 +274,7 @@ function getProgramArgs() {
 async function handleProgramArgs() {
   const { args, parsedOpts } = getProgramArgs();
   if (args.operand === 'export-events') {
-    await exportEventsAsTsv(args.options.file, args.options['overwrite-file']);
```
CharlieC3, line 277: Does this mean we'll no longer be able to overwrite a file? Or will it overwrite by default?

zone117x (commented):
Can we make the local export an opt-in feature? I made a modification (https://github.com/hirosystems/stacks-blockchain-api/commit/2f9e591467ec8cb6fadab10d1057c4981c059283) and just tested against a kubectl port-forward to a mainnet dev instance. It exported a ~50GB tsv to my machine, and I validated that every line is valid json.

rafaelcr:
that looks great @zone117x , would you mind applying that patch to this branch?

src/index.ts
rafaelcr: Postgres overwrites the given file by default

src/event-replay/event-requests.ts
zone117x: Seems fine to me

zone117x:
My guess is that the CI failure is because host docker filesystem usage causing the `./tmp/events.tsv` to be owned by a different user (the postgres image system user) and so the CI system user can't delete it

rafaelcr:
@CharlieC3 do you have a suggestion on what I can do to fix this failure? https://github.com/hirosystems/stacks-blockchain-api/actions/runs/10392592660/job/28778191413#step:7:655

I think @zone117x is right that the user difference is making the test fail here. Perhaps it won't be possible to test this particular thing on gh actions?

CharlieC3:
@rafaelcr The postgres image runs the database as the postgres user, so you're both right it is a user mismatch.
```
root@4208a0d55448:/# id postgres
uid=999(postgres) gid=999(postgres) groups=999(postgres),101(ssl-cert)
```

You may have to `chmod` and `chown` the host directory you're mounting to user `999` before starting the container

```
chown 999:999 ../src/tests-event-replay/.tmp/local/
chmod -R 777 ../src/tests-event-replay/.tmp/local/
```

rafaelcr:
That did it @CharlieC3 thanks!

zone117x (approved):

PR #2058, opened 2024-08-20
feat: export events tsv directly to postgres instance (#2048)

rafaelcr:
Releases #2048 as a cherry pick to master


zone117x (approved):
LGTM if tests pass

Issue #2043, opened 2024-08-05
Add cursor-based pagination on `/blocks` endpoint


smcclellan:
To improve the speed of requests to the `/blocks` endpoint, we can use cursor-based pagination using `height` as the cursor. This will take the place of `offset` in the endpoint currently.

PR #2060, opened 2024-08-23
feat: cursor-based pagination on blocks endpoint

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2043

Supports a new `cursor` query param. The `offset` param continues to work but with a max page limit. Page random-access is possible when using both together.

This approach is generalizable and should work with other listing endpoints like transactions in future PRs.

Example usage:

`GET /extended/v2/blocks?limit=3`
```js
{
  limit: 3,
  offset: 0,
  total: 14,
  cursor: '0x0000000000000000000000000000000000000000000000000000000000000014',
  next_cursor: null,
  prev_cursor: '0x0000000000000000000000000000000000000000000000000000000000000011',
  results: [
   { height: 14, ... },
   { height: 13, ... },
   { height: 12, ... }
  ],
}
```

`GET /extended/v2/blocks?limit=3&cursor=0x0000000000000000000000000000000000000000000000000000000000000011`
```js
{
  limit: 3,
  offset: 0,
  total: 14,
  cursor: '0x0000000000000000000000000000000000000000000000000000000000000011',
  next_cursor: '0x0000000000000000000000000000000000000000000000000000000000000014',
  prev_cursor: '0x0000000000000000000000000000000000000000000000000000000000000008',
  results: [
   { height: 11, ... },
   { height: 10, ... },
   { height: 9, ... }
  ],
}
```

`GET /extended/v2/blocks?limit=3&cursor=0x0000000000000000000000000000000000000000000000000000000000000011&offset=2`
```js
{
  limit: 3,
  offset: 0,
  total: 14,
  cursor: '0x0000000000000000000000000000000000000000000000000000000000000009',
  next_cursor: '0x0000000000000000000000000000000000000000000000000000000000000012',
  prev_cursor: '0x0000000000000000000000000000000000000000000000000000000000000006',
  results: [
   { height: 9, ... },
   { height: 8, ... },
   { height: 7, ... }
  ],
}
```

`GET /extended/v2/blocks?limit=3&cursor=0x0000000000000000000000000000000000000000000000000000000000000008&offset=-2`
```js
{
  limit: 3,
  offset: 0,
  total: 14,
  cursor: '0x0000000000000000000000000000000000000000000000000000000000000010',
  next_cursor: '0x0000000000000000000000000000000000000000000000000000000000000013',
  prev_cursor: '0x0000000000000000000000000000000000000000000000000000000000000007',
  results: [
   { height: 10, ... },
   { height: 9, ... },
   { height: 8, ... }
  ],
}
```

### DIFF
--- a/src/api/routes/v2/blocks.ts
+++ b/src/api/routes/v2/blocks.ts
@@ -7,10 +7,11 @@ import { FastifyPluginAsync } from 'fastify';
 import { Type, TypeBoxTypeProvider } from '@fastify/type-provider-typebox';
 import { Server } from 'node:http';
 import { LimitParam, OffsetParam } from '../../schemas/params';
-import { ResourceType } from '../../pagination';
+import { getPagingQueryLimit, ResourceType } from '../../pagination';
 import { PaginatedResponse } from '../../schemas/util';
 import { NakamotoBlock, NakamotoBlockSchema } from '../../schemas/entities/block';
 import { TransactionSchema } from '../../schemas/entities/transactions';
+import { BlockListV2ResponseSchema } from '../../schemas/responses/responses';
 
 export const BlockRoutesV2: FastifyPluginAsync<
   Record<never, never>,
@@ -29,20 +30,25 @@ export const BlockRoutesV2: FastifyPluginAsync<
         querystring: Type.Object({
           limit: LimitParam(ResourceType.Block),
           offset: OffsetParam(),
+          cursor: Type.Optional(Type.String({ description: 'Cursor for pagination' })),
         }),
         response: {
-          200: PaginatedResponse(NakamotoBlockSchema),
+          200: BlockListV2ResponseSchema,
         },
       },
     },
     async (req, reply) => {
       const query = req.query;
-      const { limit, offset, results, total } = await fastify.db.v2.getBlocks(query);
-      const blocks: NakamotoBlock[] = results.map(r => parseDbNakamotoBlock(r));
+      const limit = getPagingQueryLimit(ResourceType.Block, req.query.limit);
+      const blockQuery = await fastify.db.v2.getBlocks({ ...query, limit });
+      const blocks: NakamotoBlock[] = blockQuery.results.map(r => parseDbNakamotoBlock(r));
       await reply.send({
-        limit,
-        offset,
-        total,
+        limit: blockQuery.limit,
+        offset: blockQuery.offset,
+        total: blockQuery.total,
+        next_cursor: blockQuery.next_cursor,
+        prev_cursor: blockQuery.prev_cursor,
+        cursor: blockQuery.current_cursor,
         results: blocks,
       });
     }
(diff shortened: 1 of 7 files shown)


src/api/schemas/responses/responses.ts
```diff
@@ -178,3 +179,14 @@ export const RunFaucetResponseSchema = Type.Object(
   }
 );
 export type RunFaucetResponse = Static<typeof RunFaucetResponseSchema>;
+
+export const BlockListV2ResponseSchema = Type.Object({
```
rafaelcr, line 183: We should generalize this structure similar to what we did in `PaginatedResponse()` so we have something reusable like `PaginatedCursorResponse()`

src/tests/block-tests.ts
```diff
@@ -754,6 +755,139 @@ describe('block tests', () => {
     expect(fetch.status).not.toBe(200);
   });
 
+  test('blocks v2 cursor pagination', async () => {
```
rafaelcr, line 758: What would happen if there's a reorg in this case? Would a previously valid cursor return a 404? Perhaps a test for this is a good idea

src/api/routes/v2/blocks.ts
```diff
@@ -28,21 +29,34 @@ export const BlockRoutesV2: FastifyPluginAsync<
         tags: ['Blocks'],
         querystring: Type.Object({
           limit: LimitParam(ResourceType.Block),
-          offset: OffsetParam(),
+          offset: Type.Optional(
+            Type.Integer({
+              default: 0,
+              maximum: pagingQueryLimits[ResourceType.Block].maxLimit * 10, // Random access up to 10 pages
+              minimum: -pagingQueryLimits[ResourceType.Block].maxLimit * 10,
```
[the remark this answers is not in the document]
zone117x: Good idea, done!

src/api/schemas/responses/responses.ts
zone117x: Done!

src/datastore/pg-store-v2.ts
```diff
@@ -59,37 +60,101 @@ async function assertTxIdExists(sql: PgSqlClient, tx_id: string) {
...
+          FROM blocks
+          WHERE canonical = true
+          ORDER BY block_height DESC
         )
-        SELECT
-          ${sql(BLOCK_COLUMNS)},
-          (SELECT count FROM block_count)::int AS total
+        SELECT offset_block_height as block_height
+        FROM ordered_blocks 
+        WHERE block_hash = ${cursor}
```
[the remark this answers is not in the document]
zone117x: Agreed, done.

src/tests/block-tests.ts
zone117x: We _could_ still resolve the page in this case by allowing `canonical=false` in the cursor lookup, but then we'd have weird behavior like the result cursor being different (the canonical hash) than the given one. We are also not guaranteed to support non-canonical cursors in future changes. 

So yeah I think 404 is good idea. I'll add test.

zone117x: Done

PR #2073, opened 2024-09-06
fix: use total_count CTE and return it with the parsed results

### LINKED ISSUE #2061: total transactions reported as 0 when an offset is used
Example: https://api.hiro.so/extended/v2/addresses/SP1JX8274K2R0MMP1VCX56C5DCERPP6EH5XHS68MR/transactions?limit=30&offset=30

```
{
  "limit": 30,
  "offset": 30,
  "total": 0,
  "results": [
    
  ]
}
```
When `offset=` is removed:
```
{
  "limit": 30,
  "offset": 0,
  "total": 30,
  "results": [
    {
      "tx": {
```

tippenein:
- closes #2061 

the `count` column was previously counting the total of results returned, rather than results in the total set. This will compute the total separately from the offset and limit params

### DIFF
--- a/src/datastore/common.ts
+++ b/src/datastore/common.ts
@@ -251,7 +251,6 @@ export interface DbTxWithAddressTransfers extends DbTx {
   nft_transfer: number;
   nft_mint: number;
   nft_burn: number;
-  total_count: number;
 }
 
 export interface DbTxGlobalStatus {
@@ -1041,7 +1040,6 @@ export interface AddressTransfersTxQueryResult extends TxQueryResult {
   nft_transfer: number;
   nft_mint: number;
   nft_burn: number;
-  total_count: number;
 }
 
 export interface DbAddressTransactionEvent {
--- a/src/datastore/helpers.ts
+++ b/src/datastore/helpers.ts
@@ -343,7 +343,6 @@ export function parseAccountTransferSummaryTxQueryResult(
     nft_transfer: result.nft_transfer,
     nft_mint: result.nft_mint,
     nft_burn: result.nft_burn,
-    total_count: result.total_count,
   };
 }
 
--- a/src/datastore/pg-store-v2.ts
+++ b/src/datastore/pg-store-v2.ts
@@ -518,8 +518,8 @@ export class PgStoreV2 extends BasePgStoreModule {
             FROM nft_events
             WHERE sender = ${args.address} OR recipient = ${args.address}
           )
-        )
-        total_count AS (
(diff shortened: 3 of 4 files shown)


src/datastore/common.ts
```diff
@@ -1040,6 +1041,7 @@ export interface AddressTransfersTxQueryResult extends TxQueryResult {
   nft_transfer: number;
   nft_mint: number;
   nft_burn: number;
+  total_count: number;
```
rafaelcr, line 1044: These new fields are not necessary to add in types

src/datastore/helpers.ts
```diff
@@ -343,6 +343,7 @@ export function parseAccountTransferSummaryTxQueryResult(
     nft_transfer: result.nft_transfer,
     nft_mint: result.nft_mint,
     nft_burn: result.nft_burn,
+    total_count: result.total_count,
```
rafaelcr, line 346: Also not necessary

src/datastore/pg-store-v2.ts
```diff
@@ -572,15 +578,15 @@ export class PgStoreV2 extends BasePgStoreModule {
             SELECT COUNT(*)::int FROM nft_events
             WHERE ${eventAcctCond} AND asset_event_type_id = ${DbAssetEventTypeId.Burn}
           ) AS nft_burn,
-          (COUNT(*) OVER())::int AS count
+          (SELECT total_count FROM total_count) AS total_count
```
rafaelcr, line 581: Use `count` for the column name as that's the one declared in `AddressTransfersTxQueryResult & { count: number }`

src/datastore/pg-store-v2.ts
```diff
@@ -518,6 +518,12 @@ export class PgStoreV2 extends BasePgStoreModule {
             FROM nft_events
             WHERE sender = ${args.address} OR recipient = ${args.address}
           )
+        ),
+        total_count AS (
+          SELECT COUNT(*)::int AS total_count
+          FROM address_txs
+          INNER JOIN txs USING (tx_id, index_block_hash, microblock_hash)
+          WHERE canonical = TRUE AND microblock_canonical = TRUE
```
rafaelcr, line 526: Instead of this join, add the `WHERE canonical = TRUE AND microblock_canonical = TRUE` filters to the union tables above

rafaelcr (changes requested):

rafaelcr (approved):

PR #2086, opened 2024-09-24
test: refactor and optimize test suites

rafaelcr:
* Relocate all test suites so they live inside `tests/`
* Fix TS config so it covers all test files correctly
* Rename all test files to `*.test.ts` so it's easier to match them in Jest configs and so they appear as test files in VScode
* Integrate test suite runs into matrix groups in GH actions
* Simplify test scripts in `package.json`
* Remove upload to Coveralls
* Upgrade all GH actions


zone117x (approved):
Very nice clean up!

PR #2091, opened 2024-09-25
fix: prune and restore mempool transactions with equal nonces for the same sender

### LINKED ISSUE #2041: automatically mark mempool transactions with conflicting nonces as failed
The API returns a bunch of mempool transactions that are impossible to mine given that they conflict with other **confirmed** transactions from the same senders with the same nonce, so they end up being stuck for days (until they get pruned after 256 blocks). For example, this transaction with nonce `4461` is still pending:

Screenshot 2024-08-02 at 22 33 06

However, that sender STX address already has a confirmed transaction with the same nonce (in this case a coinbase, but the same happens with any tx type):

Screenshot 2024-08-02 at 22 34 20

These stuck transactions can cause user confusion but they most definitely affect the fee estimation algorithms. (I'm not sure why the Stacks node does not drop them automatically.)

* We should add logic to the API that upon a tx confirmation, it checks the mempool table for txs from that sender with that nonce and marks them as failed or pruned.

rafaelcr:
This PR alters how transactions for the **same sender with the same nonce** are pruned from and restored to the mempool.

* If transaction A with nonce 1 exists in the mempool and a new block comes which confirms transaction B with nonce 1 as well, transaction A is pruned.
* If there is a future re-org that restores transaction B to the mempool, transaction A gets restored as well so they can both compete in the mempool similar to an RBF situation.

Fixes #2041

### DIFF
--- a/src/datastore/pg-write-store.ts
+++ b/src/datastore/pg-write-store.ts
@@ -1,11 +1,5 @@
 import * as assert from 'assert';
-import {
-  getOrAdd,
-  I32_MAX,
-  getIbdBlockHeight,
-  getUintEnvOrDefault,
-  unwrapOptionalProp,
-} from '../helpers';
+import { getOrAdd, I32_MAX, getIbdBlockHeight, getUintEnvOrDefault } from '../helpers';
 import {
   DbBlock,
   DbTx,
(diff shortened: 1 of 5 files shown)


zone117x (approved):
Nice fix 💯

Issue #1230, opened 2022-07-07
add cache handlers for individual items (blocks, microblocks, principals)


rafaelcr:
We currently have two cache handlers that we use to create ETags for most of our endpoints:

1. **Chain tip**: Latest block/microblock hash
2. **Mempool**: Hash digest of all non-pruned mempool `tx_id`s

While these work great for endpoints which depend on global chain or mempool status like `/extended/v1/blocks` or `/extended/v1/tx/mempool`, they can incorrectly return a cache miss for others that are concerned with individual items. Some examples:

* `/extended/v1/blocks/:block_hash` would be a cache miss when the chain advances, even though the individual block would not have changed at all
* `/extended/v1/tx/:tx_id` similar (but already being fixed in https://github.com/hirosystems/stacks-blockchain-api/pull/1229)
* `/extended/v1/address/:principal/transactions` would be a cache miss even though that principal did not receive or send any new transactions
* `/extended/v1/address/:principal/balances` same as above
* etc...

We should create an individual item cache handlers for each of these (and more) use cases so we can maximize our number of cached responses

smcclellan:
@rafaelcr Can this be closed? The cached/uncached percentage has been improving the last several months, at ~75% now.

rafaelcr:
@smcclellan we should probably icebox this one. it's a good improvement that we should make at some point.

PR #2097, opened 2024-09-26
feat: add principal cache etag to account endpoints

rafaelcr:
Creates an ETag for endpoints that serve confirmed data for a single principal (address or contract ID) and adds it to all single-account endpoints.

The ETag is formed with the SHA256 of a string containing the last transaction IDs of those that changed the account's STX, FT or NFT balance:
```ts
sha256(`${activity.stx_tx_id}:${activity.ft_tx_id}:${activity.nft_tx_id}`)
```
---
Other changes:
* Remove code that was moved to the API toolkit
* Consider `sponsor_address` in entries to the `principal_stx_txs` table, including a migration to backfill missing transactions

A future PR will add a similar ETag but also considering that account's mempool activity.

#1230

### DIFF
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -4404,4 +4404,44 @@ export class PgStore extends BasePgStore {
     }
     return result;
   }
+
+  /** Retrieves the last transaction IDs with STX, FT and NFT activity for a principal */
+  async getPrincipalLastActivityTxIds(
+    principal: string
+  ): Promise<{ stx_tx_id: string | null; ft_tx_id: string | null; nft_tx_id: string | null }> {
+    const result = await this.sql<
+      { stx_tx_id: string | null; ft_tx_id: string | null; nft_tx_id: string | null }[]
+    >`
+      WITH last_stx AS (
+        SELECT tx_id
+        FROM principal_stx_txs
+        WHERE principal = ${principal} AND canonical = true AND microblock_canonical = true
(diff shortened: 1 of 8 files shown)


src/datastore/pg-store.ts
```diff
@@ -4404,4 +4404,44 @@ export class PgStore extends BasePgStore {
...
+  /** Retrieves the last transaction IDs with STX, FT and NFT activity for a principal */
+  async getPrincipalLastActivityTxIds(
+    principal: string
+  ): Promise<{ stx_tx_id: string | null; ft_tx_id: string | null; nft_tx_id: string | null }> {
+    const result = await this.sql<
+      { stx_tx_id: string | null; ft_tx_id: string | null; nft_tx_id: string | null }[]
+    >`
+      WITH last_stx AS (
+        SELECT tx_id
+        FROM principal_stx_txs
```
zone117x, line 4417: I think this might not be considering `sponsor_address`. We'd need to update the `principal_stx_txs` table update logic for that. Thoughts?

rafaelcr: Good catch. Does being a sponsor generate an `stx_event`? If so I think we're covered. Otherwise we do need to patch the `principal_stx_txs` table.

zone117x: > Does being a sponsor generate an stx_event?

I don't think so

rafaelcr:
@zone117x I've added the fix for `sponsor_address`, can you take another look?

zone117x (approved):
LGTM!

Issue #2035, opened 2024-07-23
Add property for estimated balance to balances endpoint


markmhendrickson:
As discussed [here](https://github.com/hirosystems/stacks-blockchain-api/issues/2026), Leather would benefit from an `estimated_balance` property in addition to the `balance` as returned by the `balances` endpoint for a given Stacks address.

This estimated balance would subtract the total of all outbound transfers from its confirmed balance for a given address.

Leather currently calculates such a balance through two separate queries and a final calculation:

1. fetch balance from `api.hiro.so/extended/v1/address/<address>/balances`
2. fetch pending txs from `api.hiro.so/extended/v1/tx/mempool?address=<address>&limit=50`
3. do calculation: `availableBalance = balance - sum of all outbound transaction amounts`

smcclellan:
Thank you for filing this, @markmhendrickson! I'd like us to separate this into two tickets:
1. The first will capture this new available balance estimation, and add it to a new endpoint.
2. The second will add this new field to the existing endpoint.
The primary reason for this is that the first item is an easy green light, without complication. We can move ahead with the implementation there. The second item requires quite a bit more consideration and maintenance burden around performance and pagination, and therefore would move slower through our development process.

Would implementing the first item, the new endpoint, in the near term be helpful on your end?

markmhendrickson:
Thanks @smcclellan, I suppose step 1 by itself may give us a minor benefit of simplifying our code. However, step 2 would have the most impact by helping us cut down on the number of API queries we generate substantially.

rafaelcr:
@markmhendrickson we are planning on adding the `estimated_balance` key to the balances response, however, we first need to adjust the Etag generation logic a bit to make it be account specific instead of chain tip block height specific (as it is today) so we don't run into any caching issues when wallet users call this endpoint. We'll keep you updated on the work here.

wmnonstx:
@rafaelcr what adjustments are made to "estimated_balance" compared to available balance. 

In Leather we've been accounting for outgoing STX, all pending STX fees for example. 

On the wishlist were any STX transfers for contract calls (which will not always be obvious) but could be deduced from the post-conditions. They would have to be very explicit "transfer exactly xxx STX" adjusting based on "transfer no more than..." or "no less then..." could lead to to much or to small an adjustment.

314159265359879:
@rafaelcr what adjustments are made to "estimated_balance" compared to available balance.

[...]

PR #2100, opened 2024-09-30
feat: add cache handler for principal activity including mempool transactions

rafaelcr:
Expands on #2097 to add a new ETag that supports principal activity currently in the mempool. This will enable endpoints for feature requests like #2035 

This ETag is now used in:
```
/extended/v1/address/:principal/mempool
/extended/v1/address/:principal/nonces
```

#1230

### DIFF
--- a/src/api/controllers/cache-controller.ts
+++ b/src/api/controllers/cache-controller.ts
@@ -127,15 +127,12 @@ async function calculateETag(
         const params = req.params as { address?: string; principal?: string };
         const principal = params.address ?? params.principal;
         if (!principal) return ETAG_EMPTY;
-        return await db.sqlTransaction(async sql => {
-          const activity = await db.getPrincipalLastActivityTxIds(sql, principal);
-          let text = `${activity.stx_tx_id}:${activity.ft_tx_id}:${activity.nft_tx_id}`;
-          if (etagType == ETagType.principalMempool) {
-            const mempoolTxs = await db.getPrincipalMempoolTxIds(sql, principal);
-            if (mempoolTxs.length) text = `${text}:${mempoolTxs.join(':')}`;
-          }
-          return sha256(text);
-        });
+        const activity = await db.getPrincipalLastActivityTxIds(
+          principal,
+          etagType == ETagType.principalMempool
+        );
+        if (!activity.length) return ETAG_EMPTY;
+        return sha256(activity.join(':'));
     }
   } catch (error) {
     logger.error(error, `Unable to calculate ${etagType} etag`);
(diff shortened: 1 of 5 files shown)


src/datastore/pg-store.ts
```diff
@@ -4407,41 +4407,53 @@ export class PgStore extends BasePgStore {
...
+        )
+        ${
+          includeMempool
+            ? this.sql`UNION
+            (
+              SELECT tx_id
+              FROM mempool_txs
+              WHERE sender_address = ${principal}
+                OR sponsor_address = ${principal}
+                OR token_transfer_recipient_address = ${principal}
```
zone117x, line 4450: Should this have a `LIMIT 1`?

rafaelcr: Since I can't sort the transactions by block I need to take all of them to get a complete picture of that account's mempool activity. Unless I sort by receipt time and the only the most recent one? What do you think?

zone117x: Hmm I think order and limit 1 by receipt time is probably better than grabbing all of them. If we do have to grab all of them, we should ignore `pruned=true` right? Otherwise this can return a huge amount of rows

rafaelcr: Oh right, good catch. I'll do that

src/datastore/pg-store.ts
```diff
@@ -4407,41 +4407,56 @@ export class PgStore extends BasePgStore {
...
+          includeMempool
+            ? this.sql`UNION
+            (
+              SELECT tx_id
+              FROM mempool_txs
+              WHERE pruned = false AND
+                (sender_address = ${principal}
+                OR sponsor_address = ${principal}
+                OR token_transfer_recipient_address = ${principal})
+              ORDER BY receipt_time DESC, sender_address DESC, nonce DESC
```
rafaelcr, line 4452: @zone117x I added this sorting just in case affecting transactions are received inside the same `/new_mempool_tx` event

src/api/controllers/cache-controller.ts
```diff
@@ -129,23 +121,21 @@ async function calculateETag(
...
-      }
+        const activity = await db.getPrincipalLastActivityTxIds(
+          principal,
+          etagType == ETagType.principalMempool
+        );
+        if (!activity.length) return ETAG_EMPTY;
+        return sha256(activity.join(':'));
+    }
+  } catch (error) {
+    logger.error(error, `Unable to calculate ${etagType} etag`);
```
zone117x, line 138: Did you encounter errors here? I'm curious what they were

zone117x (approved):
LGTM!

Issue #2099, opened 2024-09-30
Get user balances at historic block not working


radicleart:
**Describe the bug**

The api call to [get user balances](https://docs.hiro.so/stacks/api/accounts/balances) at block height is broken.

**To Reproduce**

1. Go to 'https://api.hiro.so/extended/v1/address/SP3N4AJFZZYC4BK99H53XP8KDGXFGQ2PRSQP2HGT6/balances?until_block=160000'
2. See error
```
"statusCode": 400,
"error": "Bad Request",
"message": "can't handle both 'unanchored' and 'until_block' in the same request"
```

I tried explicitly setting unanchored but get same response; https://api.hiro.so/extended/v1/address/SP3N4AJFZZYC4BK99H53XP8KDGXFGQ2PRSQP2HGT6/balances?until_block=160000&unanchored=false

**Expected behavior**
User balances returned

zone117x:
Good catch! Aiming to get a fix in for next release.

radicleart:
> Good catch! Aiming to get a fix in for next release.

How often are releases @zone117x? This is for the sBTC community vote which is due to go live on the 14th - is before then ?

zone117x:
The fix for this issue has been deployed.

PR #2101, opened 2024-10-01
fix: query param `until_block` not working in several endpoints

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2099

The mutually-exclusive query param check for `unanchored` and `until_block` check had a regression from the Fastify refactor when `unanchored` was changed to default to `false`. This hotfix checks if the value is set to `true` rather than `!== undefined`.


rafaelcr (approved):
LGTM

PR #2102, opened 2024-10-01
feat: tx list function name fuzzy search

### LINKED ISSUE #2164: Investigate fuzzy search query performance discrepancy
There’s a significant difference in performance when running the same query locally vs. in the dev environment, even though the query uses the same execution strategy in both environments.

Query:

```
SELECT 
	  *
      FROM stacks_blockchain_api.txs
      WHERE canonical = true AND microblock_canonical = true AND block_height <= 157521 AND burn_block_time <= 1730591999
	  AND (
        CASE 
          WHEN EXISTS (
            SELECT 1
            FROM pg_extension
            WHERE extname = 'pg_trgm'
          ) 
          THEN similarity(contract_call_function_name, 'swab') > 0.3
          ELSE contract_call_function_name ILIKE '%swab%'
          END
      )
      ORDER BY burn_block_time DESC, block_height DESC, microblock_sequence DESC, tx_index DESC
      LIMIT 30
      OFFSET 0;
```

Analysis output:

```
[
  {
    "Plan": {
      "Node Type": "Limit",
      "Parallel Aware": false,
      "Async Capable": false,
      "Startup Cost": 821.56,
      "Total Cost": 887.96,
      "Plan Rows": 30,
      "Plan Width": 2475,
      "Actual Startup Time": 730.829,
      "Actual Total Time": 730.837,
      "Actual Rows": 30,
      "Actual Loops": 1,
      "Plans": [
        {
          "Node Type": "Seq Scan",
          "Parent Relationship": "InitPlan",
          "Subplan Name": "InitPlan 1 (returns $0)",
          "Parallel Aware": false,
          "Async Capable": false,
          "Relation Name": "pg_extension",
          "Alias": "pg_extension",
          "Startup Cost": 0.00,
          "Total Cost": 1.01,
          "Plan Rows": 1,
          "Plan Width": 0,
          "Actual Startup Time": 0.019,
          "Actual Total Time": 0.020,
          "Actual Rows": 1,
          "Actual Loops": 1,
          "Filter": "(extname = 'pg_trgm'::name)",
          "Rows Removed by Filter": 1
        },
        {
          "Node Type": "Incremental Sort",
          "Parent Relationship": "Outer",
          "Parallel Aware": false,
          "Async Capable": false,
          "Startup Cost": 820.54,
          "Total Cost": 13030750.71,
          "Plan Rows": 5886325,
          "Plan Width": 2475,
          "Actual Startup Time": 730.827,
          "Actual Total Time": 730.831,
          "Actual Rows": 30,
          "Actual Loops": 1,
          "Sort Key": ["txs.burn_block_time DESC", "txs.block_height DESC", "txs.microblock_sequence DESC", "txs.tx_index DESC"],
          "Presorted Key": ["txs.burn_block_time"],
          "Full-sort Groups": {
            "Group Count": 1,
            "Sort Methods Used": ["quicksort"],
            "Sort Space Memory": {
              "Average Sort Space Used": 85,
              "Peak Sort Space Used": 85
            }
          },
          "Plans": [
            {
              "Node Type": "Index Scan",
              "Parent Relationship": "Outer",
              "Parallel Aware": false,
              "Async Capable": false,
              "Scan Direction": "Backward",
              "Index Name": "txs_burn_block_time_index",
              "Relation Name": "txs",
              "Alias": "txs",
              "Startup Cost": 0.43,
              "Total Cost": 12688723.86,
              "Plan Rows": 5886325,
              "Plan Width": 2475,
              "Actual Startup Time": 717.984,
              "Actual Total Time": 730.722,
              "Actual Rows": 31,
              "Actual Loops": 1,
              "Index Cond": "(burn_block_time <= 1730591999)",
              "Rows Removed by Index Recheck": 0,
              "Filter": "(canonical AND microblock_canonical AND (block_height <= 157521) AND CASE WHEN $0 THEN (similarity(contract_call_function_name, 'swab'::text) > '0.3'::double precision) ELSE (contract_call_function_name ~~* '%swab%'::text) END)",
              "Rows Removed by Filter": 981643
            }
          ]
        }
      ]
    },
    "Planning Time": 0.363,
    "Triggers": [
    ],
    "Execution Time": 730.896
  }
]
```

He1DAr:
Implemented fuzzy search for transaction list query using the `pg_trgm` extension, starting with indexing the `contract_call_function_name` column. Future plan is to expand the index to cover more columns, improving search flexibility, e.g.:
```
CREATE INDEX idx_search_term_trgm
ON public.txs
USING gin (
  contract_call_function_name gin_trgm_ops,
  contract_name gin_trgm_ops
  memo gin_trgm_ops
  ....
);
```

### DIFF
--- a/src/api/routes/tx.ts
+++ b/src/api/routes/tx.ts
@@ -113,6 +113,12 @@ export const TxRoutes: FastifyPluginAsync<
               examples: [1706745599],
             })
           ),
+          search_term: Type.Optional(
+            Type.String({
+              description: 'Option to search for transactions by a search term',
+              examples: ['swap'],
+            })
+          ),
           contract_id: Type.Optional(
             Type.String({
               description: 'Option to filter results by contract ID',
@@ -178,6 +184,11 @@ export const TxRoutes: FastifyPluginAsync<
         contractId = req.query.contract_id;
       }
 
+      let searchTerm: string | undefined;
+      if (typeof req.query.search_term === 'string') {
+        searchTerm = req.query.search_term;
+      }
+
       const { results: txResults, total } = await fastify.db.getTxList({
         offset,
         limit,
@@ -188,6 +199,7 @@ export const TxRoutes: FastifyPluginAsync<
         startTime: req.query.start_time,
         endTime: req.query.end_time,
         contractId,
+        searchTerm,
         functionName: req.query.function_name,
         nonce: req.query.nonce,
         order: req.query.order,
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -101,6 +101,8 @@ import { parseBlockParam } from '../api/routes/v2/schemas';
 
 export const MIGRATIONS_DIR = path.join(REPO_DIR, 'migrations');
 
+const TRGM_SIMILARITY_THRESHOLD = 0.3;
+
 /**
  * This is the main interface between the API and the Postgres database. It contains all methods that
  * query the DB in search for blockchain data to be returned via endpoints or WebSockets/Socket.IO.
@@ -1416,6 +1418,7 @@ export class PgStore extends BasePgStore {
     startTime,
     endTime,
     contractId,
+    searchTerm,
     functionName,
     nonce,
     order,
@@ -1430,6 +1433,7 @@ export class PgStore extends BasePgStore {
     startTime?: number;
     endTime?: number;
     contractId?: string;
+    searchTerm?: string;
     functionName?: string;
     nonce?: number;
     order?: 'desc' | 'asc';
@@ -1468,6 +1472,23 @@ export class PgStore extends BasePgStore {
       const contractIdFilterSql = contractId
         ? sql`AND contract_call_contract_id = ${contractId}`
         : sql``;
+
+      const searchTermFilterSql = searchTerm
+        ? sql`
+          AND (
+        CASE 
+          WHEN EXISTS (
+            SELECT 1
+            FROM pg_extension
+            WHERE extname = 'pg_trgm'
+          ) 
+          THEN similarity(contract_call_function_name, ${searchTerm}) > ${TRGM_SIMILARITY_THRESHOLD}
+          ELSE contract_call_function_name ILIKE '%' || ${searchTerm} || '%'
+          END
+          )
+        `
+        : sql``;
+
       const contractFuncFilterSql = functionName
         ? sql`AND contract_call_function_name = ${functionName}`
         : sql``;
@@ -1479,6 +1500,7 @@ export class PgStore extends BasePgStore {
         !startTime &&
         !endTime &&
         !contractId &&
+        !searchTerm &&
         !functionName &&
         !nonce;
 
@@ -1497,6 +1519,7 @@ export class PgStore extends BasePgStore {
         ${startTimeFilterSql}
         ${endTimeFilterSql}
         ${contractIdFilterSql}
+        ${searchTermFilterSql}
         ${contractFuncFilterSql}
         ${nonceFilterSql}
       `;
@@ -1511,6 +1534,7 @@ export class PgStore extends BasePgStore {
         ${startTimeFilterSql}
         ${endTimeFilterSql}
         ${contractIdFilterSql}
+        ${searchTermFilterSql}
         ${contractFuncFilterSql}
         ${nonceFilterSql}
         ${orderBySql}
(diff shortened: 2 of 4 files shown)


src/datastore/pg-store.ts
```diff
@@ -1467,6 +1471,10 @@ export class PgStore extends BasePgStore {
       const contractIdFilterSql = contractId
         ? sql`AND contract_call_contract_id = ${contractId}`
         : sql``;
+
+      const searchTermFilterSql = searchTerm
+        ? sql`AND similarity(contract_call_function_name, ${searchTerm}) > ${TRGM_SIMILARITY_THRESHOLD}`
```
zone117x, line 1476: Have you been able to test this `similarity` performance on our staging environments?

He1DAr: Yep, I got the same result but with more forgiving match (allowing more typos, etc.)

src/api/routes/tx.ts
```diff
@@ -113,6 +113,12 @@ export const TxRoutes: FastifyPluginAsync<
               examples: [1706745599],
             })
           ),
+          search_term: Type.Optional(
+            Type.String({
+              description: 'Option to search for transactions by a search term',
+              examples: ['swap'],
+            })
+          ),
```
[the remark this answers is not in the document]
He1DAr: The idea is to have the `search_term` as a broad search filter that looks into multiple fields. My plan so far is to incrementally add more fields to the same index and test performance, e.g.:
```
CREATE INDEX idx_search_term_trgm
ON public.txs
USING gin (
  contract_call_function_name gin_trgm_ops,
  contract_name gin_trgm_ops
  memo gin_trgm_ops
  ....
);
```
Open to suggestions ofc.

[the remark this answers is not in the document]
zone117x: That seems fine to me. In the future we could always add another query param that lets users specify a given field(s) to search. And for now it can default to "all".

src/datastore/pg-store.ts
```diff
@@ -101,6 +101,8 @@ import { parseBlockParam } from '../api/routes/v2/schemas';
 
 export const MIGRATIONS_DIR = path.join(REPO_DIR, 'migrations');
 
+const TRGM_SIMILARITY_THRESHOLD = 0.3;
```
rafaelcr, line 104: Why `0.3`? Should we make this an ENV variable to be able to tweak it on our deployments?

migrations/1727186561690_idx-contract-call-function-name-trgm.js
```diff
@@ -0,0 +1,19 @@
+/** @param { import("node-pg-migrate").MigrationBuilder } pgm */
+exports.up = pgm => {
+  pgm.sql('CREATE EXTENSION IF NOT EXISTS pg_trgm;');
```
rafaelcr, line 3: API deployers might not have this extension installed. You should validate first with 
```sql
SELECT * FROM pg_available_extensions WHERE name = 'pg_trgm';
```
and also consider this possibility in your query functions

rafaelcr (changes requested):

src/datastore/pg-store.ts
He1DAr: It's actually the default similarity factor (thought defining it explicitly would make it clearer and easier to change if needed), I don't see it changing frequently unless we decide to make the search term more forgiving/strict in the future (based on user feedback, etc.).

src/datastore/pg-store.ts
```diff
@@ -1468,6 +1483,15 @@ export class PgStore extends BasePgStore {
       const contractIdFilterSql = contractId
         ? sql`AND contract_call_contract_id = ${contractId}`
         : sql``;
+
+      const pgTrgmInstalled = await this.isPgTrgmInstalled(sql);
```
zone117x, line 1487: Can this be inlined to avoid another round-trip query to postgres? E.g. something like
```ts
const searchTermFilterSql = searchTerm
  ? sql`
      AND (
        CASE
          WHEN EXISTS (
            SELECT 1
            FROM pg_extension
            WHERE extname = 'pg_trgm'
          )
          THEN similarity(contract_call_function_name, ${searchTerm}) > ${TRGM_SIMILARITY_THRESHOLD}
          ELSE contract_call_function_name ILIKE '%' || ${searchTerm} || '%'
        END
      )
    `
  : sql``;
```

zone117x (commented):
The code looks good to me. My last concern before merging is around query performance. @He1DAr can you work with devops to get this branch deployed to our dev environment?

rafaelcr (approved):
LGTM, but I agree that we should deploy this to dev to give it a test before merging

saralab:
@zone117x @rafaelcr What's the latest on this? We discussed the next steps here a while back, I would like to close all stale PRs ASAP

zone117x:
In the associated issue I outlined some possible next steps that could be taken to make this feature work https://github.com/hirosystems/stacks-blockchain-api/issues/2164#issuecomment-2748487556

PR #2103, opened 2024-10-01
feat: add block etag

### LINKED ISSUE #1230: add cache handlers for individual items (blocks, microblocks, principals)

rafaelcr:
Adds a cache handler for endpoints that show data for individual blocks:
```ts
`${index_block_hash}:${canonical}`
```

Fixes #1230

### DIFF
--- a/src/api/controllers/cache-controller.ts
+++ b/src/api/controllers/cache-controller.ts
@@ -8,6 +8,7 @@ import {
   sha256,
 } from '@hirosystems/api-toolkit';
 import { FastifyReply, FastifyRequest } from 'fastify';
+import { BlockParams } from '../routes/v2/schemas';
 
 /**
  * Describes a key-value to be saved into a request's locals, representing the current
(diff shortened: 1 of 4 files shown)


zone117x (approved):
Nice 💯

PR #2104, opened 2024-10-02
feat: return estimated balance in account balance endpoints

### LINKED ISSUE #2035: Add property for estimated balance to balances endpoint

rafaelcr:
This PR adds an optional `estimated_balance` entry to the balance endpoint responses which reflects the current confirmed STX balance plus any STX delta (sent or received) in transactions currently visible in the mempool.

Fixes #2035

### DIFF
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -4475,9 +4475,12 @@ export class PgStore extends BasePgStore {
             (
               SELECT tx_id
               FROM mempool_txs
-              WHERE sender_address = ${principal}
+              WHERE pruned = false AND
+                (sender_address = ${principal}
                 OR sponsor_address = ${principal}
-                OR token_transfer_recipient_address = ${principal}
+                OR token_transfer_recipient_address = ${principal})
+              ORDER BY receipt_time DESC, sender_address DESC, nonce DESC
+              LIMIT 1
             )`
             : this.sql``
         }
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -2493,7 +2493,7 @@ export class PgStore extends BasePgStore {
         WHERE sponsor_address = ${principal}
       ),
       received AS (
-        SELECT token_transfer_amount
+        SELECT token_transfer_amount AS total
         FROM mempool_txs
         WHERE token_transfer_recipient_address = ${principal}
       )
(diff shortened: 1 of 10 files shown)


src/datastore/pg-store.ts
```diff
@@ -2476,6 +2476,36 @@ export class PgStore extends BasePgStore {
...
   }
 
+  /**
+   * Returns the total STX balance delta affecting a principal from transactions currently in the
+   * mempool.
+   */
+  async getPrincipalMempoolStxBalanceDelta(sql: PgSqlClient, principal: string): Promise<bigint> {
+    const results = await sql<{ delta: string }[]>`
+      WITH sent AS (
+        SELECT token_transfer_amount + fee_rate AS total
```
zone117x, line 2486: What happens here with tx types that have `token_transfer_amount==null`, for example contract-calls?

rafaelcr: Good catch, done

src/datastore/pg-store.ts
```diff
@@ -2483,14 +2483,14 @@ export class PgStore extends BasePgStore {
   async getPrincipalMempoolStxBalanceDelta(sql: PgSqlClient, principal: string): Promise<bigint> {
     const results = await sql<{ delta: string }[]>`
       WITH sent AS (
-        SELECT token_transfer_amount + fee_rate AS total
+        SELECT token_transfer_amount + COALESCE(fee_rate, 0) AS total
```
zone117x, line 2486: This can end up running `null + <number>` right, does it need `COALESCE(token_transfer_amount, 0)`?

zone117x: Might be useful to add another type of tx in the tests

rafaelcr: Oh duh, added the coalesce to the wrong field lol... yep I will add a test for this

zone117x (approved):
Great job with this feature!

PR #2110, opened 2024-10-07
fix: add preexisting declaration copy step in build

### LINKED ISSUE #2109: Types for new API client aren't working.
The intended way of using the new `openapi-fetch` client, doesn't work yet; making it hard for developers to transition.

Issue: The `generated/schema` and `types` files are only shipped in `src/` but not in the `lib/` dir, which is used by node (and correctly so).

---

Ref: https://discord.com/channels/621759717756370964/1288837893581832265/1292097445693755393

---

Potential further issue: It doesn't seem like the `params` (e.g. path txid) are type specified in the generated spec either. So fixing the above issue might still leave developers confused about how to add the `opts` object correctly (as well as the `path` {} patterns).

janniks:
- closes #2109

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -17,6 +17,7 @@
       },
       "devDependencies": {
         "concurrently": "7.6.0",
+        "cpx": "1.5.0",
         "http-server": "14.0.0",
         "microbundle": "0.15.1",
         "rimraf": "6.0.1",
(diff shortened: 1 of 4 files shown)


zone117x (commented):
Have you tried renaming `schema.d.ts` to `schema.ts`?

janniks:
Renaming also works, and "re"generates the d.ts files, but also creates empty .js files for the .ts files.
TSC apparently isn't meant for using .d.ts files, it only compiles and creates new declaration files.

_Tested this out and get the types in my local setup._ 🙌

zone117x (approved):
🎉

PR #2117, opened 2024-10-11
fix: update mempool garbage collection logic for 3.0

rafaelcr:
Changes the logic for mempool transaction garbage collection depending on which Stacks epoch is active so it adapts to the corresponding implementation on stacks-core:
* If epoch is 2.5 or lower, prune after 256 blocks
* If epoch is 3.0 or higher, prune after 2560 minutes

### DIFF
--- a/tests/api/mempool.test.ts
+++ b/tests/api/mempool.test.ts
@@ -42,41 +42,74 @@ describe('mempool tests', () => {
     await migrate('down');
   });
 
-  test('garbage collection', async () => {
-    const garbageThresholdOrig = process.env.STACKS_MEMPOOL_TX_GARBAGE_COLLECTION_THRESHOLD;
-    process.env.STACKS_MEMPOOL_TX_GARBAGE_COLLECTION_THRESHOLD = '2';
-    try {
-      // Insert 5 blocks with 1 mempool tx each.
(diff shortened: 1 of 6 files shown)


tests/api/mempool.test.ts
```diff
@@ -42,41 +42,74 @@ describe('mempool tests', () => {
...
+  test('garbage collection post 3.0', async () => {
+    // Insert 3 txs spaced out so garbage collection kicks in.
+    for (let block_height = 1; block_height <= 3; block_height++) {
+      const block = new TestBlockBuilder({
+        block_height: block_height,
+        index_block_hash: `0x0${block_height}`,
+        parent_index_block_hash: `0x0${block_height - 1}`,
+        signer_bitvec: '1111',
+      })
+        .addTx({ tx_id: `0x111${block_height}`, nonce: block_height })
```
zone117x, line 89: Should this insert artificially aged txs so that we can test the epoch3 collection behavior?

rafaelcr: That part happens in line 92 when the mempool transactions are inserted with receipt time of 1

zone117x: Oh I missed that!

zone117x (approved):

PR #2120, opened 2024-10-14
fix: randomize key order for testnet stx faucet transactions

rafaelcr:
This PR makes a few fixes to the testnet STX faucet:
* Instead of forming an array of `TestnetNetwork`s with the `STACKS_FAUCET_NODE_HOST` and `STACKS_FAUCET_NODE_PORT` values from .env plus the standard testnet network, these ENV vars are now treated only as **override** values. This fixes an issue where both pairs of values essentially pointed to the same network nodes causing faucet transactions to be duplicated in certain scenarios.
* Instead of always attempting to send faucet STX from the first key, keys are now selected in random order. If a key fails from nonce issues, other keys will be tried until we run out of keys. This spreads out the load and reduces the probability of failed transactions.
* Nonce values are now taken from the API nonce estimator, which includes transactions currently in the mempool. This reduces the possibility of nonce conflicts.
* Transaction fees are now estimated by stacks.js, with a fallback to a default fee of 200 ustx.

### DIFF
--- a/src/api/routes/faucets.ts
+++ b/src/api/routes/faucets.ts
@@ -409,7 +409,7 @@ export const FaucetRoutes: FastifyPluginAsync<
               }
               // Try with the next key. Wrap around the keys array if necessary.
               keyIndex++;
-              if (keyIndex > STX_FAUCET_KEYS.length) keyIndex = 0;
+              if (keyIndex >= STX_FAUCET_KEYS.length) keyIndex = 0;
               logger.warn(
                 `StxFaucet transaction failed for sender ${senderAddress}, trying with next key: ${error}`
               );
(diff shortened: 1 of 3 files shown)


zone117x (approved):
LGTM!

src/api/routes/faucets.ts
```diff
@@ -230,9 +214,73 @@ export const FaucetRoutes: FastifyPluginAsync<
...
+        senderKey,
+        network,
+        memo: 'faucet',
+        anchorMode: AnchorMode.Any,
+        nonce,
+      };
+      if (fee) options.fee = fee;
+
+      // Detect possible custom network chain ID
+      network.chainId = await fetchNetworkChainID(network);
```
CharlieC3, line 269: @zone117x Just curious, why not use the `STACKS_CHAIN_ID` [env var](https://github.com/hirosystems/stacks-blockchain-api/blob/fix/faucet-keys/.env#L102C1-L102C16) here?

PR #2121, opened 2024-10-14
chore: testing with devnet

tippenein:
containerized clarinet devnet for api testing


docker/devnet/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+      - "20443:20443"
+      - '20444:20444'
+      - "18443:18443"
+      - "5490:5490"
+    volumes:
+      - chainstate:/chainstate
+      - ../stacks-blockchain/:/app/config
+      - ../stacks-blockchain/.chaindata:/tmp/stacks-blockchain-data
+    environment:
+      - CLARINET_VERSION=v2.8.0
```
hugocaillard, line 30: Is it needed?

docker/devnet/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+      - STACKS_EVENT_OBSERVER=host.docker.internal:3700
```
hugocaillard, line 31: Is it needed?

docker/devnet/settings/Devnet.toml
```diff
@@ -0,0 +1,169 @@
...
+# orchestrator_port = 20445
+# bitcoin_node_p2p_port = 18444
+bitcoin_node_rpc_port = 18443
+bitcoin_node_username = "devnet"
+bitcoin_node_password = "devnet"
+# bitcoin_controller_block_time = 30_000
+stacks_node_rpc_port = 20443
+# stacks_node_p2p_port = 20444
+# stacks_api_port = 3999
+stacks_api_events_port = 3700
```
hugocaillard, line 97: i think this is a now impact since the devnet doesn't run the api. (also it's the default), but I wouldn't set it here to avoid confusion

hugocaillard (commented):
A few questions and comments.
Mostly to avoid unecessay configs that can be confusing

Issue #2089, opened 2024-09-25
Ingest `signer_signature_hash` from `/new_block` payload


zone117x:
This field is required to identify a Stacks block when given a `signer_signature_hash` like when dealing with `/stackerdb_chunks` messages.

There are a couple other additional new fields that we should also consider:
https://github.com/stacks-network/stacks-core/blob/447bb1519865be75a402f51cbb2d97f0db97186c/testnet/stacks-node/src/event_dispatcher.rs#L664-L675

PR #2125, opened 2024-10-18
feat: ingest `signer_signature` from `/new_block` event and expose in new endpoint

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2089

Ingest the `signer_signature` data from the `/new_block` payload. These are stored as a `bytea[]` type in the `blocks` table.

In epoch3.0 the `signer_signature` can be up to 4000 entries in length, 65 bytes each. So a new paginated endpoint is added to expose these rather than including them in existing block responses.

### Example
`GET /extended/v2/blocks/{height_or_hash}/signer-signature`
```json
{
  "limit": 500,
  "offset": 0,
  "total": 2,
  "results": [
    "0x00389af3b37f7a0bdd521bb053bf31c95f62082ed30ac6c6ed1eab9ef2453fe62368827d0741adef4ead4b48b6b3af0108836c95e6ecef76fc563f1194628e8062",
    "0x0028c9d37a458df35f5a23d4799e97739facae07143642baedc101e7cf9d470d7a2de562bc5efaa914b15182035ce520cd7a3639f342776796da9c8c6001c84048"
  ]
}
```

### DIFF
--- a/src/api/routes/v2/blocks.ts
+++ b/src/api/routes/v2/blocks.ts
@@ -9,9 +9,16 @@ import { Server } from 'node:http';
 import { CursorOffsetParam, LimitParam, OffsetParam } from '../../schemas/params';
 import { getPagingQueryLimit, pagingQueryLimits, ResourceType } from '../../pagination';
 import { PaginatedResponse } from '../../schemas/util';
-import { NakamotoBlock, NakamotoBlockSchema } from '../../schemas/entities/block';
+import {
+  NakamotoBlock,
+  NakamotoBlockSchema,
+  SignerSignatureSchema,
+} from '../../schemas/entities/block';
 import { TransactionSchema } from '../../schemas/entities/transactions';
-import { BlockListV2ResponseSchema } from '../../schemas/responses/responses';
+import {
+  BlockListV2ResponseSchema,
+  BlockSignerSignatureResponseSchema,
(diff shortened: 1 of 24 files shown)


src/api/routes/v2/blocks.ts
```diff
@@ -174,5 +181,53 @@ export const BlockRoutesV2: FastifyPluginAsync<
     }
   );
 
+  fastify.get(
+    '/:height_or_hash/signer-signature',
```
rafaelcr, line 185: I think plural might make more sense here given that it's a paginated endpoint i.e. `signer-signatures`

rafaelcr (approved):
Awesome, LGTM. Just a suggestion about the path name but I'm fine either way

src/api/routes/v2/blocks.ts
zone117x: There's weirdness around this because IIUC the initial (and eventual) design from stacks-core is a single `signer_signature` field for a block, and this array value is only an epoch3.0 quirk. So singular naming I think would be for future-compatibility (it would return an array with 1 item, still kinda weird).

... but if core does make that change in the future, we'd probably still want to receive and expose this list of all signer signatures, so plural naming is also future-compatible. I'll make that change.

Issue #2131, opened 2024-10-21
Rosetta: use Nakamoto block timestamp


zone117x:
The `timestamp` field currently returned for rosetta responses is the `burn_block` (bitcoin block) timestamp. This should be updated to use the Nakamoto Stacks block timestamp for epoch3 blocks, otherwise blocks will return the same timestamp for all blocks in a tenure.

PR #2132, opened 2024-10-21
Fix(rosetta): Nakamoto timestamps

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2131

In Rosetta block responses, the `timestamp` field uses Nakamoto block times in epoch3 (now that they are consensus-level / reliable). 

In epoch2.x, the burn-block-time is still used.


rafaelcr (approved):
LGTM

PR #2133, opened 2024-10-21
fix: event-replay block parsing outdated and incorrect

zone117x:
The event-replay logic was using a copy&pasted function for parsing `/new_block` data, which had become outdated and incorrect. 

This removes the duplicated code, and both event-replay and regular block ingestion use the same (correct) block parsing code.


rafaelcr (approved):
💯

Issue #2124, opened 2024-10-18
Include block `tenure_height`


zone117x:
Related issue: https://github.com/hirosystems/explorer/issues/1624

API currently doesn't expose (or know about) tenure heights for blocks.

Blocked on https://github.com/stacks-network/stacks-core/issues/5318

Issue #2130, opened 2024-10-21
Rosetta: include `tenure_height`


zone117x:
Related & blocking issue: https://github.com/hirosystems/stacks-blockchain-api/issues/2124

Exchanges uses Rosetta will likely want to use tenure-height rather that block-height for confirmation logic. Tenure height can be added to the Rosetta responses once https://github.com/hirosystems/stacks-blockchain-api/issues/2124 is done.

PR #2134, opened 2024-10-21
feat: include `tenure-height` in block responses

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2124

And progress towards unblocking https://github.com/hirosystems/stacks-blockchain-api/issues/2130

Ingest the `tenure_height` value added to `/new_block` events in https://github.com/stacks-network/stacks-core/pull/5333 (just added in core, requires a new RC).

If the `tenure_height` is available, it's always returned in block endpoint responses. If it's missing and the block is epoch2.x then the `block_height` is returned as the `tenure_height` (which is always accurate to assume). 

If it's missing in an epoch3 block, then `-1` is returned to indicate an error. The API will also print warnings logs when ingesting an epoch3 block event that is missing `tenure_change` to indicate that the stacks-core version needs to be updated. This was done to prevent breaking current Nakamoto deployments that don't yet have an update stacks-core.


rafaelcr (approved):
LGTM

Issue #1813, opened 2024-01-10
specify which transaction replaced another via RBF


rafaelcr:
In a mempool transaction that was replaced-by-fee ([example](https://api.stg.hiro.so/extended/v1/tx/0x7e70eff873644af46fcb32eac885120c74f0079cfccea6439b96b26330c542a1)), we should return the transaction ID of the new transaction. This will allow the Explorer to show this in the transaction page.

PR #2141, opened 2024-10-24
feat: replacing txid when mempool transaction is dropped by fee and r…

usagi32:
solves #1813 

This will provide extra information for a transaction which is replaced by another. This can be used for replace-by-fee or maybe replace-across-fork but not sure though if replace-across-fork gives a new_txid or uses the previous one. 

Nevertheless, @zone117x


usagi32:
There is a [requirement ](https://github.com/stacks-network/stacks-core/pull/5381)in the stacks core repo to be merged for this pull

migrations/1715844926068_add-replacing-txid.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable camelcase */
+/** @param { import("node-pg-migrate").MigrationBuilder } pgm */
+
+exports.up = pgm => {
+  pgm.addColumn('mempool_txs', {
+    replacing_txid: {
+      type: 'string',
```
rafaelcr, line 7: This needs to be `bytea`

src/datastore/common.ts
```diff
@@ -1235,6 +1238,7 @@ export interface MempoolTxInsertValues {
   type_id: DbTxTypeId;
   anchor_mode: DbTxAnchorMode;
   status: DbTxStatus;
+  replacing_txid: string | null,
```
rafaelcr, line 1241: `PgBytea | null`

rafaelcr (commented):
Apologies for the very extensive delay @usagi32 , left a few comments. After that it LGTM

PR #2144, opened 2024-10-27
Add Route to Fetch Blocks by Tenure Height

bowtiedbot:
Add a route that opens an endpoint on the API that allows users to fetch the blocks by tenure height:
```bash
/extended/v2/tenure-height/{height}/blocks
```

This endpoint returns the same data as `/extended/v2/burn-blocks/{height_or_hash}/blocks` does, but by the tenure height instead of the burn height/hash.

### DIFF
--- a/src/api/init.ts
+++ b/src/api/init.ts
@@ -47,6 +47,7 @@ import {
 } from '@hirosystems/api-toolkit';
 import { BlockRoutesV2 } from './routes/v2/blocks';
 import { BurnBlockRoutesV2 } from './routes/v2/burn-blocks';
+import { TenureHeightRoutesV2 } from './routes/v2/tenure-heights';
 import { MempoolRoutesV2 } from './routes/v2/mempool';
 import { SmartContractRoutesV2 } from './routes/v2/smart-contracts';
 import { AddressRoutesV2 } from './routes/v2/addresses';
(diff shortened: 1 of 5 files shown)


src/api/init.ts
```diff
@@ -99,6 +100,7 @@ export const StacksApiRoutes: FastifyPluginAsync<
     async fastify => {
       await fastify.register(BlockRoutesV2, { prefix: '/blocks' });
       await fastify.register(BurnBlockRoutesV2, { prefix: '/burn-blocks' });
+      await fastify.register(TenureHeightRoutesV2, { prefix: '/tenure-height' });
```
rafaelcr, line 103: For v2 endpoints we're trying to keep route names as REST-ful as possible, so we want to group endpoints according to the entity that they return. Adding `tenure-height` here would not fit the standard.

Could you instead add a `tenure_height` query param to the `/extended/v2/blocks` endpoint? That will also allow you to get cursor pagination for free and will work much better with API clients.

rafaelcr (changes requested):
Hi @bowtiedbot thanks for the contribution and apologies for the review delay.

src/api/init.ts
bowtiedbot: Hello @rafaelcr! I am currently trying to modify the query from pg-store-v2, but when parsing a tenure height query, the results list is empty. Can you please take a look over it and let me know what I'm doing wrong? Thank you!

https://github.com/degen-lab/stacks-blockchain-api/blob/test/get-blocks-by-tenure-height/src/datastore/pg-store-v2.ts#L76-L140

PR #2154, opened 2024-11-01
remove unused docs content

ryanwaits:
## What does this PR do?

Since all of our docs content now exists entirely in the docs repo, this removes unused `content` directory to avoid any confusion or duplication of old/outdated information.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

No

## Checklist
- [x] Delete `content` directory


ryanwaits:
FYI, it doesn't look like we are using this `content` directory anymore so I figured I would make it easy and get the PR set up. Feel free to close if this is not the case.

zone117x (approved):
LGTM!

PR #2172, opened 2024-11-19
feat: option to modify estimated fees in /v2/fees/transaction proxy

zone117x:
Adds a hook to the `/v2/fees/transaction` endpoint in the RPC proxy. 

Configure using the env var `STACKS_CORE_FEE_ESTIMATION_MODIFIER`. The fee estimations returned by the stacks-node will be multiplied by this value.

For example with `STACKS_CORE_FEE_ESTIMATION_MODIFIER=0.5`, if the response from stacks-core is:
```
estimations: [
  { fee: 400 },
  { fee: 800 },
  { fee: 1000 }
]
```
Then the modified response will be:
```
estimations: [
  { fee: 200 },
  { fee: 400 },
  { fee: 500 }
]
```

### DIFF
--- a/src/api/routes/core-node-rpc-proxy.ts
+++ b/src/api/routes/core-node-rpc-proxy.ts
@@ -152,9 +152,7 @@ export const CoreNodeRpcProxyRouter: FastifyPluginAsync<
     upstream: `http://${stacksNodeRpcEndpoint}`,
     rewritePrefix: '/v2',
     http2: false,
-    // Use the `node:http` agent for requests https://github.com/fastify/fastify-reply-from?tab=readme-ov-file#globalagent
     globalAgent: true,
-    http: {},
     preValidation: async (req, reply) => {
       if (getReqUrl(req).pathname !== '/v2/transactions') {
         return;
--- a/src/api/routes/core-node-rpc-proxy.ts
+++ b/src/api/routes/core-node-rpc-proxy.ts
@@ -21,6 +21,38 @@ function getReqUrl(req: { url: string; hostname: string }): URL {
   return new URL(req.url, `http://${req.hostname}`);
 }
 
+// https://github.com/stacks-network/stacks-core/blob/20d5137438c7d169ea97dd2b6a4d51b8374a4751/stackslib/src/chainstate/stacks/db/blocks.rs#L338
+const MINIMUM_TX_FEE_RATE_PER_BYTE = 1;
+
+function getFeeEstimationModifier(): number | null {
+  const feeEstimationModifier = process.env['STACKS_CORE_FEE_ESTIMATION_MODIFIER'];
+  if (!feeEstimationModifier) {
+    return null;
+  }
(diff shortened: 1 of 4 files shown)


src/api/routes/core-node-rpc-proxy.ts
```diff
@@ -21,6 +21,38 @@ function getReqUrl(req: { url: string; hostname: string }): URL {
   return new URL(req.url, `http://${req.hostname}`);
 }
 
+// https://github.com/stacks-network/stacks-core/blob/20d5137438c7d169ea97dd2b6a4d51b8374a4751/stackslib/src/chainstate/stacks/db/blocks.rs#L338
+const MINIMUM_TX_FEE_RATE_PER_BYTE = 1;
+
+function getFeeEstimationModifier(): number | null {
```
rafaelcr, line 27: Can we do this in a lazy way? e.g. parse once and keep it cached in a global var

src/api/routes/core-node-rpc-proxy.ts
```diff
@@ -121,6 +153,9 @@ export const CoreNodeRpcProxyRouter: FastifyPluginAsync<
     upstream: `http://${stacksNodeRpcEndpoint}`,
     rewritePrefix: '/v2',
     http2: false,
+    // Use the `node:http` agent for requests https://github.com/fastify/fastify-reply-from?tab=readme-ov-file#globalagent
+    globalAgent: true,
+    http: {},
```
zone117x, line 158: Fastify defaults to using Fetch/undici which isn't supported by the `nock` testing lib used to mock network responses. These settings cause it to use the `node:http` global agent. From what I understand, it's a bit slower but probably fine.

src/api/routes/core-node-rpc-proxy.ts
zone117x: Let me see if I can do that in a way that still lets the test configure it..

src/api/routes/core-node-rpc-proxy.ts
rafaelcr: Curious, have you tried using the `MockAgent`? https://github.com/hirosystems/token-metadata-api/blob/0c02a159290925963d516403d712582e79c7db2d/tests/token-queue/process-smart-contract-job.test.ts#L44-L56

src/api/routes/core-node-rpc-proxy.ts
zone117x: Done

src/api/routes/core-node-rpc-proxy.ts
zone117x: Nice looks like that works https://github.com/hirosystems/stacks-blockchain-api/pull/2172/commits/6ab509e5087f9601ef3ba52479a3cd66b3000490

rafaelcr (approved):
LGTM!

obycode (approved):
LGTM!

Issue #2146, opened 2024-10-28
subscribeBlocks & subscribeMempool Not Working


dhwndud408:
Hi! dev teams.

I am making code with your api.
Here is my computer info.
Mac OS Sonoma14.4
Node 18.19.1
NPM 10.2.4
API version: 8.2.0

Here is the typescript code

```
import * as stacks from '@stacks/blockchain-api-client';
const socketUrl = "https://api.mainnet.hiro.so";
const socket = new stacks.StacksApiSocketClient({ url: socketUrl });

let client = await stacks.connectWebSocketClient('wss://api.mainnet.hiro.so/');


sub = await socket.subscribeBlocks((event) => {
   console.log(event)
}
sub = await socket.subscribeBlocks((event) => {
console.log(event)
}
```
After about 6 hours, the websocket stops.

I also see the this report.
https://github.com/hirosystems/stacks-blockchain-api/issues/1885

The admin said it was resolved on above link, but it seems the issue has resurfaced with the Nakamoto update.

zone117x:
Did you try the code in the referenced issue?

> if you specify the `transports` option explicitly then the code works, here's an example:
> 
> 
> 
> ```js
> 
> let s = document.createElement('script');
> 
> s.src = "https://www.unpkg.com/@stacks/blockchain-api-client@7.10.0/lib/index.umd.js";
> 
> document.head.appendChild(s);
> 
> s.onload = () => {
> 
>   let client = new StacksBlockchainApiClient.StacksApiSocketClient({
> 
>       url: 'https://api.mainnet.hiro.so/',
> 
>       socketOpts: {
> 
>           transports: ['websocket'], // <-- workaround
> 
>       }
> 
>   });
> 
>   client.subscribeBlocks(block => console.log('block', block));
> 
>   client.subscribeMempool(tx => console.log('mempool', tx));
> 
> };
> 
> ```

 _Originally posted by @zone117x in [#1885](https://github.com/hirosystems/stacks-blockchain-api/issues/1885#issuecomment-2096304075)_



Also, please follow these socket-io debugging instructions and include the error debug logs here.

dhwndud408:
> Did you try the code in the referenced issue?
> 
> > if you specify the `transports` option explicitly then the code works, here's an example:
> > let s = document.createElement('script');
> > 
> > s.src = "https://www.unpkg.com/@stacks/blockchain-api-client@7.10.0/lib/index.umd.js";
> > 
[...]

PR #2174, opened 2024-11-21
Fix: socket-io reconnect

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2146

There's a bug in the socket-io client reconnection logic, where an invalid empty topic was being sent. 
```js
client.socket.on('connect_error', error => {
  console.error('Socket connection was denied by the server', error);
  // > Socket connection was denied by the server Error: Invalid topic: 
});
```

This PR fixes the bug in the client, and also updates the server code to ignore empty topics so that existing client versions will work.


rafaelcr (approved):
LGTM

Issue #2187, opened 2024-12-17
Change tBTC faucet amounts


CharlieC3:
Currently the tBTC faucet allows users to request 0.5 tBTC. I feel this default is too high now that sBTC is available and people will be requesting tBTC to play with it on testnet.

We should update the default amount to 0.01 tBTC and allow users to request more, which would be the original amount of 0.5.

PR #2188, opened 2024-12-17
Feat/tbtc faucet amounts

CharlieC3:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2187

* Changes the default tBTC faucet amount from 0.5 -> 0.01 tBTC
* Adds a URL query option to the tBTC faucet to allow the user the request a larger amount of tBTC (the original amount of 0.5)
* Updates some Bitcoin-related text from `testnet` to `regtest` since we've replaced Bitcoin testnet with regtest on Stacks testnet

### DIFF
--- a/src/api/routes/faucets.ts
+++ b/src/api/routes/faucets.ts
@@ -74,25 +74,25 @@ export const FaucetRoutes: FastifyPluginAsync<
       preHandler: missingBtcConfigMiddleware,
       schema: {
         operationId: 'run_faucet_btc',
-        summary: 'Add testnet BTC tokens to address',
-        description: `Add 0.01 BTC token to the specified testnet BTC address.
+        summary: 'Add regtest BTC tokens to address',
+        description: `Add 0.01 BTC token to the specified regtest BTC address.
 
-        The endpoint returns the transaction ID, which you can use to view the transaction in a testnet Bitcoin block
+        The endpoint returns the transaction ID, which you can use to view the transaction in a regtest Bitcoin block
         explorer. The tokens are delivered once the transaction has been included in a block.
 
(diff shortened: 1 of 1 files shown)


CharlieC3:
Will fix the btc faucet test in a bit.

obycode:
If we're concerned about the amount of BTC available, I'd recommend lowering this even further. If we think about in Sats, 0.01 is still a million Sats, so we could go much lower. We'd also want to make sure that our Stacks miners are keeping their spend low, so that other miners can join in with low amounts.

To help out, we could also:
- Add a sink, an easy way for people to send their tBTC back to the faucet
- Make sure that our stackers are sending their tBTC back to the faucet (or directly specifying a faucet address in their stacking commands)

rafaelcr (approved):

CharlieC3:
Agreed on lowering the rBTC amount even more in wake of recent discussions, and I like the ideas you mentioned! Those can definitely be ironed out or configured in Q1 2025.
@obycode What default rBTC were you thinking would be suitable if not 0.01? 1k sats?

obycode:
> What default rBTC were you thinking would be suitable if not 0.01? 1k sats?

We may need to go a little higher than 1k because we have to account for what the node considers dust amounts, and will not accept transactions below, which is something like 500-600 sats for a simple transfer. Maybe 10k would be good?

aldur:
Thank you for this! If you let me know once 8.4.0 gets deployed to testnet API, I'll update the sBTC bridge code to request less.

CharlieC3:
@aldur This is live! default rBTC faucet requests are transferring 10k sats
https://mempool.bitcoin.regtest.hiro.so/tx/9d19c0f1ceff1c928d2776f49881869447cfd346d0ceea6d00af5fbb3edbd069

PR #2189, opened 2024-12-17
Update develop with master

CharlieC3:
We're unable to cut `beta` branches from `develop` as it seems the past few releases aren't in `develop`.
I'd like to release the new btc faucet amount changes this week if possible, so updating `develop` is needed to do that.


zone117x (approved):
Just need to fix the package lock conflict (probably just small conflicts from the dep bots)

PR #2192, opened 2025-01-06
feat: support redis stream as event source

rafaelcr:
Adds an option to use a new Redis stream connection instead of the classic HTTP event server


src/datastore/pg-store.ts
```diff
@@ -4524,4 +4524,11 @@ export class PgStore extends BasePgStore {
     `;
     if (result.count) return result[0];
   }
+
+  async getLastEventObserverRequestSequenceId(): Promise<string | undefined> {
+    const result = await this.sql<{ sequence_id: string }[]>`
+      SELECT sequence_id FROM event_observer_requests ORDER BY id DESC LIMIT 1
```
zone117x, line 4530: This might be super slow on mainnet without an index given the size of the table. Fine for now though

rafaelcr: No worries, `id` is the primary key of the table so this query ends up being super fast

zone117x (approved):
💯

PR #2200, opened 2025-01-16
feat: add prom metrics for stacks block count at previous burn block

rafaelcr:
Observe the following metric:

- `stacks_blocks_in_previous_burn_block`: Number of Stacks blocks produced in previous Bitcoin block

This will help us create alerts for when a Bitcoin block gap happens or blocks stop being produced. We also update the chain tip `burn_block_height` with each new burn block event.

### DIFF
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -4531,6 +4531,6 @@ export class PgStore extends BasePgStore {
       FROM blocks
       WHERE burn_block_height = ${burnBlockHeight} AND canonical = TRUE
     `;
-    return parseInt(result[0].count ?? '0');
+    return parseInt(result[0]?.count ?? '0');
   }
 }
--- a/src/datastore/pg-store.ts
+++ b/src/datastore/pg-store.ts
@@ -4525,11 +4525,11 @@ export class PgStore extends BasePgStore {
     if (result.count) return result[0];
   }
 
-  async getStacksBlockCountAtBurnBlock(burnBlockHeight: number): Promise<number> {
+  async getStacksBlockCountAtPreviousBurnBlock(): Promise<number> {
     const result = await this.sql<{ count: string }[]>`
       SELECT COUNT(*) AS count
       FROM blocks
-      WHERE burn_block_height = ${burnBlockHeight} AND canonical = TRUE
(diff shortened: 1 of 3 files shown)


src/datastore/pg-store.ts
```diff
@@ -4524,4 +4524,13 @@ export class PgStore extends BasePgStore {
...
     if (result.count) return result[0];
   }
+
+  async getStacksBlockCountAtBurnBlock(burnBlockHeight: number): Promise<number> {
+    const result = await this.sql<{ count: number }[]>`
+      SELECT COUNT(*) AS count
+      FROM blocks
+      WHERE burn_block_height = ${burnBlockHeight} AND canonical = TRUE
+    `;
+    return result[0].count ?? 0;
```
CharlieC3, line 4534: If an API writer just recently booted up and has not yet received a burn block event, do you know what this metric to show up as? I'm _guessing_ that the prometheus server simply won't display the `stacks_blocks_in_previous_burn_block` metric until a burn block is received and this code path is triggered. Are you able to verify this locally?

I'm asking because I want to avoid the scenario where a writer is restarted and displays the value `0` for a while until it receives the next burn block. This could cause red herrings in our alerts.

rafaelcr: good catch @CharlieC3 , fixed in latest commit so it first appears as blank instead of `0`

src/event-stream/event-server.ts
```diff
@@ -686,7 +699,14 @@ function createMessageProcessorQueue(): EventMessageHandler {
...
       return processorQueue
-        .add(() => observeEvent('burn_block', () => handleBurnBlockMessage(msg, db)))
+        .add(async () => {
+          await observeEvent('burn_block', () => handleBurnBlockMessage(msg, db));
+          if (metrics) {
+            metrics.blocksInPreviousBurnBlock.set(
+              await db.getStacksBlockCountAtBurnBlock(msg.burn_block_height - 1)
+            );
+          }
+        })
```
zone117x, line 709: Should we have this in the event-ingestion path? It could be added to the `async collect() { ... }` function of the `metrics.blocksInPreviousBurnBlock` instance so it's only queried when prom metrics are fetched.

rafaelcr: Great idea, done

zone117x (approved):
LGTM!

PR #2204, opened 2025-01-22
feat: store total transaction size in blocks table

rafaelcr:
This is computed by the sum of all confirmed transaction byte sizes, and will be necessary to calculate tenure fullness in fee estimations.


janniks (approved):
Generally looks great!

Only weird to see `tx_total_size: 1,` in the tests, I'm assuming they don't have a "real" transaction with a correct size?

rafaelcr:
thanks @janniks , exactly, they're only simulated values for tests

migrations/1737567411419_block-tx-total-size.js
```diff
@@ -0,0 +1,25 @@
...
+    },
+  });
+  pgm.sql(`
+    UPDATE blocks
+    SET tx_total_size = (
+      SELECT SUM(OCTET_LENGTH(raw_tx))
+      FROM txs
+      WHERE index_block_hash = blocks.index_block_hash
+    )  
+  `);
```
zone117x, line 20: My gut feeling is that this will result in a several hour long migration for most external deployments (e.g. exchanges). Seems like it would be fine to let the new column be nullable and we'd only have this available for new blocks.

PR #2207, opened 2025-01-26
fix: make tx_total_size column nullable

rafaelcr:
Since we just need it for fee estimations, it's ok if this column remains null for past blocks


zone117x (approved):
👍

Issue #2199, opened 2025-01-16
Add estimated_inbound and estimated_outbound values in balance requests


alter-eggo:
**Is your feature request related to a problem? Please describe.**
To get immediately available balance that user can send we need to do 2 requests - for balance and mempool transactions, so 
`available balance = total balance - locked balance - outbound txs value`
It would be great to do only 1 request to get available balance

**Describe the solution you'd like**
There is already `estimated_balance` in the balance request which uses inbound/outbound mempool txs values in calc.
so it would be great to add `estimated_inbound` and `estimated_outbound` values in balance request response

alter-eggo:
also wonder guys if its possible to add estimated balance to sip10 tokens as well?

<img width="1249" alt="Image" src="https://github.com/user-attachments/assets/7bf97b9a-09e5-4af6-9d4c-6fc16df0623b" />

PR #2208, opened 2025-01-27
feat: show pending inbound and outbound stx balances

rafaelcr:
Adds `pending_balance_inbound` and `pending_balance_outbound` values to `/balances` endpoints, reflecting balance changes from mempool transactions.

#2199


zone117x (approved):
💯

PR #2217, opened 2025-02-05
fix: use an independent sql connection for mempool stats

rafaelcr:
Previously, this debounced mempool stats calculator query was sharing a single sql connection with the pg writer. If the debounced trigger coincided with a new block ingestion, both tasks reused the same SQL transaction resulting in a race condition: as soon as the block ingestion ended, the mempool stats query was cut short because the transaction was committed. This resulted in a `CONNECTION_DESTROYED` postgres error as described here: https://github.com/stacks-network/stacks-core/issues/5663

This PR moves the mempool stats query to its own SQL connection


zone117x (approved):
Hopefully this finally fixes the issues around this!

wileyj:
confirmed i was able to reach chain tip using this branch, thanks!

Issue #1895, opened 2024-03-19
`/extended/v1/address/:address/balances` needs optimization


rafaelcr:
We're getting OOM errors from a large number of calls to this endpoint. Perhaps it can be optimized or refactored into a new one to use less memory.

smcclellan:
Priority could adjust after Fastify migration.

wafeishushu:
same problem

Issue #2095, opened 2024-09-26
stx_supply and stx account balance endpoint perf issues


zone117x:
The query used for `extended/v1/stx_supply` is getting expensive and taking several seconds (or timing out) when not getting a cache-hit. 

We now have an ft balance table which can calculate the total STX supply near-instantly, which we should switch to using.

EDIT:
Account stx balance endpoints are also starting to have perf issues due to this same underlying problem.

zone117x:
Recently, a series of stx balance read queries have been growing at something like a sublinear rate to the number of tx rows associated with a given address (or at least growing at a constant (O(1)) rate), which is starting to become a problem due to the volume of txs that can be associated with a given address.
We have a new-ish table that tracks balances for each `address<->token`, which means this endpoint could issue balances queries that only need to read a single row. It's the same table powers these fast endpoints:  https://api.hiro.so/extended/v1/tokens/ft/stx/holders

We could make that change for non-STX tokens right now, but for STX balances there is a known bug where some address balances are slightly off. Perhaps its worth making the change anyway knowing some balances will be off, and then we'll get more information that issue and hopefully can fix it.

wileyj:
copypaste from #2226 since tracking was moved here:

This presents almost identically to https://github.com/hirosystems/stacks-blockchain-api/issues/2147


In this case,  querying a an API like:
```
http://localhost:3999/extended/v1/address/{principle}/balances
```

can take quite long (in some cases, i waited over 120s). the host resources/db resources were mostly unchanged and other requests were served during the wait, so it doesn't seem to be blocking. 



```
~$ time curl -sL localhost:3999/extended/v1/address/{principle}/balances | jq -r
{
  "stx": {
    "balance": "14433751",
    "total_sent": "152295994",
    "total_received": "255265978",
    "total_fees_sent": "88536233",
    "total_miner_rewards_received": "0",
    "lock_tx_id": "",
    "locked": "0",
    "lock_height": 0
...
real	0m31.282s
user	0m0.039s
sys	0m0.006s
```

After the first query for the balance, subsequent requests return quickly (as expected), so it feels really similar #2147

PR #2229, opened 2025-03-11
feat: new account balances endpoints, optimized for high tx volume

zone117x:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/1895
Fixes the account-specific endpoint issue in https://github.com/hirosystems/stacks-blockchain-api/issues/2095

This PR implements two new address balance endpoints
* `GET /extended/v2/addresses/:addr/balances/stx`
* `GET /extended/v2/addresses/:addr/balances/ft`

These have the following optimizations:
* Directly queries stx and ft token balances from the `ft_balances` table, rather than aggregating the tx credits/debits/fees over tens of thousands of rows for some accounts.
* FT endpoint is paginated
* Only involve the pox4 events table for determining locked amount.
* Skip `token_offering_locked` query now that all token offering accounts are unlocked.
* Remove `nft_holdings` property from response, there is already an optimized endpoint `/extended/v1/tokens/nft/holdings?principal=<addr>` which includes more useful data than this property.

These are new endpoint rather than modifying the existing one because of breaking changes:
* The ft response objects do not contain `total_sent` and `total_received` fields, because the holders table only tracks the `balance`.
* Does not include the `nft_holdings` property.
* The endpoint does not support the `?until_block=<height>` param, because the holders table contains balances at the chaintip.
* Mempool/pending balance estimations are opt-in with a `?include_mempool=true` query param.

The old endpoints `/extended/v1/address/:addr/balances` and `/extended/v1/address/:addr/stx` have been marked as deprecated.

Sample of `EXPLAIN ANALYZE` timing for the previous FT balance query vs the new one:
```
Execution Time: 1371.540ms (old)
Execution Time: 0.066ms (new)
```

TODO:
* [x] Tests


src/api/routes/v2/addresses.ts
```diff
@@ -118,5 +125,86 @@ export const AddressRoutesV2: FastifyPluginAsync<
...
+    },
+    async (req, reply) => {
+      const stxAddress = req.params.principal;
+      validatePrincipal(stxAddress);
+
+      const result = await fastify.db.sqlTransaction(async sql => {
+        const chainTip = await fastify.db.getChainTip(sql);
+
+        // Get balances for fungible tokens
+        const ftBalancesResult = await fastify.db.v2.getFungibleTokenHolderBalances({
```
rafaelcr, line 154: I think this is a good opportunity to paginate this list somehow... I know there are some addresses that have hundreds of tokens (DEX contracts and such).

Maybe a good approach would be to move these to a new endpoint? `/:principal/balances/ft` (similar to the NFT holdings endpoint). Paginating a nested object in this endpoint would be kind of weird imo

zone117x: Good call, I'll work on that

zone117x: Done

rafaelcr (approved):
LGTM!

PR #2240, opened 2025-03-17
feat: endpoint to get the balance of a specific ft for a given account

zone117x:
In https://github.com/hirosystems/stacks-blockchain-api/pull/2229 we made the new `:addr/balances/ft` endpoint paginated. In some use cases, the deprecated "get all FT balances" endpoint was being used to lookup a single FT balance, which could now require multiple paginated requests using the new endpoint. 

This PR adds a new endpoint to get the balance of a specific FT for a given address:
```
GET /extended/v2/addresses/:addr/balances/ft/:token
```

Example:
```
GET /extended/v2/addresses/SPQC38PW542EQJ5M11CR25P7BS1CA6QT4TBXGB3M/balances/ft/SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token::sbtc-token
```
```json
{
  "balance": "481723"
}
```

### DIFF
--- a/src/api/routes/v2/addresses.ts
+++ b/src/api/routes/v2/addresses.ts
@@ -264,5 +264,50 @@ export const AddressRoutesV2: FastifyPluginAsync<
     }
   );
 
+  fastify.get(
+    '/:principal/balances/ft/:token',
+    {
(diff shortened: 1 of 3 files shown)


rafaelcr (approved):
Nice!

PR #2243, opened 2025-03-17
ci(vercel): move openapi files to vercel output directory after build

rafaelcr:
Make sure `openapi.json` and `openapi.yaml` are in the Vercel output directory after building docs


zone117x (approved):
Nice!

PR #2244, opened 2025-03-17
Ci/image stream

deantchi:
Add steps to also build/push into Google Artifact Registry using `environments` in a Matrix strategy.  Test build can be viewed [here](https://github.com/hirosystems/stacks-blockchain-api/actions/runs/13925730135).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,6 @@ on:
       - develop
       - beta
       - nakamoto
-      - ci/image-stream
     tags-ignore:
       - "**"
     paths-ignore:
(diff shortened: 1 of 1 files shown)


deantchi:
Moved the GAR replication into a separate step to prevent the Docker push step from running three times due to the matrix.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,8 @@ on:
       - "**/package.json"
   pull_request:
   workflow_dispatch:
+env:
+  GIT_BRANCH: ${{ github.ref_name }}
```
CharlieC3, line 18: This is available by default in the `GITHUB_REF_NAME` env var.

CharlieC3 (changes requested):
Looks great! Just one minor change.

Once we introduce CD to this workflow, we'll be able to change the gar upload to an environment only once the env's release gate is opened.

PR #2253, opened 2025-04-07
feat: add `Warning` http response deprecation notices

zone117x:
Adds a `Warning: 299 - "Deprecated ..."` http response header for endpoints marked as deprecated. This should help with developer awareness. 

E.g.
```
curl -i 'https://api.hiro.so/extended/v1/address/SP318Q55DEKHRXJK696033DQN5C54D9K2EE6DHRWP/stx'
HTTP/2 200 
date: Mon, 07 Apr 2025 12:19:21 GMT
...
warning: 299 - "Deprecated: See https://docs.hiro.so/stacks/api for more information"
...
```

There are no deprecation-related http headers available and widely in-use today, however, this one is at least [RFC spec-compliant](https://stackoverflow.com/a/29623798/794962)

---

This is implemented as a generic Fastify plugin, and is a good candidate to port over to our api-toolkit repo.

### DIFF
--- a/src/api/deprecation-plugin.ts
+++ b/src/api/deprecation-plugin.ts
@@ -1,0 +1,39 @@
+import fp from 'fastify-plugin';
+import { FastifyPluginAsync } from 'fastify';
+
+declare module 'fastify' {
+  interface FastifySchema {
+    deprecatedMessage?: string;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


rafaelcr (approved):
Great idea! I agree this should eventually be moved to the toolkit

PR #2255, opened 2025-04-10
feat: mark /mempool/dropped endpoint as legacy deprecated

rafaelcr:
Users with their own deployments can still enable this by setting `STACKS_API_ENABLE_LEGACY_ENDPOINTS` as true in their ENV.


CharlieC3 (approved):
:shipit:

PR #2258, opened 2025-04-18
fix: optimize contract event list db index

rafaelcr:
Optimizes the `/extended/v1/contract/${contractId}/events` endpoint by combining two indexes into one, which prevents the query planner from picking a sub-optimal index and then doing a filter. The planner will use this new index instead, optimizing query response times by ~99%.


CharlieC3 (approved):
Nice! :shipit:

PR #2259, opened 2025-04-18
fix: optimize query for transactions in block

rafaelcr:
Optimizes query for `/extended/v2/blocks/${blockHash}/transactions` by first looking for the `index_block_hash` of the relevant block in the `blocks` table and then using that value in other filters, instead of looking for the block in the `txs` table.

Reduces query time by 99%.


CharlieC3 (approved):
:shipit:

PR #2263, opened 2025-04-28
fix: ignore unanchored param for BNS names endpoint

rafaelcr:
Fixes 500 status error:
```
relation "nft_custody_unanchored" does not exist -> /v1/addresses/stacks/SP2E4MERK02YV3J5MC6SHT1XRVANEJ5DRD80NB3FV?unanchored=true -> 500 -> GET
```
We dropped the `nft_custody_unanchored` table a while ago because we no longer support microblocks.


CharlieC3 (approved):
:shipit:

PR #2266, opened 2025-04-28
fix: keep a table of latest stacks node event timestamps

rafaelcr:
Optimizes the query which retrieves the latest Stacks node event timestamps for Prometheus metrics (currently taking ~3,000ms to complete) by creating a new table that pre-computes the required values.

### DIFF
--- a/migrations/1745877592337_event-observer-timestamps.js
+++ b/migrations/1745877592337_event-observer-timestamps.js
@@ -1,0 +1,20 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = pgm => {
+  pgm.createTable('event_observer_timestamps', {
+    event_path: {
+      type: 'string',
+      primaryKey: true,
+    },
+    id: {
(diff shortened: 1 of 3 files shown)


CharlieC3 (approved):
:shipit:

migrations/1745877592337_event-observer-timestamps.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = pgm => {
+  pgm.createTable('event_observer_timestamps', {
```
zone117x, line 6: Should this new table have a ordered index on the `receive_timestamp` column?

rafaelcr: I don't think it's necessary because it will contain only 1 row per event type `/new_block` `/mempool_tx` etc. and we will always retrieve all rows at the same time for prometheus, not sorted by timestamp

PR #2271, opened 2025-05-02
feat: specify transaction id being replaced in mempool results

### LINKED ISSUE #1813: specify which transaction replaced another via RBF

rafaelcr:
Adds a `replacing_tx_id` key to mempool transaction responses that will include another transaction ID if this represents an RBF operation.

Replaces #2141 
Fixes #1813


tests/api/mempool.test.ts
```diff
@@ -2129,8 +2207,8 @@ describe('mempool tests', () => {
     });
   });
 
-  test('prunes transactions with nonces that were already confirmed', async () => {
-    // Initial block
+  test('prunes and restores replaced-by-fee transactions', async () => {
```
rafaelcr, line 2210: @zone117x this is the main test of the new functionality

zone117x (commented):
The code looks great, nice tests 💯 

The new sql query is in the block ingestion path -- have we tested its performance?

janniks (approved):
Nice test 👍

zone117x (approved):

PR #2277, opened 2025-05-12
fix: consider microblock transactions in balance calculations

### LINKED ISSUE #2267: new stx balance endpoint returning incorrect amounts in some cases
Reported via Leather: https://github.com/leather-io/extension/issues/6214

Some accounts are reported with incorrect balances. Seems to happen with accounts that have STX locked, such that when the wallet subtracts balance - locked the amount ends up negative.

The current theory is that there could be a re-org bug in the queries introduced in this PR https://github.com/hirosystems/stacks-blockchain-api/pull/2030 but we need to reproduce in a test to confirm.

rafaelcr:
The API has Stacks 2.x logic to avoid re-inserting transactions to the DB during a `/new_block` event when those transactions were previously inserted during a `/new_microblock` event. However, the new /extended/v2 balance endpoint logic only considers `/new_block` transactions when adding/subtracting from account balances. This is causing some old 2.x blocks to warp balances in accounts that happened to have transactions confirmed in previous microblocks.

This PR changes this behavior so these balance calculations always include all transactions regardless of microblocks. It also includes a migration that will execute a query to recompute all transaction balances to correct the issue on deployed environments.

Fixes #2267

### DIFF
--- a/migrations/1747151261415_recalculate-balances.js
+++ b/migrations/1747151261415_recalculate-balances.js
@@ -1,0 +1,97 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = pgm => {
+  // Recalculate STX balances
+  pgm.sql(`
+    WITH all_balances AS (
+        SELECT sender AS address, -SUM(amount) AS balance_change
+        FROM stx_events
+        WHERE asset_event_type_id IN (1, 3) -- Transfers and Burns affect the sender's balance
+        AND canonical = true AND microblock_canonical = true
+        GROUP BY sender
+      UNION ALL
+        SELECT recipient AS address, SUM(amount) AS balance_change
+        FROM stx_events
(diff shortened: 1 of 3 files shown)


migrations/1747151261415_recalculate-balances.js
```diff
@@ -0,0 +1,97 @@
...
+      GROUP BY recipient
+    ),
+    all_addresses AS (
+      SELECT address FROM net_balances
+      UNION
+      SELECT address FROM fees
+      UNION
+      SELECT address FROM rewards
+    )
+    INSERT INTO ft_balances (address, balance, token)
```
zone117x, line 58: Does this work with an already populated table? I think the table might need to be dropped first, or this needs something like 
```sql
        ON CONFLICT (address, token)
        DO UPDATE ...
```

rafaelcr: Whoops, forgot that, thanks. Added the TRUNCATE query.

zone117x (approved):
LGTM!

Issue #2148, opened 2024-10-29
Tenure endpoint for epoch 3.0


obycode:
It might be nice to add a tenure endpoint that includes things like number of blocks in the tenure and total cost of all blocks in that tenure.

friedger:
For the current tenure, this would be the budget use until now

PR #2285, opened 2025-05-15
feat: add /extended/v2/block-tenures/:height/blocks endpoint

rafaelcr:
Adds a new endpoint under `/extended/v2/block-tenures/:height/blocks` so users can get blocks only for a specific tenure. This endpoint builds over the code for the v2 blocks endpoint so it also inherits cursor pagination.

A new route is created so we can open the possibility of adding a REST endpoint to get stats or data from a specific tenure as requested in #2148

Replaces #2144


zone117x (approved):
Looks great!

PR #2289, opened 2025-05-22
fix: count stx mint data at block 0 towards account balances

rafaelcr:
Our `chain_tip` table starts with `block_height` at zero, so when the API received the block 0 boot data from the Stacks node it is always marked as non-canonical.

This affected FT balance calculations because STX mint events contained within were not being considered, especially in devnet or testnet deployments where devs were configuring `[[ustx_balance]]` settings in their Stacks nodes.

This PR makes it so block 0 STX mint data is always considered for balances.

Related to #2267


zone117x (approved):
Nice fix!

PR #2291, opened 2025-05-26
feat: SNP integration

zone117x:
Implements SNP event streaming support in the API. An integration test harness is also implemented which spins up a local SNP server, populates it with events from a tsv (simulating a stacks-node), then ensures the API is able to sync all events from SNP.


src/event-stream/snp-event-stream.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  async handleMsg(messageId: string, timestamp: string, path: string, body: any) {
+    this.logger.debug(`Received SNP stream event ${path}, msgId: ${messageId}`);
+
+    const response = await this.eventServer.fastifyInstance.inject({
+      method: 'POST',
+      url: path,
+      payload: body,
+    });
```
zone117x, line 59: The SNP event streaming implementation is currently layered on top of the existing http event observer code, as you can see here with the `fastify.inject(...)`. There's a significant amount of code in the event observer around logging, IDB, etc which is able to be reused here without a significant refactoring and/or code duplication. 

I _think_ this is fine, but probably not the most optimized in terms of memory usage. I'm assuming this approach results in event bodies being copied in memory at least once.

rafaelcr: I think this is good tradeoff for now IMO, we could revisit this decision if we see significant memory crunch in our writer

rafaelcr (approved):
This looks great! Thanks @zone117x !

Issue #2303, opened 2025-06-24
SNP mode does not start


CharlieC3:
Despite setting `SNP_EVENT_STREAMING` to `true` and `SNP_REDIS_URL`, the API seems to keep trying to run the standard event observer.

PR #2304, opened 2025-06-24
fix: ensure some ops only run when SNP is not enabled

CharlieC3:
Closes https://github.com/hirosystems/stacks-blockchain-api/issues/2303

* Adds some conditional checks to:
    * skip the chain ID check when SNP mode is enabled
    * skip core node monitoring if SNP is enabled
* Adds missing function call to start SNP stream

Seems to be working in dev with these changes!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -147,7 +147,8 @@ async function init(): Promise<void> {
     });
 
     const skipChainIdCheck = parseBoolean(process.env['SKIP_STACKS_CHAIN_ID_CHECK']);
-    if (!skipChainIdCheck) {
+    const snpEnabled = parseBoolean(process.env['SNP_EVENT_STREAMING']);
+    if (!skipChainIdCheck && !snpEnabled) {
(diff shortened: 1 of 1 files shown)


rafaelcr (approved):
Great! Thanks for the fix!

rafaelcr:
hey @CharlieC3 yep we should definitely only be streaming chain events. Signer events will just be ignored but they still spend some CPU on request/response time and they also get saved to the Stacks events table

CharlieC3:
Hmm changing the API to only ingest `chainEvents` from SNP doesn’t seem to be working, after connecting to SNP there is simply no activity in the logs.
While the SNP client may support multiple types, it seems [the SNP backend](https://github.com/hirosystems/salt-n-pepper/blob/main/src/redis/redis-broker.ts#L10) is lagging behind a bit and currently only supports all events.

I'll merge this PR for now since it seems switching to only chain events will be more complicated than I previously thought. That effort can be placed in a separate PR.

PR #2311, opened 2025-07-02
fix: only ingest snp chain events to avoid db bloat

CharlieC3:
Ensures only relevant chain events are ingested when using SNP, as opposed to ingesting all events.
Tested and verified against the latest release of SNP as a part of https://github.com/hirosystems/salt-n-pepper/pull/28


rafaelcr (approved):
:shipit:

PR #2314, opened 2025-07-03
fix: use burn_block_time in 2.x blocks and block_time after 3.x

rafaelcr:
Stacks core now sends `block_time` for Nakamoto blocks after https://github.com/stacks-network/stacks-core/pull/5113 was merged, and `null` for Stacks 2.x blocks.

This PR changes our DB code so we always try to use the `block_time` but default to `burn_block_time` for 2.x blocks.


src/event-stream/event-server.ts
```diff
@@ -1000,17 +1000,11 @@ export function parseNewBlockMessage(
...
+    // blocks. We'll use the parent burn block timestamp as the receipt date for those.
+    block_time: msg.block_time ?? msg.burn_block_time ?? Math.round(Date.now() / 1000),
   };
 
-  if (!blockData.block_time) {
-    // If running in IBD mode, we use the parent burn block timestamp as the receipt date,
-    // otherwise, use the current timestamp.
-    const stacksBlockReceiptDate = isEventReplay
-      ? msg.burn_block_time
-      : Math.round(Date.now() / 1000);
```
haneytron, line 1010: Ahh, I see it came from here.

haneytron (approved):
LGTM; see 2 comments (non-blocking)

src/event-stream/event-server.ts
```diff
@@ -1000,17 +1000,11 @@ export function parseNewBlockMessage(
   const parsedTxs: CoreNodeParsedTxMessage[] = [];
   const blockData: CoreNodeMsgBlockData = {
     ...msg,
+    // Nakamoto blocks now include their own `block_time`, but this will be empty for pre-Nakamoto
+    // blocks. We'll use the parent burn block timestamp as the receipt date for those.
+    block_time: msg.block_time ?? msg.burn_block_time ?? Math.round(Date.now() / 1000),
```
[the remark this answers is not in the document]
rafaelcr: Good idea, done

src/event-stream/event-server.ts
```diff
@@ -997,20 +997,20 @@ export function parseNewBlockMessage(
 ) {
   const counts = newCoreNoreBlockEventCounts();
 
+  // Nakamoto blocks now include their own `block_time`, but this will be empty for pre-Nakamoto
+  // blocks. We'll use the parent burn block timestamp as the receipt date for those. If both are
+  // blank, there's something wrong with Stacks core.
+  const block_time = msg.block_time ?? msg.burn_block_time;
+  if (!block_time) {
```
csgui, line 1004: if `block_time == 0` the error will be thrown. Isn't more appropriate to explicitly checks for both null and undefined, but not falsy values like 0? Like `if (blockTime == null) {...}`

rafaelcr: Hah you read my mind! Just added this

PR #2326, opened 2025-07-21
fix: optimize replace-by-fee mempool calculations

rafaelcr:
* Add indexes to optimize for accounts with very high nonces
* Instead of looping through `tx_id`s with JS and performing 1 query per ID, send all IDs to postgres in a single query (batched by sets of 500 to avoid overwhelming postgres).

Related to #2271


CharlieC3 (approved):
Nicely done! :shipit:

PR #2330, opened 2025-07-22
feat: tx type filter

He1DAr:
## Description

We're planning to support filtering by tx type in Stacks explorer search. This PR aims to add a new filter `type` to allow user filtering txs by type.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [ ] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/migrations/1753214286445_txs-status-index.js
+++ b/migrations/1753214286445_txs-status-index.js
@@ -1,0 +1,16 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = pgm => {
+  pgm.createIndex('txs', 'status', {
+    where: 'canonical = TRUE AND microblock_canonical = TRUE',
+    name: 'idx_txs_status_optimized'
+  });
+};
+
+exports.down = pgm => {
+  pgm.dropIndex('txs', 'status', {
+    name: 'idx_txs_status_optimized'
+  });
+};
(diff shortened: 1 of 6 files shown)


migrations/1753214286445_txs-status-index.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = pgm => {
+  pgm.createIndex('txs', 'status', {
+    where: 'canonical = TRUE AND microblock_canonical = TRUE',
+    name: 'idx_txs_status_optimized'
```
rafaelcr, line 8: No need to provide a name, `node-pg-migrate` will define one for you with the current standards we use

migrations/1753214286445_txs-status-index.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = pgm => {
+  pgm.createIndex('txs', 'status', {
```
rafaelcr, line 6: Adding the index is great but I don't think it will help a lot especially for the `success` status, given that the vast majority of transactions are successful.

I'm worried that running the endpoint on mainnet will cause performance issues when it tries to calculate the `COUNT(*)` of records to show in the `total`. We'll need to run load test on this.

rafaelcr (commented):
This looks very good @He1DAr but we need to run some load tests on this before we can merge. Let's discuss how we can do this.

PR #2333, opened 2025-07-28
feat(prometheus): add chain tip metric

rafaelcr:
Adds a metric to reflect canonical chain tip
```
# HELP chain_tip_block_height Current chain tip block height
# TYPE chain_tip_block_height gauge
chain_tip_block_height 24
```


CharlieC3 (approved):
:shipit: Thanks for the quick turnaround!

PR #2355, opened 2025-10-03
fix: calculate burnchain reward reorgs correctly

rafaelcr:


CharlieC3:
I currently am testing the migration and how long it takes to finish in dev. The writer pod gets killed after an hour like you said, but I think it's because Istio is cutting off the connection rather than postgres. Upon restart of the writer, it crashes again saying a migration is already running, and I've confirmed this via PGHero. So it seems like Istio killing the connection and causing the writer to crash may be ok, since the migration continues.

I'll continue monitoring its progress, this will likely be a upgrade deployed next week.

PR #2368, opened 2025-10-25
chore: remove unused uuid package

pradel:
## Description

While exploring the codebase, I noticed that the uuid package that is defined in the dependencies of the project is actually not used, so it can be removed.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

No

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


rafaelcr (approved):
Thanks @pradel !

PR #2381, opened 2025-11-26
feat: migrate default postgres schema

### LINKED ISSUE #2373: Update default schema
https://github.com/hirosystems/stacks-blockchain-api/blob/32f879f7e934d26b35f869c549a096b4b2fa5af2/.env#L6

It's generally not a good practice to use the `public` schema in a PG database, and instead to use one created for the project. Ideally the default schema would be `stacks_blockchain_api`, like the database.

rmottley-hiro:
- Migrate schema from public to stacks-blockchain-api. Permission to start migration requires affirmative consent from user

Closes #2373 

BREAKING CHANGE: This change can be made without a full rebuild from genesis. However, it is considered breaking in the interest of data safety

## Testing information

Testing is recommended. Ensure `PG_OLD_SCHEMA` matches the current schema name in your database. `PG_SCHEMA` is pre-set, but can be modified as the user wishes. **Uncomment `PG_MIGRATE_SCHEMA_ON_STARTUP`** (app will fail to start until this is set as a value to prevent accidental migration)

## Checklist
- [x] `npm run test` passes
- [ ] Changelog is updated

### DIFF
--- a/tests/utils/test-helpers.ts
+++ b/tests/utils/test-helpers.ts
@@ -45,7 +45,13 @@ import { CoreRpcPoxInfo, StacksCoreRpcClient } from '../../src/core-rpc/client';
 import { DbBlock, DbTx, DbTxStatus } from '../../src/datastore/common';
 import { PgWriteStore } from '../../src/datastore/pg-write-store';
 import { BitcoinAddressFormat, ECPair, getBitcoinAddressFromKey } from '../../src/ec-helpers';
-import { coerceToBuffer, connectPostgres, runMigrations, timeout } from '@hirosystems/api-toolkit';
+import {
+  coerceToBuffer,
+  connectPostgres,
+  PgConnectionArgs,
+  runMigrations,
+  timeout,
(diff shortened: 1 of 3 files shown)


rmottley-hiro:
> Thanks for this PR @rmottley-hiro ! If we merge this change into the `next` branch we don't need to add the extra steps for users to accept this change as it will be breaking by default. Let's discuss this on monday 👍

Sounds good! 👍

tests/utils/test-helpers.ts
```diff
@@ -52,6 +52,11 @@ import { AddressStxBalance } from '../../src/api/schemas/entities/addresses';
 import { ServerStatusResponse } from '../../src/api/schemas/responses/responses';
 
 export async function migrate(direction: 'up' | 'down') {
+  const connArgs = await createSchema();
```
rafaelcr: It would be cleaner if you separate the schema creation from the connection arguments
```suggestion
  const connArgs = getConnectionArgs();
  await createSchema(connArgs);
```

PR #2404, opened 2025-12-12
fix: create new schema on db start in docker-compose files

rmottley-hiro:
- Ensure new `stacks_blockchain_api` schema is created as the postgres docker container spins up


CharlieC3 (approved):
:shipit: LGTM!

PR #2590, opened 2026-06-22
fix: update to latest codec, parse new post conditions

rafa-stacks:
Update `@stacks/codec` to `2.0.0-pox5.5` and parse new post conditions.


brice-stacks (approved):
Lgtm

PR #2593, opened 2026-06-24
ci: switch to new release method

rafa-stacks:
Use the new manual release method and push v8 Docker images to DockerHub and GHCR


bgridley-stacks (approved):
Approved with comments from AI:


---
🔴 High — npm publish almost certainly lacks working auth

publish-npm runs npm publish --provenance with id-token: write but no NPM_TOKEN, no NODE_AUTH_TOKEN, no registry-url on setup-node, and no .npmrc (confirmed — none exist in the repo).

The id-token: write + --provenance combo signals an intent to use npm OIDC trusted publishing, but:
- .nvmrc pins Node v22, which ships npm 10.x. OIDC trusted publishing requires npm ≥ 11.5.1, and the npm install -g npm@~11.10.0 upgrade step that existed in the old ci.yml was removed and not carried into release.yml.
- Provenance attestation is not authentication — without a token or working OIDC, npm publish will fail with ENEEDAUTH.

Note this is also net-new functionality (the old semantic-release config didn't include @semantic-release/npm), so it can't be assumed to work from prior runs. Action: either add registry-url: 'https://registry.npmjs.org' to setup-node + NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}, or add an npm i -g npm@latest step and confirm trusted publishing is configured on npmjs.com for @stacks/blockchain-api-client.

🟠 Medium — jobs run in parallel with no ordering → partial releases

publish-npm, publish-docker, and github-release have no needs between them. If npm fails (see above), Docker images and a GitHub release still get created — a half-published version that's awkward to recover from (npm version numbers can't be reused). Consider chaining: publish-docker/github-release needs: publish-npm, or at least gate the GitHub release on the publish jobs.

🟠 Medium — release.sh tag/merge ordering is ambiguous

The script bumps versions on the current branch (no commit) then prints:
Create and merge Pull Request with version bump
Push the tag: git tag v${VERSION} && git push --tags
Trigger the Release workflow
But git tag v$VERSION tags the local commit, while the PR merge (especially squash) produces a different commit on master. The tag — which release.yml checks out via ref: v${{ inputs.version }} — should point at the merged commit. Recommend the script either not suggest tagging from local, or explicitly say "after merge, git checkout master && git pull && git tag v$VERSION". As written it's an easy way to publish the wrong tree.

PR #2607, opened 2026-07-14
fix: deprecate microblock endpoints and improve faucet error handling

rafa-stacks:
## Summary

Two related bits of API cleanup:

### Deprecations
- Marked the three microblock endpoints (`get_microblock_list`, `get_microblock_by_hash`, `get_unanchored_txs`) as `deprecated: true` with a note that microblocks were removed in the Nakamoto upgrade.
- Marked `GET /extended/v1/faucets/btc/:address` (`get_btc_balance`) as deprecated.

### Faucet error handling
Prompted by three user error reports where faucet failures surfaced as raw `500`s that leaked internal details (e.g. `127.0.0.1:18443`, an internal `*.svc.cluster.local` node URL) or opaque messages. Added an encapsulated `setErrorHandler` on the faucet routes that translates failures into sanitized responses while logging full detail server-side:

- **Backing node unreachable** (`ECONNREFUSED`/`ECONNRESET`/`ETIMEDOUT`/etc., incl. `cause`-nested and `fetch failed`) → `503`.
- **Database unavailable** (`isPgConnectionError`) → `503`.
- **Faucet out of funds** — Stacks `NotEnoughFunds` (STX/sBTC) and BTC `not enough total amount in utxo set` → `503` with an "out of funds" message.
- **Anything else** → sanitized `500`.
- **Validation / explicit 4xx** → passed through unchanged.
- Guards on `reply.sent` so an already-flushed response is never rewritten.

Also: the STX faucet now rotates to its other configured keys on `NotEnoughFunds` (previously only on nonce conflicts), so a single drained key doesn't fail a request when other keys still hold funds.

## Note for reviewers / ops
Two of the source reports are ultimately deployment/funding issues, not code bugs — the code change makes them fail gracefully rather than fixing the root cause:
- `testnet-pox5` STX faucet account is drained and needs refilling.
- `private-1` BTC faucet has no *spendable* (mature/confirmed) UTXOs.
- The earlier `testnet-pox5` BTC report needs a reachable bitcoind regtest node, or the BTC faucet disabled on that host.

## Testing
- `tsc --noEmit` and `eslint` pass.
- Existing `tests/krypton/faucet-btc` assertions cover only the middleware 403 paths, which are untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/api/routes/v1/faucets.ts
+++ b/src/api/routes/v1/faucets.ts
@@ -241,7 +241,7 @@ export const FaucetRoutes: FastifyPluginAsync<
       preHandler: [btcFaucetEnabledMiddleware, missingBtcConfigMiddleware],
       schema: {
         operationId: 'run_faucet_btc',
-        summary: 'Get BTC testnet tokens',
+summary: 'Get BTC regtest tokens',
         description: `Add 0.01 BTC token to the specified regtest BTC address.
 
         The endpoint returns the transaction ID, which you can use to view the transaction in a regtest Bitcoin block
--- a/src/api/routes/v1/faucets.ts
+++ b/src/api/routes/v1/faucets.ts
@@ -241,7 +241,7 @@ export const FaucetRoutes: FastifyPluginAsync<
       preHandler: [btcFaucetEnabledMiddleware, missingBtcConfigMiddleware],
       schema: {
         operationId: 'run_faucet_btc',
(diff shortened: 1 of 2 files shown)


bgridley-stacks (approved):
LGTM

PR #2635, opened 2026-08-12
feat: add signer_manager to PoX cycle signer endpoints

rafa-stacks:
Adds a `signer_manager` field to the signer entries returned by the PoX cycle endpoints:

- `GET /extended/v2/pox/cycles/:cycle_number/signers`
- `GET /extended/v2/pox/cycles/:cycle_number/signers/:signer_key`

The field contains the principal of the signer manager contract that registered the entry's signing key via pox-5 `register-signer` (e.g. `SP...ADDR.signer-manager`), or `null` if the key has no registration.

```json
{
  "signing_key": "0x038e3c4529395611be9abf6fa3b6987e81d402385e3d605a073f42f407565a4a3d",
  "signer_address": "STRYYQQ9M8KAF4NS7WNZQYY59X93XEKR31JP64CP",
  "signer_manager": "STRYYQQ9M8KAF4NS7WNZQYY59X93XEKR31JP64CP.signer-manager",
  "weight": 5,
  "stacked_amount": "686251350000000000",
  "weight_percent": 55.55,
  "stacked_amount_percent": 50,
  "solo_stacker_count": 1,
  "pooled_stacker_count": 0
}
```

### DIFF
--- a/src/api/schemas/v1/entities/pox.ts
+++ b/src/api/schemas/v1/entities/pox.ts
@@ -1,4 +1,5 @@
 import { Static, Type } from '@sinclair/typebox';
+import { SmartContractIdParamSchema } from '../params.js';
 
 export const PoolDelegationSchema = Type.Object({
   stacker: Type.String({
@@ -39,6 +40,10 @@ export type PoxCycle = Static<typeof PoxCycleSchema>;
 export const PoxSignerSchema = Type.Object({
   signing_key: Type.String(),
   signer_address: Type.String({ description: 'The Stacks address derived from the signing_key.' }),
+  signer_manager: Type.Union([SmartContractIdParamSchema, Type.Null()], {
+    description:
+      'The principal of the signer manager contract that registered this signing key, or null if the key has no registration.',
+  }),
   weight: Type.Integer(),
   stacked_amount: Type.String(),
   weight_percent: Type.Number(),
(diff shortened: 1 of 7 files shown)


src/api/schemas/v1/entities/pox.ts
```diff
@@ -39,6 +40,10 @@ export type PoxCycle = Static<typeof PoxCycleSchema>;
 export const PoxSignerSchema = Type.Object({
   signing_key: Type.String(),
   signer_address: Type.String({ description: 'The Stacks address derived from the signing_key.' }),
+  signer_manager: Type.Union([SmartContractIdParamSchema, Type.Null()], {
```
adriano-stacks, line 43: There can be more than one signer manager associated to a signer (in which case, the signer's weight is the sum of the STX staked by each signer manager). 

For an example, see `0285f4c617c4065e429d1666dcfd885c52e223acfc042aa2c2855adc93034e6975` in cycle 141, who has registered with signer-managers `SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.signer-manager-xverse-v1` and `SP8HK160YD5GHXP69VGA0TC7AQJ1X4CDW3XVERSE.xverse-signer-manager-2`.

adriano-stacks (changes requested):

brice-stacks:
Is the data in `staking_signer` only storing the latest, but the endpoint allows you to select any cycle number? Couldn't that result in wrong data for old cycles?

rafa-stacks:
After discussing with @brice-stacks , a new endpoint is needed to map all the signer manager key business rules correctly. Closing this PR and replacing it with https://github.com/stx-labs/stacks-blockchain-api/pull/2636

PR #2639, opened 2026-08-19
feat: v3 STX supply endpoint backed by a materialized `chain_tip` counter

rafa-stacks:
Adds `GET /extended/v3/tokens/stx/supply`, which returns the total liquid STX supply from a single-row `chain_tip` lookup, replacing the v1 endpoint whose queries aggregate over the entire `stx_events` and `miner_rewards` tables on every request.

```json
{
  "total": "1470469916700000",
  "projected_total_2050": "2318000000000000"
}
```

Amounts are string-quoted micro-STX (µSTX) integers, following v3 conventions (v1 returned decimal STX). No query parameters; responses are served behind the chain-tip ETag cache.


brice-stacks (approved):
lgtm

PR #2643, opened 2026-08-25
feat: switch to the new canonical chain tip immediately during Nakamoto re-orgs

### LINKED ISSUE #2299: switch to the new canonical chain tip earlier during re-orgs
Today, the API has the following logic to detect re-orgs:
- Chain tip **A** is at block `400`
- A new block `398` is received which builds off of `397`. This creates chain tip **B**. It's saved but considered non-canonical.
- Blocks `399` and `400` come, building on **B**. Still non-canonical.
- Block `401` comes on **B**. **B** is now the canonical chain and **A** is orphaned.

This logic is similar to how Bitcoin block re-orgs are handled on the L1 and it made sense when Stacks was at version 2.x.

After 3.x (Nakamoto), however, Stacks Signers are responsible for keeping track of the canonical chain. This means that if the API receives a block it was already validated by signers and it should already be considered the new canonical chain tip. The API should then consider this new chain as canonical immediately:
- Chain tip **A** is at block `400`
- A new block `398` is received which builds off of `397`. This creates chain tip **B** and it's immediately marked as canonical. **A** is now orphaned.

Making this change would allow the Explorer to show blocks much faster during a re-org, and it would even open the possibility of creating a new API endpoint which recent re-org information so the Explorer could display a message letting users know a re-org is in process.

This would be a very large structural change on the API.

rafa-stacks:
Closes #2299.

Since Stacks 3.x, blocks only reach the API's event stream after signers have validated them, so every `/new_block` event is the node's canonical tip. The API still applied the 2.x "longest chain wins" rule, so during a re-org onto a shorter fork it kept serving the stale chain until the fork overtook the old tip. This PR makes every incoming **Nakamoto** block immediately canonical (even when that moves the chain tip *backwards*) while keeping legacy behavior for pre-Nakamoto blocks so 2.x event replay is unchanged.

### DIFF
--- a/src/datastore/helpers.ts
+++ b/src/datastore/helpers.ts
@@ -1389,6 +1389,15 @@ export function convertTxQueryResultToDbMempoolTx(txs: TxQueryResult[]): DbMempo
   return dbMempoolTxs;
 }
 
+/**
+ * Determines if a block was produced under Nakamoto (epoch 3.0) rules. Nakamoto blocks are
+ * validated by signers before they are announced by stacks-nodes, which is signaled by the
+ * presence of a `signer_bitvec` value.
+ */
+export function isNakamotoBlock(block: Pick<DbBlock, 'signer_bitvec'>): boolean {
+  return block.signer_bitvec !== null;
+}
+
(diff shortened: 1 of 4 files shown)


brice-stacks (approved):
I think this looks good!

